PR #29, opened 2016-03-09
Fix thread-safety of logging that leads to unpredictable freezes of the GUI

zol3:
textArea.append is not threadsafe, which made the GUI freeze. Using invokeLater is the documented way to handle the EventQueue when multithreading.

### DIFF
--- a/src/org/myrobotlab/control/widget/Console.java
+++ b/src/org/myrobotlab/control/widget/Console.java
@@ -214,9 +214,17 @@ public class Console extends AppenderSkeleton implements Appender<ILoggingEvent>
 	public void doAppend(ILoggingEvent loggingEvent) throws LogbackException {
 		//append(loggingEvent);
 		if (logging) {
-			String msg = String.format("[%s] %s", loggingEvent.getThreadName(), loggingEvent.toString()).trim();
-			textArea.append(msg + "\n");
+			final String msg = String.format("[%s] %s", loggingEvent.getThreadName(), loggingEvent.toString()).trim();
+
+			// textarea not threadsafe, needs invokelater
+			EventQueue.invokeLater(new Runnable() {
+		            //@Override
+		            public void run() {  	
+		            	textArea.append(msg + "\n");
+		            }
+			});
 		}
 		
 	}
+	
 }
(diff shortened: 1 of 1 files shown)


kwatters:
Ahoy zol3!  Thanks for the patch!  It looks like there might be a classpath issue in the build.xml.  I'll have a look at your patch after lunch and merge it in!  Welcome!

zol3:
Hi kwatters,

import java.awt.EventQueue needs to go on top. I overlooked it when forking/copying/pushing, my bad. Need to run now, will fix it in a few hours.

cheers, Zoltan

zol3:
Ok, I cleaned up my eclipse and pushed the missing line. Should be fine now.

zol3:
There were 2 other instances where textarea.append was called in Console, so I prepped them up too, in case something is calling them. 

However, I'm a little confused about the log4j implementation. As I understand it, doAppend isn't supposed do any writing, but call append(). Also it seems that the entire skeleton class is being nuked, so filters etc wouldn't work. Is there any reason for this? Also, where is the log4j configuration, couldnt find it? thanks, Z

kwatters:
Hi Zol3!  Thanks for following up on this!  I believe the idea is that there is no log4j.xml or log4j.properties file.  I think GroG will have to chime in on the details of that a bit more.  As for the swing based GUI...  it's been a bit neglected as we've started moving towards the new AngularJS based UI.  You can check out that service by starting the WebGui service in MRL.  Reality is, the swing gui is still used by a lot of people, so we need to/should continue to maintain it.  (Sorry GroG) 
I'll have a look at the patch and we'll merge it.  Thanks for helping out!!!  It takes a community to build a robot :)

supertick:
Hi zol3.  Thanks for the contribution.
Logging should be :
1. a reasonable log level when starting
2. controllable without restart or external xml configuration
Logging was abstracted some in the process of moving from log4j to slf4j ... it's still messy...  slf4j is still the goal, but a binary bridge or logback will need to be bundled with the build.

I have a tendency not to want to deal with the Swing gui as kwatters mentioned, but he's right - people still depend on it.

PR #100, opened 2017-08-15
Update mrlpy integration

AutonomicPerfectionist:
This pull request provides a massive refactor of the PythonProxy service to add features including message routing and native method return values. Also adds the GoogleAssistant service, a preliminary test in adding Assistant into MRL (manual authentication through the virtualenv required currently). Dependency on com.nativeServices was added, but unknown if it extracts correctly (should extract similarly to InMoov and create the folder in the base MRL directory).


AutonomicPerfection:
Hmmm.... Seems as if the pull request also included some previous commits... Have to see whether these will cause a problem

kwatters:
Please resubmit the pull request without all of your eclipse files (.metadata).   Looks cool. I watched the video you posted.

AutonomicPerfection:
Whoops, didn't realize those were added. Funny thing is, I didn't use Eclipse (running over SSH connection). Files must have appeared during a test with NFS where I was seeing how Eclipse performed. Thanks for the heads up, and I'll fix it right away

LunDev:
Don't know about the added feature itself, but otherwise LGTM

AutonomicPerfectionist:
Hmmm... That's not supposed to happen... Yeah, the dependency may be removed as long as native services are not started by the java end (don't call PythonProxy.startNativeService()). Could you describe the error, so that I can see what's up?

supertick:
kwatters reported the dependency did not load when he did a
 java -jar myrobotlab -install
which is supposed to install all dependencies. I suspect you did not setup com.nativeServices in the repo with an ivy file ?  Before I commented that line out, kwatters reported errors downloading this dependency.

AutonomicPerfectionist:
Yep, that's exactly the problem. I forgot to change the URL on my git local back to official repo, so I pushed the dependency to my fork. New pull request in repo to fix this.

PR #126, opened 2017-09-21
servo update try 2

supertick:


supertick:
do not accept this merge request - for viewing purposes only - please review setAutoDisable / Enable

moz4r:
yeh cleaner !

just one thing it is risky to set autodisable to true by default for all, no ? but why not
In all cases we can set default autoEnable to true and remove exposed method too

supertick:
closing

PR #219, opened 2018-06-18
fixed compiler errors in Inmooov2.java Runtime.java and Security.java by making some necessary methods static

DragonEagle:
Some methods were not declared as static that needed to be for MRL to compile

### DIFF
--- a/src/main/java/org/myrobotlab/service/Runtime.java
+++ b/src/main/java/org/myrobotlab/service/Runtime.java
@@ -2266,20 +2266,20 @@ public class Runtime extends Service implements MessageListener {
 	  runtime.setLocale(new Locale(language));
   }
 
-  public void setLocale(String language) {
+  public static void setLocale(String language) {
 	  setLocale(new Locale(language));
   }
 
-  public void setLocale(String language, String country) {
+  public static void setLocale(String language, String country) {
 	  setLocale(new Locale(language, country));
   }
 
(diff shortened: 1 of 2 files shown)


kwatters:
Welcome DragonEagle!  I didn't realize the build was broken until I updated  today.  So, these methods shouldn't be static.  As far as I understand the idea is that Security is a singleton and you should use it with the getInstance() method.

Security.getInstance()

I know GroG was part the way though doing this and probably committed some stuff that he wasn't ready to commit.   For now , I feel like it probably makes more sense to just roll back GroG's commits until he has a chance to clean up and finish that work.

DragonEagle:
I could go ahead and make it a singleton.  Wouldn't take me but an hour or 2

kwatters:
Awesome, I think that's cool, but at this point, I want GroG to chime in.  He was doing the work here, so I don't want to speak for him.  I've reverted the head of develop to just before where the build broke , so for now, at least the build is passing again.

DragonEagle:
Then I will just roll back for now and keep my ears open

kwatters:
Thank you very much for helping!

supertick:
Sorry I left it in a broken state, I was trying to give moz4r a way to proceed with centralized localization - but I was in the middle of refactoring all the speech service.  Turns out, this was a much bigger job than I thought it was going to be.   @kwatters - thanks for backing it out ...

PR #249, opened 2018-08-14
motor refactor and virtual joystick

supertick:
This pull request is a refactoring of Motor & MotorController.
Its been tested and works for Sabertooth controller with port motors.
Further refactoring should be done to promote more code into the abstracts and remove more code from the concrete services.

It also includes the capability of "virtualizing" Joystick controllers by cloning them into json files then reloading them with the same "capabilities" as real hardware, but accessible in a programmatic way.

e.g. joy.moveAxis("z", 0.03)  or joy.pressButton("b")

Please review, thanks


supertick:
I agree completely,
I have been committing on the develop integration branch directly for way too long.  "Finally", I am changing this bad behavior (and introducing other bad behaviors).

I appreciate your excellent critique MaVo and will make efforts to do smaller controlled merge requests.

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
```diff
@@ -162,59 +147,39 @@ public boolean isAttached(MotorController controller) {
 
   @Override
   public boolean isInverted() {
```
kwatters: this should be moved to the Mapper class , right?

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
```diff
@@ -224,45 +189,47 @@ public void moveTo(double newPos) {
...
-    // inverted
+    this.inverted = invert;
     broadcastState();
   }
 
   // ---- Servo begin ---------
-  public void setMinMax(double min, double max) {
-    powerMap.setMin(min);
-    powerMap.setMax(max);
+  public void setMinMaxOutput(double min, double max) {
```
kwatters: this should call setMin / setMax on a Mapper class.

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
```diff
@@ -224,45 +189,47 @@ public void moveTo(double newPos) {
...
+    minOutput = min;
+    maxOutput = max;
     broadcastState();
   }
 
-  public void setPowerLevel(double power) {
-    powerLevel = power;
+  public void map(double minX, double maxX, double minY, double maxY) {
```
kwatters: This method should not exist here, it should be on the mapper class.

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
```diff
@@ -417,4 +385,36 @@ public boolean isAttached(String name) {
     return ret;
   }
 
+  public Double getMinX() {
```
kwatters: all of these should delegate the return value from the underlying mapper object.

src/main/java/org/myrobotlab/service/abstracts/AbstractMotorController.java
```diff
@@ -1,17 +1,224 @@
...
+import org.myrobotlab.framework.interfaces.Attachable;
 import org.myrobotlab.math.Mapper;
+import org.myrobotlab.service.Motor;
+import org.myrobotlab.service.Runtime;
+import org.myrobotlab.service.interfaces.MotorControl;
 import org.myrobotlab.service.interfaces.MotorController;
 
 public abstract class AbstractMotorController extends Service implements MotorController {
+  
+  protected transient Map<String, AbstractMotor> motors = new HashMap<String, AbstractMotor>();
```
kwatters: Why AbstractMotor, we should probably pass the interface here...  What are we missing on the interface that AbstractMotor has ?

src/main/java/org/myrobotlab/service/abstracts/AbstractMotorController.java
```diff
@@ -1,17 +1,224 @@
...
+  public void map(double minX, double maxX, double minY, double maxY) {
+    this.minX = minX;
+    this.maxX = maxX;
+    this.minY = minY;
+    this.maxY = maxY;
+    broadcastState();
+  }
+  
+  // TODO - switch to MotorControl interface instead of AbstractMotor
+  public double calcOutput(AbstractMotor mc) {
```
kwatters: All of this logic should probably move to the Mapper object..  also, we should probably pass the interface here, not the abstract...

src/main/java/org/myrobotlab/service/abstracts/AbstractMotorController.java
```diff
@@ -1,17 +1,224 @@
...
+  public void motorMove(String name) {
+    motorMove((MotorControl) Runtime.getService(name));
+  }
+
+  // FIXME - if kept they should be put in interface
+  public void motorMoveTo(String name, Integer position) {
+    error("not implemented");
+  }
+  
+  @Override
```
kwatters: should throw an unimplemented exception or they should have an implementation

src/main/java/org/myrobotlab/service/abstracts/AbstractMotorController.java
```diff
@@ -1,17 +1,224 @@
...
+  public void detach(MotorControl device) {
+    motors.remove(device);
+  }
+  
+  /**
+   * Routing Attach - routes ServiceInterface.attach(service) to appropriate
+   * methods for this class
+   */
+  @Override
+  public void attach(Attachable motor) throws Exception {
```
kwatters: This is a vague interface..  seems like we might want to pass the MotorControl interface rather than just blindly allowing anything that's attachable?  Honestly, I'm not sure how we should handle interfaces like this.  The intended behavior of this method could become ambiguous in some scenarios.

src/main/java/org/myrobotlab/service/interfaces/MotorControl.java
```diff
@@ -115,7 +115,7 @@
 	 */
 	void setInverted(boolean invert);
 
-	void setPowerLevel(double power);
+	// void setPowerLevel(double power);
```
kwatters: maybe keep this around and just deprecate it?  you can delicate to the prefered moveTo method.

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
supertick: I don't believe so .. I moved the fields and decomposed Mapper into AbstractMotor - 
Mapper I think is flawed in a couple ways, setting or re-setting map values will remove min/max.
I think Mapper does more than it should and the cost is removal of min/max when creating new map values.  I didn't want to risk cause problems to Servos (which uses mappers), which I don't want to touch now.

I also don't want to bury data in sub-components ... 

In general terms,
     The MotorController will use the MotorControl data to correctly drive a motor...

src/main/java/org/myrobotlab/service/abstracts/AbstractMotorController.java
supertick: I Agree,  I just havent done it yet, and I "when its satisfactory" I'll promote them to the interface

src/main/java/org/myrobotlab/service/abstracts/AbstractMotorController.java
supertick: As a basic building block ... an end user I hope will become accustom to "attaching" services..
This is a "routing" attach - which the implementation or abstract will "route" to the appropriate attach.
The implementation, or abstract can choose to implement more class definition parameters, or interfaces as it evolves.  Imagine how legos have an almost universal interface, even if what you stick together doesn't make sense to someone else - e.g. my lego build looks like a pile of plastic vs a helicopter

src/main/java/org/myrobotlab/service/interfaces/MotorControl.java
supertick: setPowerLevel did the "calculation" of data in the MotorControl - which is incorrect. 
The calculation of doing what a MotorController needs to do to provide the final data sent to the hardware is ONLY in the MotorController..

The MotorControl only provides its data for the MotorController to operate on.
As a MotorControl I want to control a motor with the move functions, this function is a MotorController function

src/main/java/org/myrobotlab/service/abstracts/AbstractMotorController.java
supertick: maybe ... this shows a nice red glowing error bar in the gui, which might be nicer to teaching a new user that continuous dc motors without feedback cannot be movedTo

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
kwatters: So, if i want to swap in a new non-linear mapper, how would I do that ?  
ultimately, I want to just say MotorContol.setMapper(myMapper)  and it should work.
the interface really only needs to know about calcOutput ...

kwatters:
FYI I'm good with these changes.. though I think we should revisit the "mapper" in ogre.  I do really like the idea of having that being pluggable.  At a minimum to provide linear / log / exponential mappings and probably a few other fancier ones likes sigmoid or a piece-wise linear mapping to account for a dead zone.

PR #250, opened 2018-08-16
AbstractMotorController and Control refactored with new MapperInterface

supertick:
AbstractMotorController and AbstractMotorControl refactored again to use a new MapperInterface which currently support MapperLinear.

Supports First Law of MotorControl 

- A MotorController will provide a cloned default mapper if the MotorControl does not come with one.. if the MotorControl comes with its "own" mapper - the null values of this mapper are merged with the "default" MotorController mapper (during attaching) - but ALWAYS the MotorControl's mapper (not the MotorController's) is used to calculate the value INSIDE the MotorController !

Stubbed out 2 functions in Adrduino & Adafruit16CServoDriver and did a "basic" implementation since these cannot extend from AbstractMotorController (or they currently don't)

All this gets much more simple when Servo is considered a  MotorControl and ServoControllers are just MotorController ... it would be nice if someday this would happen


src/main/java/org/myrobotlab/math/MapperLinear.java
```diff
@@ -0,0 +1,237 @@
...
+    if (minInput != null && in < minInput) {
+      log.warn("clipping input {} to {}", in, minInput);
+      in = minInput;
+    }
+    if (maxInput != null && in > maxInput) {
+      log.warn("clipping input {} to {}", in, maxInput);
+      in = maxInput;
+    }
+    
+    double c = minY + ((in - minX) * (maxY - minY)) / (maxX - minX);
```
kwatters: No protection against divide by zero here..

src/main/java/org/myrobotlab/math/MapperLinear.java
```diff
@@ -0,0 +1,237 @@
...
+  public double getMinY() {
+    return minY;
+  }
+
+  public boolean isInverted() {
+    return inverted;
+  }
+
+  public void setInverted(boolean invert) {
+    // change state ONLY if needed
```
kwatters: This should just set a boolean flag , and that flag should be honored in the calcOutput method.

src/main/java/org/myrobotlab/math/MapperLinear.java
```diff
@@ -0,0 +1,237 @@
...
+    if (minInput == null) minInput = other.minInput;
+    if (maxInput == null) maxInput = other.maxInput;
+    if (minOutput == null) minOutput = other.minOutput;
+    if (maxOutput == null) maxOutput = other.maxOutput;
+    } else {
+      log.error("don't know how to merge %s into a MapperLinear mapper", in.getClass().getCanonicalName());
+    }
+  }
+
+  public static void main(String[] args) {
```
kwatters: move this to a unit test :)

src/main/java/org/myrobotlab/math/MapperLinear.java
```diff
@@ -0,0 +1,237 @@
...
+    this.maxInput = maxInput;
+  }
+
+  public String toString() {
+    return String.format(" < i < %.2f => map(%.2f,%.2f,%.2f,%.2f) => %.2f < o < %.2f ", minInput, maxInput, minX, maxX, minY, maxY, minOutput, maxOutput);
+  }
+
+  public void map(double minX, double maxX, double minY, double maxY) {
+    if (maxY - minY == 0) {
+      log.error("maxY - minY == 0 ..  DIVIDE BY ZERO ERROR COMING !");
```
kwatters: This protects against divide by zero if this mapper has had map called on it.  (individually setting max/min values would skip this protection.

src/main/java/org/myrobotlab/math/MapperInterface.java
```diff
@@ -0,0 +1,27 @@
+package org.myrobotlab.math;
+
+public interface MapperInterface {
+
+  double calcOutput(double in);
```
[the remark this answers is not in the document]
supertick: I think they got promoted up when the initial copy came from Mapper ..
I'll delete them as this is "net new" stuffs

src/main/java/org/myrobotlab/math/MapperInterface.java
```diff
@@ -0,0 +1,27 @@
...
+
+  int calcOutputInt(int in);
+
+  int calcOutputInt(double in);
+
+  double getMinOutput();
+
+  double getMaxOutput();
+
+  void setMinMaxOutput(double minOutput, double maxOutput);
```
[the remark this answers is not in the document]
supertick: ok ... for symmetry sake

src/main/java/org/myrobotlab/math/MapperLinear.java
supertick: Heh .. there is in the sense of a throw ;)
But sure I'll do a check with error event

kwatters: I think logging an error/warning  and ultimately, return "0.0"  ... in that it's *safe*.. if you divide by zero.. motor output should be 0.   you can't return null with this method signature..   (alternatively, we could throw an exception, but I feel like an uncaught exception like this breaks people scripts sometimes.)

src/main/java/org/myrobotlab/math/MapperLinear.java
supertick: GAH !!!! .. copied over from original Mapper ... scraping clean ..

supertick: actually .. I'll need to look at this .. one of the reasons this is messy, is because minY maxY were also used to describe minInput & maxInput ..  which has now been seperated..   minX maxX minY maxY are "now" ONLY describing  range mapping - NOT CLIPPING ...  Since they are separate we can choose what we want inverted.  One previous problem was min/max  was destroyed when a new range map was specified - which we certainly do not want to do.

src/main/java/org/myrobotlab/math/MapperInterface.java
```diff
@@ -0,0 +1,27 @@
+package org.myrobotlab.math;
+
+public interface MapperInterface {
```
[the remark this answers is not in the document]
supertick: moved it to  org.myrobotlab.math.interfaces and changed it to Mapper name .. this will hold it until the original math.Mapper is refactored away

src/main/java/org/myrobotlab/math/MapperLinear.java
supertick: Divide by zero should be protected where its calculated.
Additionally, the old mapper has "partially" separated clipping values minOutput & maxOutput but the minX and maxX values were overloaded to be clipping input :(

So, it should be one way or the other.  minY maxY should be used as clipping out values, or minX, maxX should have separate clipping input values.  The way it currently is in the original Mapper is least desirable.

I decided to have all separate
- min/max Input
- map values
- min/max Output

This allows the possibility of the relationship of ranges to be modified differently than "clip" values.

src/main/java/org/myrobotlab/math/MapperLinear.java
```diff
@@ -167,13 +177,18 @@ public Double getMaxInput() {
    * @see org.myrobotlab.math.MapperInterface#setMinMaxInput(double, double)
    */
   @Override
-  public void setMinMaxInput(Double minInput, Double maxInput) {
-    this.minInput = minInput;
-    this.maxInput = maxInput;
+  public void setMinMaxInput(Double min, Double max) {
+    if (min != null && max != null && max < min) {
```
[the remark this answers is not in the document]
supertick: Double is needed when you want to merge 2 mappers.
It seems to fit where the MotorControl has input set and output unset,
and the Controller has output set and no input.
When they attach - a complete mapper is created and used.
I think the design is general enough and avoids the over-complex situation of mappers within mappers f(g(x)) .. instead its a merge and only f(x) exists

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
```diff
@@ -109,6 +109,10 @@
   public AbstractMotor(String n) {
     super(n);
     subscribe(Runtime.getInstance().getName(), "registered", this.getName(), "onRegistered");
+    // "top" half of the mapper is set by the control
+    // so that we "try" to maintain a standard default of -1.0 <=> 1.0 with same input limits
+    // "bottom" half of the mapper will be set by the controller
+    mapper.map(-1.0, 1.0, null, null);
```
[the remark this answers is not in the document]
supertick: as a "user" yes ..
as a general MotorControl I think it would be valid, its giving the back end map to the motor controller

src/main/java/org/myrobotlab/math/MapperLinear.java
```diff
@@ -136,9 +141,14 @@ public Double getMaxOutput() {
    * @see org.myrobotlab.math.MapperInterface#setMinMaxOutput(double, double)
    */
   @Override
-  public void setMinMaxOutput(Double minOutput, Double maxOutput) {
-    this.minOutput = minOutput;
-    this.maxOutput = maxOutput;
+  public void setMinMaxOutput(Double min, Double max) {
+    if (min != null && max != null && max < min) {
```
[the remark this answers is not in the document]
supertick: will be changing the name to setOutputLimits(Double p0, Double p1) ..
and removing individual setMin/setMax methods (all 4)
setOutputLimits(-1, 3) will be valid, so will setOutputLimits(3, -1)   - internally they will be the same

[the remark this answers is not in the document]
kwatters: I'm ok with that approach, however, I do worry that automatically flipping the min/max values on input will yield confusing side effects.  (I hope I'm wrong about that.)

src/main/java/org/myrobotlab/math/MapperLinear.java
```diff
@@ -18,13 +18,11 @@
   Double minY;
   Double maxY;
 
-  // clipping
-  Double minOutput;
-  Double maxOutput;
+  // limits
```
kwatters: Doesn't dropping the word "output" from these variables make it more confusing?  I feel like we should have 4 parameters for the mapper.   inputMin, inputMax, outputMin, outputMax ...  There have been many times when people were confused about input vs output max/min settings for mappings..

src/main/java/org/myrobotlab/math/MapperLinear.java
```diff
@@ -115,80 +99,29 @@ public void setInverted(boolean invert) {
...
   }
 
-  /*
-   * (non-Javadoc)
-   * 
-   * @see org.myrobotlab.math.MapperInterface#setMinMaxInput(double, double)
-   */
   @Override
-  public void setMinMaxInput(Double min, Double max) {
+  public void setLimits(Double min, Double max) {
```
kwatters: Are we no longer caring out input limits vs output limits?

src/main/java/org/myrobotlab/math/MapperLinear.java
supertick: Done ...  and much more put into the unit test MapperLinearTest.java

src/main/java/org/myrobotlab/math/MapperLinear.java
supertick: After thinking about it for a while, there is no need for input limits and output limits - If the point is to calculate output, output limits are only needed.  

mapper1
This is a minimal mapper

I think there was way more confusion about the composite of mappers F(G(x)) - which the First Law of MotorControl has removed (Yay!)
mapper2

This is the composite, and it doesn't work and I would say is responsible for 90% of the confusion - its immensely more simple if some default values are put into the MotorControl mapper and it is used directly.

src/main/java/org/myrobotlab/math/MapperLinear.java
supertick: no the world has become much more simple

src/main/java/org/myrobotlab/math/MapperLinear.java
```diff
@@ -53,11 +53,12 @@ final public Double calcOutput(Double in) {
...
 
-      if (!inverted) {
-        c = minY + ((in - minX) * (maxY - minY)) / (maxX - minX);
-      } else {
-        c = minY + ((in - maxX) * (maxY - minY)) / (minX - maxX);
-      }
+      c = minY + ((in - minX) * (maxY - minY)) / (maxX - minX);
+      
+      if (inverted) {
+        c = c * -1;
```
kwatters: I don't think inverted means multiply by -1.  if output limit is from 0 to 5.   and we invert the control.  let's say "c" = 2.0 ... if we invert, I think the expected behavior should be 3.0  .. not -2.0

kwatters: Think about the normal case to be   "the output is y above the minY"
Then the inverted case is  "the output is y below the maxY"  ...

supertick: Thanks, excellent catch - the devil made me do that "optimization"
I'll put in an "asymmetrical"  invert junit test.
Keep the tomatoes coming !
local_08_temp-1346310620-503f11dc-620x348

PR #251, opened 2018-08-18
maybe easy implementation of festival for raspi (and other Linux)

supertick:
LocalSpeech uses festival for Linux


supertick:
merging small changes to support Xrobots desire to use festival on Raspi in LocalSpeech

Issue #256, opened 2018-09-10
Virtual inmmov seem borked


moz4r:
tested from 2 computers, not sure if it is me or not.
test with full and also service test script.
JME windows stay black and stucked, that crash mrl.

also simple command crash :

jmonkeyengine = Runtime.start('jmonkeyengine','JMonkeyEngine')

moz4r:
At first glance, worky inside eclipse, noworky from standalone jar

moz4r:
dependencies issue ! need to find it ...

moz4r:
ok that was a very strange bug. race condition about jme3 audio engine and maryTts. Maybe not the only thing but worky now like before.

hairygael:
testing version 1.1.157 (although the servos cannot be testing proprerly because of autoEnable refactor)
I realize Virtual InMoov is borked.
It just doesn't move and stands in a frozen position. This must be related to a lot of refactoring with virtual stuff.

supertick:
![gestures](https://user-images.githubusercontent.com/1266144/128637054-3d956fcd-f48d-4c27-b3d9-801be1246760.gif)
If started from the InMoov2 UI - it appears to work..
JMonkeyEngine will also start by just running
jme = runtime.start('jme','JMonkeyEngine')
But in this case it will not load all InMoov models etc - as this is not inherently JMonkeyEngine's design.
It can be done manually with more code if desired though.

Setting this to verify - will close if no activity.

hairygael:
I just verified and it is okay and worky with version 534

supertick:
closing

PR #257, opened 2018-09-11
Fix race condition for jme3 / virtual inmoov

moz4r:
https://github.com/MyRobotLab/myrobotlab/issues/256

### DIFF
--- a/src/main/java/org/myrobotlab/service/InMoov.java
+++ b/src/main/java/org/myrobotlab/service/InMoov.java
@@ -2053,6 +2053,7 @@ public class InMoov extends Service {
       // settings.setEmulateMouse(false);
       // settings.setUseJoysticks(false);
       settings.setUseInput(true);
+      settings.setAudioRenderer(null);
       vinMoovApp.setSettings(settings);
       vinMoovApp.setShowSettings(false);
       vinMoovApp.setDisplayStatView(debugVinmoov);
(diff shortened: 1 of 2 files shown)


kwatters:
What was the actual error message?  If someone uses maryspeech will vinmoov be broken? Maybe the actual fix should be around the dependency management?

LunDev:
@kwatters 
Copied this from the shoutbox:

> moz4r: the question is why now ... but I think we don't use audio in jme, for now, can disable it
> moz4r: progress
> moz4r: bug I think there is a fuckin conflict between marytts engine and jme3 audio engine

I couldn't get vInMoov worky either when I tried it a few days ago. But as this was my first time trying to use it, I thought something in my setup was borked and didn't look any further.
Disabling the audio for JME doesn't seem to bad (as far as I know it wasn't used anywhere?).

kwatters:
Any idea what the underlying dependency issue was ?  It'd be nice to know if we're pulling in the wrong jar somewhere.  It's possible that one of the MarySpeech voices is pulling in a conflicting dependency somewhere.  I don't know.  If you have an error message/stack trace, especially if it shows a ClassNotFound or a NoSuchMethod exception..  That would be good to know about because there is likely a deeper issue if that's the case.

kwatters:
@moz4r  Nice fix!  We should delete this branch now that it's merged.

PR #261, opened 2018-09-12
 fix captureGesture + programAb session separator

moz4r:


src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1897,9 +1898,13 @@ public static void main(String[] args) {
       vleft.connect("COM3");
       vright.connect("COM4");
       Runtime.start("gui", "SwingGui");
-
+      Runtime.start("python", "Python");
       InMoov i01 = (InMoov) Runtime.start("i01", "InMoov");
       i01.startAll(leftPort, rightPort);
+      i01.moveArm("left", 20.0, 10.0, 5.0, 40.0);
```
kwatters: Why is this code here?  It doesn't seem like we should be moving the arm here.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -354,8 +355,11 @@ public String captureGesture() {
 
   public String captureGesture(String gestureName) {
     StringBuffer script = new StringBuffer();
+    DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
```
kwatters: It's better to create this object once as a private member variable of the class, rather than re-creating it each time it's being used.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1897,9 +1898,13 @@ public static void main(String[] args) {
       vleft.connect("COM3");
       vright.connect("COM4");
       Runtime.start("gui", "SwingGui");
```
kwatters: Why are we starting the swing gui here?  (also python?)

src/main/java/org/myrobotlab/service/InMoovArm.java
```diff
@@ -203,7 +204,7 @@ public Servo getRotate() {
   }
 
   public String getScript(String inMoovServiceName) {
-    return String.format("%s.moveArm(\"%s\",%d,%d,%d,%d)\n", inMoovServiceName, side, bicep.getPos(), rotate.getPos(), shoulder.getPos(), omoplate.getPos());
+    return String.format((Locale)null,"%s.moveArm(\"%s\",%.2f,%.2f,%.2f,%.2f)\n", inMoovServiceName, side, bicep.getPos(), rotate.getPos(), shoulder.getPos(), omoplate.getPos());
```
kwatters: casting a null value to a locale object here doesn't make any sense.  Also, why are we limiting the resolution to 2 decimal places for the move command?

src/main/java/org/myrobotlab/service/InMoovEyelids.java
```diff
@@ -173,7 +174,7 @@ public long getLastActivityTime() {
   }
 
   public String getScript(String inMoovServiceName) {
-    return String.format("%s.moveEyelids(%d,%d)\n", inMoovServiceName, eyelidleft.getPos(), eyelidright.getPos());
+    return String.format((Locale)null,"%s.moveEyelids(%.2f,%.2f)\n", inMoovServiceName, eyelidleft.getPos(), eyelidright.getPos());
```
kwatters: I think this string format line is wrong.. it should be the string that you're formatting , followed by the arguments to fill that string.  Here you're casting a null value to locale..  this code is not correct.

src/main/java/org/myrobotlab/service/InMoovHand.java
```diff
@@ -278,7 +280,7 @@ public long getLastActivityTime() {
   }
 
   public String getScript(String inMoovServiceName) {
-    return String.format("%s.moveHand(\"%s\",%d,%d,%d,%d,%d,%d)\n", inMoovServiceName, side, thumb.getPos(), index.getPos(), majeure.getPos(), ringFinger.getPos(), pinky.getPos(),
+    return String.format((Locale)null,"%s.moveHand(\"%s\",%.2f,%.2f,%.2f,%.2f,%.2f,%.2f)\n", inMoovServiceName, side, thumb.getPos(), index.getPos(), majeure.getPos(), ringFinger.getPos(), pinky.getPos(),
```
kwatters: same.. as above.

src/main/java/org/myrobotlab/service/InMoovHead.java
```diff
@@ -171,7 +175,7 @@ public long getLastActivityTime() {
   }
 
   public String getScript(String inMoovServiceName) {
-    return String.format("%s.moveHead(%d,%d,%d,%d,%d,%d)\n", inMoovServiceName, neck.getPos(), rothead.getPos(), eyeX.getPos(), eyeY.getPos(), jaw.getPos(), rollNeck.getPos());
+    return String.format((Locale)null,"%s.moveHead(%.2f,%.2f,%.2f,%.2f,%.2f,%.2f)\n", inMoovServiceName, neck.getPos(), rothead.getPos(), eyeX.getPos(), eyeY.getPos(), jaw.getPos(), rollNeck.getPos());
```
kwatters: please fix the string formatting here. this is wrong.

src/main/java/org/myrobotlab/service/InMoovTorso.java
```diff
@@ -175,7 +177,7 @@ public long getLastActivityTime() {
   }
 
   public String getScript(String inMoovServiceName) {
-    return String.format("%s.moveTorso(%d,%d,%d)\n", inMoovServiceName, topStom.getPos(), midStom.getPos(), lowStom.getPos());
+    return String.format((Locale)null,"%s.moveTorso(%.2f,%.2f,%.2f)\n", inMoovServiceName, topStom.getPos(), midStom.getPos(), lowStom.getPos());
```
kwatters: also incorrect.

src/main/java/org/myrobotlab/swing/PythonGui.java
```diff
@@ -223,6 +223,10 @@ public EditorPanel addNewEditorPanel(Script script) {
 		scripts.put(script.getName(), panel);
 		return panel;
 	}
+	
+	public void onAppendScript(Script script) {
```
kwatters: This method will cause an infinite loop.   Please remove or fix.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -309,7 +309,7 @@ public Response getResponse(String username, String text) {
   }
 
   public String resolveSessionKey(String username, String botname) {
-    return username + "-" + botname;
+    return username + "_" + botname;
```
kwatters: Generally this looks ok, but a locale string could contain an underscore in it.  (That was the root motivation for making this change.)  We should choose a safer character for the session key.    It should be a character that will never show up in the username or the bot name.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -617,7 +617,7 @@ public void savePredicates() throws IOException {
     for (String session : sessions.keySet()) {
 
       // TODO: better parsing of this.
-      String[] parts = session.split("-");
+      String[] parts = session.split("_");
```
kwatters: Rather than fixing it everywhere else, why not just address the TODO here...

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: main() section for testing

src/main/java/org/myrobotlab/service/InMoovArm.java
kwatters: "(Locale)null,"  should be removed.  If you are passing a locale in there, that's fine. but you're not.  It's just being passed as null.

src/main/java/org/myrobotlab/swing/PythonGui.java
kwatters: This method calls itself .. that's a recursive method and is most certain to crash.  This is dangerous code and should be removed.

src/main/java/org/myrobotlab/service/InMoovArm.java
kwatters: You should pass something like  Runtime.getLocale() instead of null.

src/main/java/org/myrobotlab/service/ProgramAB.java
kwatters: The idea is that the session should have a proper encoding / decoding so it doesn't have to do a blind splitting on a character.  The splitting character should be escaped and obeyed.

moz4r: ok will find Something cleaner

src/main/java/org/myrobotlab/service/InMoovArm.java
moz4r: welcome Locale.English

src/main/java/org/myrobotlab/swing/PythonGui.java
moz4r: done & worky

kwatters:
The session key is the concatenation of the botname and the username.  This is used as a key to ensure the response comes from the right chatbot.   Initially a dash was used because no one was expected to use a dash in the chat bot name.  If we want to continue to do this we need to ensure that the dash charactwr

kwatters:
Sorry didn't mean to close the request.  But if you are trying to fix the session key encoding we shouldn't just push the problem around to a different character.  The string should be encoded and decoded correctly.

kwatters:
Moz4r,  I think you're making this fix way harder than it needs to be.  Please do not merge this change.
Let's take a step back and see why this is actually easy to fix.

1. you get a variable called "session" ...
2. "session" is the current user name and the current bot name.
3. you already have these variables available for calling getResponse..
4. don't parse the session variable.. just use the currentUsername and the currentBotname variables that are already available in the scope.    avoid needing the session variable at all.

So,,  you'll see the session variable isn't actually needed at all!  much better to pass the username and bot name into the methods and then there is no problem :)

Issue #258, opened 2018-09-12
Format chatbot UI


moz4r:
just to make things a little clearer and tidy

![image](https://user-images.githubusercontent.com/18124594/45417138-c495fb00-b681-11e8-98e8-72ce761fe093.png)

supertick:
I think it should follow a regular chatbox format ...
e.g.
**Anthony** :  Hello bot
**en-US** : Hello Anthony, its great seeing you.
**Anthony** : What is going on ?
**en-US** :  The weather is 30 degrees Celsius and slight winds from the northeast.

LunDev:
either that or the aligned version: e.g. align the bot's message to the left and the user's to the right
@supertick 's version is easier to implement, so I vote for that one :p

moz4r:
great ! we are agree the screenshot is what we have today

moz4r:
more sexy, progress

![image](https://user-images.githubusercontent.com/18124594/45636361-c7d91e80-baa7-11e8-8907-688ac73771d9.png)

hairygael:
I agree as well, Grog's version is tidy.

Gael Langevin
Creator of InMoov
InMoov Robot <http://www.inmoov.fr>
@inmoov <http://twitter.com/inmoov>

PR #265, opened 2018-09-18
ProgramAB UIs format

moz4r:
https://github.com/MyRobotLab/myrobotlab/issues/258


src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -291,9 +296,35 @@ public Response getResponse(String userName, String text) {
...
+
+    conversationHistory.add("<b>" + userName + " : </b>" + text);
+    conversationHistory.add("<i><b>" + botname + " : </b>" + HtmlFilter.stripHtml(response.msg) + "</i>");
+
+    // Queue conversationHistory list to prevent unlimited results..
+    if (conversationHistory.size() > maxConversationHistory)
+
+    {
+      List<String> conversationHistoryQueue = new ArrayList<String>();
+      for (int i = conversationHistory.size() - maxConversationHistory; i < conversationHistory.size(); i++) {
```
LunDev: If I understood this correctly (please correct me if I'm wrong):
We're creating a new list and copying elements to it everytime we make a request to the chatbot (and get an response) (if the conversation is running for 100 interactions).
Sorry, this looks like some (unneeded) overhead.

To start this discussion, some alternative solutions that I can currently think of:
Java has a native type Queue which supports adding in the end and removing in the front.
One List and using remove(0) and add(newelement) - would shift all elements when using an ArrayList!
Or a LinkedList to avoid shifting all elements. Now that I think about it, I think LinkedList implements the Queue interface.
Another solution would be a ring buffer, but that would require to keep track of where it starts and ends - more complex in the guis, might not be the best idea.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -596,7 +631,10 @@ public void reloadSession(String path, String userName, String botName, Boolean
     if (sessions.containsKey(botName) && sessions.get(botName).containsKey(userName)) {
       // TODO: will garbage collection clean up the bot now ?
       // Or are there other handles to it?
-      sessions.remove(sessions.get(botName).get(userName));
+      HashMap<String, ChatData> session = new HashMap<String, ChatData>();
+      session.put(userName, sessions.get(botName).get(userName));
+      log.info("{} session removed", sessions);
+      sessions.remove(botName, session);
```
LunDev: Two comments:
First something minor: the log statement "session removed" is before it's actually removed : )
Second: I would have to check myself, but did you actually verify this works? My brain keeps telling me the session you're trying to remove would have a different ObjectID and be therefore not the same as the one you just created to remove it. (I might be wrong on this one.)

src/main/java/org/myrobotlab/swing/ProgramABGui.java
```diff
@@ -40,9 +40,9 @@
   static final String START_SESSION_LABEL = "Start Session";
   // TODO: make this auto-resize when added to gui..
   private JTextField text = new JTextField("", 30);
-  private JTextArea response = new JTextArea("BOT Response :");
+  private JEditorPane response = new JEditorPane("text/html", "BOT Response :");
```
LunDev: Just a quick note: This compontent doesn't allow it's content to be modified, correct? Can't test right now.
Would otherwise need a readonly flag imo.

src/main/java/org/myrobotlab/swing/ProgramABGui.java
```diff
@@ -210,6 +190,15 @@ public void run() {
...
         botName.setText(programab.getCurrentBotName());
+        textAreacontent = "<font face=\"Verdana\" size=-1>";
+        programab.getConversationHistory().stream().forEach(conversation -> {
+          if (conversation == null) {
+            conversation = "ERROR: NULL";
+          }
+          textAreacontent = textAreacontent + conversation + "<br>";
+
+        });
+        response.setText(textAreacontent);
```
LunDev: Does this reconstruct the entire content if a new interaction is made?
I would have expected there to be a more efficient way (like only adding the new lines).
But I'm ok with this if there's no other easy way.

src/main/resources/resource/WebGui/service/js/ProgramABGui.js
```diff
@@ -29,6 +29,10 @@ angular.module('mrlapp.service.ProgramABGui', [])
         // all service data should never be written to, only read from
         $scope.currentUserName = service.currentUserName;
         $scope.currentBotName = service.currentBotName;
+        $scope.conversation = service.conversationHistory;
+                  for (var i = 0; i < $scope.conversation.length; i++) {
+        $scope.conversation[i] = $sce.trustAsHtml($scope.conversation[i]);
```
LunDev: Uh, this will kill your performance, but as long as there are no other services doing the same, it should be fine.
You're re-trusting all messages everytime a new one arrives.

src/main/java/org/myrobotlab/service/HtmlFilter.java
```diff
@@ -111,7 +111,9 @@ public void setStripHtml(boolean stripHtml) {
   public static String stripHtml(String text) {
     // TODO: something fancier but this works for now.
     String cleanText = text.replaceAll("\\<.*?\\>", " ");
-    cleanText = cleanText.replaceAll("  ", " ");
+    cleanText = cleanText.replaceAll("\\n", " ");
```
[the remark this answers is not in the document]
moz4r: There was some similar code Inside program ab to clean up text, especially the line breaks that make some tts bug. like the idea to clean up all here for all, but can revert

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -291,9 +296,35 @@ public Response getResponse(String userName, String text) {
     // EEK! clean up the API!
     invoke("publishResponse", response);
     invoke("publishResponseText", response);
-    invoke("publishText", response.msg);
+    invoke("publishText", HtmlFilter.stripHtml(response.msg));
```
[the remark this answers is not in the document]
moz4r: yup !

src/main/java/org/myrobotlab/service/ProgramAB.java
moz4r: Oh sometime I like to reinvent the wheel ahaha like it. Tks for info I think it is beter to use a native type Queue

src/main/java/org/myrobotlab/service/HtmlFilter.java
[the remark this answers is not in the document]
LunDev: I'm not saying putting it in an external Util class/method is a bad idea,
I just think it's confusing with this name.
I wouldn't expect a method called stripHTML to remove e.g. linebreaks

src/main/java/org/myrobotlab/service/ProgramAB.java
moz4r: yep log.info need to move down.
Restartsession() was broken, before this fix. because the old session was not deleteted like before. 
because a syntaxerror inside sessions.remove(string,hasmap<>).

src/main/java/org/myrobotlab/service/HtmlFilter.java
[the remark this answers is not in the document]
moz4r: new method cleanHtml() ?

src/main/java/org/myrobotlab/service/ProgramAB.java
LunDev: If you're sure it works now I'm fine.
Thinking of it: Do we have a unit test for this?

src/main/java/org/myrobotlab/service/HtmlFilter.java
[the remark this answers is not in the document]
LunDev: cleanHTML is still about HTML and not really a good descriptor if we're actually just removing linebreaks.
The removal of double whitespaces is fine *imo*.
dunno if a flag removeLinebreaks, a new class, a generalized regex service or a completly different solution would be best. Maybe someone else has a few ideas.

src/main/java/org/myrobotlab/service/ProgramAB.java
kwatters: Program AB should not be doing the HTML filtering here.  That is the job of the HtmlFilter service.

src/main/java/org/myrobotlab/service/ProgramAB.java
kwatters: Have a look at this:  https://stackoverflow.com/questions/5498865/size-limited-queue-that-holds-last-n-elements-in-java/14322473 
we should be using a fixed length queue to maintain conversation history per session.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -291,9 +296,35 @@ public Response getResponse(String userName, String text) {
...
+    Predicates preds = getChat(userName, getCurrentBotName()).predicates;
+    String botname = getCurrentBotName();
+    if (!preds.get("botname").equals("unknown") && !preds.get("botname").isEmpty()) {
+      botname = preds.get("botname");
+    }
+
+    // feed conversation list
+
+    conversationHistory.add("<b>" + userName + " : </b>" + text);
+    conversationHistory.add("<i><b>" + botname + " : </b>" + HtmlFilter.stripHtml(response.msg) + "</i>");
```
kwatters: the conversation history should be a fixed length queue and it should just store the last N response objects.   We shouldn't be mixing html markup with the history here.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -520,15 +551,23 @@ private OOBPayload parseOOB(String oobPayload) {
         ServiceInterface s = Runtime.getService(payload.getServiceName());
         if (s == null) {
           log.warn("Service name in OOB/MRL tag unknown. {}", mrlPayload);
+
+          if (payload.getParams() != null) {
+            conversationHistory.add("<font color='#FF0000'><b> -> Missing OOB : " + payload.getParams() + "</b></font>");
```
kwatters: HTML markup should not be programmatically specified here. conversation history should be fixed length queue of response objects.

src/main/java/org/myrobotlab/swing/ProgramABGui.java
```diff
@@ -194,7 +174,7 @@ public void run() {
           startSessionButton.setText("Start Session");
           startSessionButton.setBackground(Color.RED);
         } else {
-          startSessionButton.setText("Restart Chatbot");
+          startSessionButton.setText("Force reload ( kill csv )");
```
kwatters: End users should have no idea what "kill csv" mean.  Please just label this button  "Restart Chatbot"  (perhaps "Reload Chatbot" .. )   exposing end users to things like needing to "kill csv" is ugly.

src/main/java/org/myrobotlab/swing/ProgramABGui.java
kwatters: This seems wrong.. why aren't we just appending the new data to it?

src/main/resources/resource/WebGui/service/js/ProgramABGui.js
kwatters: Agreed ...  the UI should just be initializing the conversation list on inital load.. then subscribing to every time a new response is published. and appending that specific response.

src/main/java/org/myrobotlab/service/HtmlFilter.java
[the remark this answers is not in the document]
moz4r: optional flag stripHtml(String text, Boolean removeLinebreaks) , seem good solution for now.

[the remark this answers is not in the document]
kwatters: This change is not good.  you've broken the replacement of subsequent white space characters and for some reason removed new lines here.  This is the wrong thing to do.  Please revert this change.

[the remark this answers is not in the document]
kwatters: If we make any changes to HtmlFilter it should be to use a library like jsoup or jericho to strip the html out.

src/main/java/org/myrobotlab/service/HtmlFilter.java
```diff
@@ -111,7 +111,9 @@ public void setStripHtml(boolean stripHtml) {
   public static String stripHtml(String text) {
     // TODO: something fancier but this works for now.
     String cleanText = text.replaceAll("\\<.*?\\>", " ");
-    cleanText = cleanText.replaceAll("  ", " ");
+    cleanText = cleanText.replaceAll("\\n", " ");
+    cleanText = cleanText.replaceAll("\\r", " ");
+    cleanText = cleanText.replaceAll("\\s{2,}", " ");
```
kwatters: Please revert any changes to this service.  If you want to improve the html stripping capabilities, it should be done as an isolated pull request and the change should be to use an official library to do this stripping.

src/main/java/org/myrobotlab/service/ProgramAB.java
kwatters: ProgramAB by definition published HTML, do NOT strip it here!

src/main/java/org/myrobotlab/service/ProgramAB.java
kwatters: Also.. I want to point out that the "Chat" object in the programab session already has history on it!  Don't re-invent the wheel here!

src/main/java/org/myrobotlab/swing/ProgramABGui.java
kwatters: Agreed, this should be immutable .. you should be able to edit this panel.

src/main/java/org/myrobotlab/swing/ProgramABGui.java
```diff
@@ -40,9 +40,9 @@
   static final String START_SESSION_LABEL = "Start Session";
   // TODO: make this auto-resize when added to gui..
   private JTextField text = new JTextField("", 30);
-  private JTextArea response = new JTextArea("BOT Response :");
+  private JEditorPane response = new JEditorPane("text/html", "BOT Response :");
+  String textAreacontent = "";
```
kwatters: This variable shouldn't be here. please remove this line.  This should be local to the method that uses it  (it's not something that should exist in the global context.. because thats "response.getText()"  not this variable ...

src/main/resources/resource/WebGui/service/js/ProgramABGui.js
```diff
@@ -8,16 +8,16 @@ angular.module('mrlapp.service.ProgramABGui', [])
     
     // use $scope only when the variable
     // needs to interract with the display
-    $scope.currResponse = '';
     $scope.utterance = '';
-    $scope.currentText = '';
+    $scope.conversation = [];
```
kwatters: Why are these variables defined here and then again a few lines below?!

src/main/java/org/myrobotlab/service/HtmlFilter.java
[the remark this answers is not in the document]
moz4r: I will revert it, but I don't like to have 2 distant places for complementary things .. https://github.com/MyRobotLab/myrobotlab/blob/develop/src/main/java/org/myrobotlab/service/ProgramAB.java#L574

src/main/java/org/myrobotlab/service/ProgramAB.java
moz4r: I agree AND disagree : we have 3 Publishers . we can give ourselves the luxury that one of them publishes only clean text. please :)

publishResponse > response
publishResponseText > response.msg
publishText > raw text filtered without html tags

kwatters:
ProgramAB publishes "text".  Specifically "text/html"...  Yes,  html is text!  Removing the ability for ProgramAB to publish it's response as is removes functionality.  This is why the HtmlFilter service exists.  Do not remove the ability for ProgramAB to publish it's response as is.

kwatters:
Especially don't change this functionality before Nixie.  If we want to debate changing the behavior of ProgramAB , that is for Ogre...  not Nixie.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -596,7 +631,10 @@ public void reloadSession(String path, String userName, String botName, Boolean
     if (sessions.containsKey(botName) && sessions.get(botName).containsKey(userName)) {
       // TODO: will garbage collection clean up the bot now ?
       // Or are there other handles to it?
-      sessions.remove(sessions.get(botName).get(userName));
+      HashMap<String, ChatData> session = new HashMap<String, ChatData>();
```
kwatters: This should be "sessions.get(botName).remove(userName);"

moz4r:
yep I think also.. progress ! thank you guys

src/main/java/org/myrobotlab/swing/ProgramABGui.java
```diff
@@ -158,23 +188,74 @@ public void actionPerformed(ActionEvent event) {
...
+    unsubscribe("publishRequest");
+    unsubscribe("publishOOBText");
+  }
+
+  public void onText(String text) {
+    SwingUtilities.invokeLater(new Runnable() {
+      @Override
+      public void run() {
+        try {
+          responseKit.insertHTML(responseDoc, responseDoc.getLength(), "<i><b><br>" + botName.getText() + "</b>: " + text.replaceAll("\\<.*?\\>", " ").trim(), 0, 0, Tag.I);
```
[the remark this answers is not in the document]
moz4r: it is hidden :) the last parameter of insertHTML() is used to close tags, useful to save some code line. Tag.I , Tag.[...] . I just discovered that.

src/main/java/org/myrobotlab/swing/ProgramABGui.java
```diff
@@ -158,23 +188,74 @@ public void actionPerformed(ActionEvent event) {
...
+  public void onOOBText(String text) {
+    SwingUtilities.invokeLater(new Runnable() {
+      @Override
+      public void run() {
+        try {
+          String filteredOOB = "";
+          Pattern oobPattern = Pattern.compile("<param>(.+?)</param>");
+          Matcher oobMatcher = oobPattern.matcher(text);
+          filteredOOB = oobMatcher.find() ? filteredOOB = oobMatcher.group(1) : "";
+          responseKit.insertHTML(responseDoc, responseDoc.getLength(), "<font color=blue><b><br> > OOB: </b>" + filteredOOB, 0, 0, Tag.FONT);
```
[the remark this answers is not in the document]
moz4r: you have good eyes ! ">" is nothing , just : 
human: hi
> oob : hi()
bot: hello

src/main/java/org/myrobotlab/swing/ProgramABGui.java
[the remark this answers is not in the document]
LunDev: Are you sure?
I thought it would be both opening&closing tags in addition to Tag.I ?
Maybe I can't remeber how this is used from the last time I used it. Will test later if required.

src/main/java/org/myrobotlab/swing/ProgramABGui.java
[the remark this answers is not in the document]
LunDev: uh, then please encode it properly: `&gt;`
Otherwise it might get misinterpreted.

src/main/java/org/myrobotlab/swing/ProgramABGui.java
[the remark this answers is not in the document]
moz4r: yup in this syntax the last parameter will add a closed I tag

src/main/java/org/myrobotlab/swing/ProgramABGui.java
```diff
@@ -81,6 +84,11 @@ public ProgramABGui(String boundServiceName, SwingGui myService) {
     filter.setSelected(true);
 
     response.setAutoscrolls(true);
+    response.setEditable(false);
+    response.setEditorKit(responseKit);
+    response.setDocument(responseDoc);
+
+    responseKit.insertHTML(responseDoc, responseDoc.getLength(), "<font face=\\\"Verdana\\\">Conversation :<br/>", 0, 0, null);
```
LunDev: heh, sorry - found another line :P

LunDev: Then I don't think it's working the way you want it to.
I just tested out this:
```
responseKit.insertHTML(responseDoc, responseDoc.getLength(), "<font face=\\\"Verdana\\\">Conversation :<br/>", 0, 0, null);
responseKit.insertHTML(responseDoc, responseDoc.getLength(), "<i><b>" + "botname" + "</b>: " + "text" + "</i>", 0, 0, null);
```
And the resulting html markup is:
```
<html>
  <head>
    
  </head>
  <body>
    <font face="\Verdana\">Conversation :<br></font>
<i><b>botname</b>: text</i>
  </body>
</html>
```

src/main/java/org/myrobotlab/service/HtmlFilter.java
```diff
@@ -111,7 +110,7 @@ public void setStripHtml(boolean stripHtml) {
 
   // helper function to strip html tags. 
   public static String stripHtml(String text) {
-    if (StringUtils.isEmpty(text))
+    if (text.isEmpty())
```
kwatters: This will cause a null pointer exception if text is null.  Please use StringUtils.isEmpty.

src/main/java/org/myrobotlab/service/HtmlFilter.java
```diff
@@ -1,6 +1,5 @@
 package org.myrobotlab.service;
 
-import org.jivesoftware.smack.util.StringUtils;
```
kwatters: We should import this from apache commons lang, not jivesoftware.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -583,6 +579,10 @@ public String publishText(String text) {
     return text;
   }
 
+  public String publishRequest(String text) {
```
kwatters: This will cause a null pointer exception if text is null.   You should not be calling "trim" here.   Please remove the trim() statement.  "publish" methods should not alter the data that they are publishing.

src/main/java/org/myrobotlab/swing/ProgramABGui.java
```diff
@@ -140,36 +143,23 @@ public void actionPerformed(ActionEvent event) {
...
-
-    } else if (o == killAiml) {
       String path = progABPath.getText().trim();
       String user = userName.getText().trim();
       String bot = botName.getText().trim();
       swingGui.send(boundServiceName, "setPath", path);
       swingGui.send(boundServiceName, "reloadSession", path, user, bot, true);
     } else if (o == reloadSession) {
-      swingGui.send(boundServiceName, "setUsername", userName.getText().trim());
+      swingGui.send(boundServiceName, "startSession", userName.getText().trim());
```
kwatters: Start session doesn't actually reload the whole chat bot if the chat bot is already created.  Reload Session should reload the AIML and the entire chat bot brain.

src/main/java/org/myrobotlab/swing/ProgramABGui.java
```diff
@@ -178,23 +168,74 @@ public void actionPerformed(ActionEvent event) {
...
+    unsubscribe("publishRequest");
+    unsubscribe("publishOOBText");
+  }
+
+  public void onText(String text) {
+    SwingUtilities.invokeLater(new Runnable() {
+      @Override
+      public void run() {
+        try {
+          responseKit.insertHTML(responseDoc, responseDoc.getLength(), "<i><b>" + botName.getText() + "</b>: " + text.replaceAll("\\<.*?\\>", " ").trim() + "</i>", 0, 0, null);
```
kwatters: Looks like you're re-inventing html parsing here.  Please use HttpFilter.stripHtml static method.

src/main/java/org/myrobotlab/swing/ProgramABGui.java
```diff
@@ -178,23 +168,74 @@ public void actionPerformed(ActionEvent event) {
...
+    });
+  }
+
+  public void onOOBText(String text) {
+    SwingUtilities.invokeLater(new Runnable() {
+      @Override
+      public void run() {
+        try {
+          String filteredOOB = "";
+          Pattern oobPattern = Pattern.compile("<param>(.+?)</param>");
```
kwatters: Use the oobPattern that is already defined in ProgramAB, don't re-define it here.  This pattern should be defined in 1 place and 1 place only.

src/main/java/org/myrobotlab/swing/ProgramABGui.java
moz4r: you right, button identification problem. tks for pointed it

src/main/java/org/myrobotlab/swing/ProgramABGui.java
moz4r: maybe the name can be confusing, Here I don't want to get oobPattern from service ( and we shouldn't ).
I receive String from OOBText publisher, and just format text for needed informations.

kwatters:
time to merge?

moz4r:
yup, seem ok ...

Issue #264, opened 2018-09-17
Low level Services should be unaware of high level Services


supertick:

supertick:
Some care and consideration should be taken when modifying services which share functionality to other services.  

Arduino, Servo, SwingGui, WebGui are "low level".
InMoov, Tracking, WorkE are "high level" services which use low level services.

What problem is the following code solving ?
https://github.com/MyRobotLab/myrobotlab/blob/develop/src/main/java/org/myrobotlab/service/Arduino.java#L1888-L1897

perhaps we could add the functionality, but minimally - we need to add it in a generalized way.

supertick:
I got this ...
moz4r explained that InMoov before shutdown should set all servos to rest position..

I plan to make the following method in Service
void makeReadyForShutdown()

The default in Service will do nothing, however, a service like InMoov can override it and do everything it needs to before shutting down - for example make all servos rest, and say its going to shutdown .  

Initially, I'll keep it simple and this will be driven by Runtime - you can block, but in the future Runtime might spawn a thread if the service in question is not behaving, and come back with a hammer...

moz4r:
I like it, I wonder how to be sure arduino is not released before I send instructions to servo ?

supertick:
https://github.com/MyRobotLab/myrobotlab/pull/266

supertick:
fixed

moz4r:
I have some cleanup to to inside inmoov.stopservice()

MatsMRL:
The code that you are refering to is to ensure that higher level services can make some cleanup before being shutdown. For example, if I have a motor runnig, I want to be able to sen a stop command to it, before the Arduino / Serial service stops.

supertick:
Right .. now when Runtime is requested to shutdown, it will call makeReadyForShutdown on each running service.  This will give the opportunity for each service currently running to do what it has to, to safely shutdown (including stopping motors).  

The only additional change I have not pushed to the develop branch is renaming it preShutdown as per suggestion...   I'd figure I'd wait for moz4r to push his changes in first.

MatsMRL:
OK. So how does this work if the Ardino service is shutdown ? How does the Motor service get informed about that event ?

supertick:
The assumption is InMoov wants to shutdown ...
The appropriate way to shutdown everything is to call Runtime.shutdown() 
[...]

PR #266, opened 2018-09-18
Low level Services should be unaware of high level Services #264

supertick:
Merge request so that the shutdown event can notify all services of impending shutdown.
InMoov now can do whatever it needs to before Arduinos and Servos are shutdown.


supertick:
DOH ! .. Thanks MaVo - I forgot to merge this change in...

moz4r:
great ! tks will test asap

Issue #262, opened 2018-09-12
Raspi + I2C


moz4r:
A friend just tell me : NO WORKY, with service script. ( NPE while attach, line 6 )
will try soon, someone use it with nixie ?
https://github.com/MyRobotLab/pyrobotlab/blob/master/service/Adafruit16CServoDriverPi.py

moz4r:
I confirm, something is broken :(

Traceback (most recent call last):
  File "<string>", line 3, in <module>
	at org.myrobotlab.service.RasPi.createI2cDevice(RasPi.java:394)
	at org.myrobotlab.service.RasPi.attachI2CControl(RasPi.java:375)
	at org.myrobotlab.service.Adafruit16CServoDriver.attachI2CController(Adafruit16CServoDriver.java:837)
	at org.myrobotlab.service.Adafruit16CServoDriver.attach(Adafruit16CServoDriver.java:820)
	at org.myrobotlab.service.Adafruit16CServoDriver.attach(Adafruit16CServoDriver.java:805)
	at org.myrobotlab.service.Adafruit16CServoDriver.setController(Adafruit16CServoDriver.java:766)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
java.lang.NullPointerException: java.lang.NullPointerException

	at org.python.core.Py.JavaError(Py.java:552)
	at org.python.core.Py.JavaError(Py.java:543)
	at org.python.core.PyReflectedFunction.__call__(PyReflectedFunction.java:190)
	at org.python.core.PyReflectedFunction.__call__(PyReflectedFunction.java:206)
	at org.python.core.PyObject.__call__(PyObject.java:534)
	at org.python.core.PyObject.__call__(PyObject.java:540)
	at org.python.core.PyMethod.__call__(PyMethod.java:171)
	at org.python.pycode._pyx3.f$0(<string>:3)
	at org.python.pycode._pyx3.call_function(<string>)
	at org.python.core.PyTableCode.call(PyTableCode.java:171)
	at org.python.core.PyCode.call(PyCode.java:18)
	at org.python.core.Py.runCode(Py.java:1614)
[...]

PR #267, opened 2018-09-20
desperate fixes to get raspi + i2c worky again

moz4r:
updated pi4j-core to 1.2-SNAPSHOT and forced wiringPi implementation

https://github.com/MyRobotLab/myrobotlab/issues/262

### DIFF
--- a/src/main/java/org/myrobotlab/i2c/I2CFactory.java
+++ b/src/main/java/org/myrobotlab/i2c/I2CFactory.java
@@ -1,12 +1,21 @@
 package org.myrobotlab.i2c;
 
 import java.io.IOException;
+import java.util.concurrent.TimeUnit;
 
 import org.myrobotlab.framework.Platform;
 
 import com.pi4j.io.i2c.I2CBus;
+import com.pi4j.io.i2c.I2CFactory.UnsupportedBusNumberException;
+import com.pi4j.io.i2c.I2CFactoryProvider;
+import com.pi4j.io.i2c.impl.I2CProviderImpl;
 
 public class I2CFactory {
+  
+  public static final long DEFAULT_LOCKAQUIRE_TIMEOUT = 1000;
+  public static final TimeUnit DEFAULT_LOCKAQUIRE_TIMEOUT_UNITS = TimeUnit.MILLISECONDS;
+  
+  volatile static I2CFactoryProvider provider = new I2CProviderImpl();
 
   /**
    * Create new I2CBus instance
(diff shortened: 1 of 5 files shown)


LunDev:
I guess you already know your change does not compile?

> java/org/myrobotlab/i2c/AdafruitLEDBackpack.java:[57,7] error: AdafruitLEDBackpack is not abstract and does not override abstract method ioctl(long,ByteBuffer,IntBuffer) in I2CDevice

src/main/java/org/myrobotlab/i2c/I2CFactory.java
```diff
@@ -42,7 +51,14 @@ public static I2CBus getInstance(int busNumber) throws IOException {
       // TODO: fix this!!!
       //log.warn("This probable doesn't work for ARM / RasPI now.. update the code!");
       //return I2CBusImpl.getBus(busNumber);
-      return null;
+      //return null;
+      try {
+        return provider.getBus(busNumber, DEFAULT_LOCKAQUIRE_TIMEOUT, DEFAULT_LOCKAQUIRE_TIMEOUT_UNITS);
+      } catch (UnsupportedBusNumberException e) {
+        // TODO Auto-generated catch block
+        e.printStackTrace();
```
supertick: Generated code doesn't always do we want ...
Please change to a logger, because std output doesn't necessarily go to the log appender - also it can not be controlled, filtered or assigned to debug, info, warn or error

catch(UnsupportedBusNumberException e) {
       log.error("getBus threw", e);

The above will send it to the current logger, and also categorize it as error - and if the logger is created as all other loggers in mrl - it will show that this class did the logging, allowing quick identification and more filtering options.

src/main/java/org/myrobotlab/i2c/I2CProxyDeviceImpl.java
```diff
@@ -81,4 +83,16 @@ public int getAddress() {
...
+    // TODO Auto-generated method stub
+    
+  }
+
+  @Override
+  public void ioctl(long arg0, ByteBuffer arg1, IntBuffer arg2) throws IOException {
+    // TODO Auto-generated method stub
+    
+  }
+
```
supertick: are these required from interface changes ?

src/main/java/org/myrobotlab/service/RasPi.java
```diff
@@ -482,7 +487,7 @@ public boolean getWiringPi() {
           pinList.add(pindef);
         }
       } else {
-        log.error("Unknown boardtype {}", SystemInfo.getBoardType());
+        log.error(String.format("Unknown boardtype %s", SystemInfo.getBoardType()));
       }
     } catch (UnsupportedOperationException | IOException | InterruptedException e) {
       // TODO Auto-generated catch block
```
supertick: same logging thing .. if you see these - please replace with 
log.error("some intelligent text", e);  <- this will have some time text and still dump the stack trace

src/main/java/org/myrobotlab/i2c/I2CFactory.java
moz4r: oh, u right, it was laziness will fix it

src/main/java/org/myrobotlab/i2c/I2CProxyDeviceImpl.java
moz4r: not required to get servo worky with wiringPi, but not sure at all about the rest of the features... a raspi guru will look at this certainly in the future

PR #269, opened 2018-09-24
Lloyd

kwatters:
This is a pull request for all the things changed to create the Lloyd service.  Sorry for such a large merge.  It's mostly net-new stuff so I think it's pretty low risk to existing functionality.
This includes a lot of updates for solr, deeplearning4j, the document pipeline/wikipedia indexing, tracking filter...  
Of note, this pull request includes the transferlearning stuff based on a solr result set so you can interactively create a new image recognition model "on-the-fly"...


moz4r:
seem Lloyd just made a BIG level up. just a question, I did not follow everything, why don't you use/mod wikidata service ?

kwatters:
The wikipedia work here is to be able to build our own offline copy of wikipedia that is indexed in solr.  So you can use wikipedia while offline.

PR #275, opened 2018-10-01
fixed #274 Mqtt service - refactor to handle framework Msgs

### LINKED ISSUE #274: Mqtt service - refactor to handle framework Msgs
currently:
 Mqtt is a client that doesnt know how to process framework Messages

desired:
Messages in and out of an Mrl running instance available from any external Mqtt client - including the ability to "subscribe"

supertick:


src/main/java/org/myrobotlab/service/Mqtt.java
```diff
@@ -1,439 +1,538 @@
...
+			// COMMON GATEWAY REGISTERATION AND X-FORWARDED END ----------------
+
+		} catch (Exception e) {
+			error("tried decoding [%] into a message", payload);
+		}
+
+		// are all necessary ?
+		invoke("publishMqttMsg", topic, message.getPayload());
+		invoke("publishMqttMsgByte", message.getPayload());
+		invoke("publishMqttMsgString", new String(message.getPayload()), topic);
```
LunDev: Is there a reason for this?:
The first publish is `topic`, `payload`.
The third one is `payload as string`, `topic`.

src/main/java/org/myrobotlab/service/Mqtt.java
```diff
@@ -1,439 +1,538 @@
...
+		// Get a token and setup an asynchronous listener on the token which
+		// will be notified once the message has been delivered
+		message = new MqttMessage(payload);
+		message.setQos(qos);
+
+		String time = new Timestamp(System.currentTimeMillis()).toString();
+		log.info("Publishing at: " + time + " to topic \"" + topic + "\" qos " + qos);
+
+		// FIXME - see if user context is like a backend id ...
+		client.publish(topic, message, "Pub sample context", this);
```
LunDev: `"Pub sample context"` seems like a weird hardcoded value.

src/main/java/org/myrobotlab/service/Mqtt.java
```diff
@@ -1,439 +1,538 @@
...
+	 * 
+	 * @param topic
+	 *            to subscribe to (can be wild carded)
+	 * 
+	 * @param qos
+	 *            the maximum quality of service to receive messages at for this
+	 *            subscription
+	 */
+	public void subscribe(String topic, int qos) throws MqttException {
+		client.subscribe(topic, qos, "Subscribe sample context", this);
```
LunDev: `"Subscribe sample context"` seems like a weird hardcoded value.

src/main/java/org/myrobotlab/service/Mqtt.java
supertick: are you saying the parameters should be consistent ?  I would agree (although that wasn't part of the pr change, and it wasn't part of the scope of the ticket - but I would agree)
Or are you asking why so many different types ?

src/main/java/org/myrobotlab/service/Mqtt.java
supertick: Aye .. its a "userContext" and it came from a sample, but I'm unclear at the moment of its purpose

src/main/java/org/myrobotlab/service/Mqtt.java
supertick: agreed, but at the moment do not know the use of userContext

supertick:
As for your additional comment, I totally agree. We have mrl_java_formatter.xml in the root of the project.  Its a manual import, and I forgot to do it with this instance of IDE & check in.

I'll add a ticket to do a full reformatting with the standard - as formatting changes spread between different commits are messy and hard to follow #276

src/main/java/org/myrobotlab/service/Mqtt.java
supertick: done

kwatters:
seems ok to me.. good to go.

PR #277, opened 2018-10-01
closes #276 - massive single commit reformat

### LINKED ISSUE #276: Massive Reformat to standardize all Java files
Make sure all java files follow the conventions we have currently checked in the root of the project (mrl_java_formmatter.xml) in a single commit.

supertick:


kwatters:
um.. is this reformatted using the eclipse formatter that is checked in with mrl ?

supertick:
correct...

src/main/java/org/myrobotlab/arduino/BoardInfo.java
```diff
@@ -1,163 +1,138 @@
...
+   * public String getName() { return boardName; }
+   */
+  /*
+   * 
+   * public boolean isUnknown() { return (boardName == null) ||
+   * boardName.equals("unknown"); }
+   */
+  /*
+   * 
+   * public void setType(int boardId) { this.boardId = boardId; switch (boardId)
```
kwatters: this formatting is bogus...

src/main/java/org/myrobotlab/arduino/BoardInfo.java
```diff
@@ -1,163 +1,138 @@
...
+    // valid = true;
+    this.version = version;
+  }
+
+  /*
+   * public boolean isValid() { return valid; }
+   */
+
+  /*
+   * public void setType(String board) { boardName = board; if
```
kwatters: this formatting is not good either.

supertick:
is there a question if I used the correct formatter or if it was effectively applied, or are you saying the one we currently have checked in is no good ?

src/main/java/org/myrobotlab/arduino/BoardInfo.java
LunDev: (I had something like this happen here as well when I used the auto-format. If someone finds a solution, please post.)

supertick:
If the problem is what the auto-formatter does with multi-lined comments - yeah I don't like it either

There is a work-around if you use valid JavaDoc html pre notation
e.g.

/**  <- must have 2nd javadoc *
*<pre>
  
          my indented and formatted code

</pre>
*/

kwatters:
I noticed some case statements that look like they all got smushed onto a single line..  not very readable.
this will also make it so all existing pull requests will need to be merged.

src/main/java/org/myrobotlab/arduino/BoardInfo.java
kwatters: oh. i see this is in a comment. i didn't realize.. it's nearly impossible to know what's in this pull request because of it's size.

src/main/java/org/myrobotlab/arduino/BoardInfo.java
kwatters: i didn't realize this was inside of a comment.

supertick:
I'm fine with squashing it - if other pr's do not finish in sequence, but it should be done before a release certainly ..

kwatters:
i'd like a change like this to go in after we have a stable build. but as far as i can tell, it's been a while since there's been a successful build.  (including unit tests..)

kwatters:
commits like this area always a bit painful.. perhaps it just makes sense to bit the bullet and do it now.
I would like to propose an increase to the max line length for code and comments though..

PR #281, opened 2018-10-03
new service: KeyboardSimulator

LunDev:
This service allows to simulate KeyboardEvents (Key-Presses & Key-Releases) and provides some useful helper-methods (Key-combinations & combined press&release).

If you have an idea for a better name, please change it.
This service currently doesn't have an unit test, because I had no idea how to test it (checking if the key is actually pressed - maybe?).

This PR doesn't target Nixie, maybe Ogre.

Related PR: #282

### DIFF
--- a/src/main/java/org/myrobotlab/service/KeyboardSim.java
+++ b/src/main/java/org/myrobotlab/service/KeyboardSim.java
@@ -1,0 +1,157 @@
+package org.myrobotlab.service;
+
+import java.awt.AWTException;
+import java.awt.Robot;
+import java.awt.event.KeyEvent;
+
+import org.myrobotlab.framework.Service;
+import org.myrobotlab.framework.ServiceType;
+import org.myrobotlab.logging.Level;
+import org.myrobotlab.logging.LoggerFactory;
+import org.myrobotlab.logging.LoggingFactory;
+import org.slf4j.Logger;
+
+/**
+ * @author MaVo (MyRobotLab) / LunDev (GitHub)
+ */
+
+public class KeyboardSim extends Service {
+	static final long serialVersionUID = 1L;
+	static final Logger log = LoggerFactory.getLogger(KeyboardSim.class);
+
+	private Robot robot;
+
+	public KeyboardSim(String n) {
+		super(n);
+	}
+
+	@Override
+	public void startService() {
+		super.startService();
+		try {
+			robot = new Robot();
+		} catch (AWTException e) {
+			log.error("could not create java.awt.Robot");
+			log.error(e.getMessage());
+		}
+	}
+
+	/**
+	 * simulate pressing and releasing a keyboard key after the desired duration
+	 * 
+	 * @param keycode  keycode of key (use KeyEvent.VK_XXX, where XXX is your key)
+	 * @param duration key press duration in milliseconds
+	 */
+	public void pressAndRelease(int keycode, int duration) {
+		press(keycode);
+		try {
+			Thread.sleep(duration);
+		} catch (InterruptedException e) {
+		}
+		release(keycode);
+	}
+
+	/**
+	 * simulate pressing a keyboard key (does NOT release the key!)
+	 * 
+	 * @param keycode  keycode of key (use KeyEvent.VK_XXX, where XXX is your key)
+	 * @param duration key press duration in milliseconds
+	 */
+	public void press(int keycode) {
+		robot.keyPress(keycode);
+	}
+
+	/**
+	 * simulate releasing a keyboard key (does NOT press the key!)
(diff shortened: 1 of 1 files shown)


moz4r:
You can kill two birds with one stone ( not sure for English expression … and I don't like to kill birds ) about junit test : interface with keyboard service

LunDev:
actually tried that, but couldn't get it to run on a system without a keyboard attached. (didn't try for long, might want to look into that again. Thanks!)

btw: I only know this phrase in English. In German we kill flies - not with a stone though :P

src/main/java/org/myrobotlab/service/KeyboardSim.java
```diff
@@ -0,0 +1,157 @@
...
+	public void startService() {
+		super.startService();
+		try {
+			robot = new Robot();
+		} catch (AWTException e) {
+			log.error("could not create java.awt.Robot");
+			log.error(e.getMessage());
+		}
+	}
+
```
supertick: I think we starting Peers in startService is preferable because it give the user an option to do a Runtime.create() without the peers being created.  But I can't think of a reason to delay the creation of the awt Robot.  I'd put it in the constructor.

The error/exception catch you can do  on a single line.  The way you currently have it you want get the stack trace. But if you do it this way :
log.error("could not create java.awt.Robot", e);   you do.

So mebbe get rid of  startService completely & constructor like this 
```java
public KeyboardSim(String n) {
		super(n);
                try {
			robot = new Robot();
		} catch (Exception e) {
			log.error("could not create java.awt.Robot", e);
		}
	}
```
Some might disagree too but I'd catch Exception, as any exception coming from new'ing a Robot would be FATAL although maybe only  AWTException  has been documented...

src/main/java/org/myrobotlab/service/KeyboardSim.java
```diff
@@ -0,0 +1,157 @@
...
+	 */
+	public void pressAndRelease(int keycode, int duration) {
+		press(keycode);
+		try {
+			Thread.sleep(duration);
+		} catch (InterruptedException e) {
+		}
+		release(keycode);
+	}
+
```
supertick: Ya .. the Thread.sleep is a boring exception Zzzzz ..
just use the Service's sleep then no need for exception handling

```java
public void pressAndRelease(int keycode, int duration) {
		press(keycode);
		sleep(duration);
		release(keycode);
	}
```

src/main/java/org/myrobotlab/service/KeyboardSim.java
```diff
@@ -0,0 +1,157 @@
...
+		LoggingFactory.init(Level.INFO);
+
+		Runtime.start("swing", "SwingGui");
+		KeyboardSim keysim = (KeyboardSim) Runtime.start("keysim", "KeyboardSim");
+
+		try {
+			Thread.sleep(1000);
+		} catch (InterruptedException e) {
+		}
+
```
supertick: Service.sleep(1000)

src/main/java/org/myrobotlab/service/KeyboardSim.java
```diff
@@ -0,0 +1,157 @@
...
+	/**
+	 * simulate pressing a keyboard key (does NOT release the key!)
+	 * 
+	 * @param keycode  keycode of key (use KeyEvent.VK_XXX, where XXX is your key)
+	 * @param duration key press duration in milliseconds
+	 */
+	public void press(int keycode) {
+		robot.keyPress(keycode);
+	}
+
```
supertick: Need String functions .. so happy users can 
keyboard.press("A");   !!!!

```java
public void press(String key)
```

supertick:
Great idea MaVo - had some inline comments

src/main/java/org/myrobotlab/service/KeyboardSim.java
LunDev: I will add this, but I think this will result in a huge switch block : )
Will still look for a better solution.

supertick: How about this ?
https://stackoverflow.com/questions/15260282/converting-a-char-into-java-keyevent-keycode

LunDev: ya, something like this, might get funky when trying to support non alpha-numeric characters though - needs testing.

supertick:
Accepting pull request as is - the string interface is just an enhancement if applicable and not messy. 
Thanks for the new Service MaVo !

LunDev:
but but ... what about the 3 other (very valid) points?
Guess they will have to be resolved in the next pr as well.

PR #282, opened 2018-10-03
new service: MouseSimulator

LunDev:
This service allows to simulate MouseEvents (Button-Presses & Button-Releases) and provides some useful helper-methods (Button-combinations & combined press&release) as well as options to move the mouse and scroll the wheel.

If you have an idea for a better name, please change it.
This service currently doesn't have an unit test, because I had no idea how to test it (checking if the button is actually pressed - maybe?).

This PR doesn't target Nixie, maybe Ogre.

Related PR: #281

### DIFF
--- a/src/main/java/org/myrobotlab/service/MouseSim.java
+++ b/src/main/java/org/myrobotlab/service/MouseSim.java
@@ -1,0 +1,180 @@
+package org.myrobotlab.service;
+
+import java.awt.AWTException;
+
+import java.awt.Robot;
+import java.awt.event.InputEvent;
+
+import org.myrobotlab.framework.Service;
+import org.myrobotlab.framework.ServiceType;
+import org.myrobotlab.logging.Level;
+import org.myrobotlab.logging.LoggerFactory;
+import org.myrobotlab.logging.LoggingFactory;
+import org.slf4j.Logger;
+
+/**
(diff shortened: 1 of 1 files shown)


LunDev:
todo: #281 comments, do the same here.

src/main/java/org/myrobotlab/service/MouseSim.java
```diff
@@ -0,0 +1,180 @@
...
+	 */
+	public void pressAndRelease(int keycode, int duration) {
+		press(keycode);
+		try {
+			Thread.sleep(duration);
+		} catch (InterruptedException e) {
+		}
+		release(keycode);
+	}
+
```
supertick: use the Services sleep to keep your code clean
```java
public void pressAndRelease(int keycode, int duration) {
		press(keycode);
                 sleep(duration);
		release(keycode);
	}
```

supertick:
I think it would be nice to have a couple mouse button methods which targeted specific buttons. 
e.g. mouSim.pressButton1(100) mouSim.pressButton1() moutSim.releaseButton1() .. etc.  ..  enums can be a bit of a pain to interface with - if the Service's methods use simple type data, it allows exposure to a variety of interfaces like REST/json or Cli

LunDev:
(an update will follow - working on that one locally already)

PR #285, opened 2018-10-03
Lloyd2

kwatters:
sorry for such an omnibus pull request.  including in this, update for programab (no more aimlif! also beginning of locale specific support there in chatbots.)  Fixing latency issues with mjpeg streamer, a bunch of solr / deeplearning4j work  (for lloyd service.. that's stuff is generally worky, but still work in progress.)

### DIFF
--- a/src/main/java/org/myrobotlab/document/transformer/XPathExtractor.java
+++ b/src/main/java/org/myrobotlab/document/transformer/XPathExtractor.java
@@ -96,8 +96,7 @@ public class XPathExtractor extends AbstractStage {
       try {
         processXml(xml, doc);
       } catch (XPathExpressionException | SAXException | IOException e) {
-        // TODO Auto-generated catch block
-        e.printStackTrace();
+        log.warn("Exception processing xml", e);
         continue;
       }
 
@@ -131,8 +130,7 @@ public class XPathExtractor extends AbstractStage {
     try {
       fstream = new FileInputStream(filename);
     } catch (FileNotFoundException e) {
-      log.warn("XPATH Extractor config file not found: {}" , filename);
-      e.printStackTrace();
+      log.warn("XPATH Extractor config file not found: {}" , filename, e);
       return null;
     }
     DataInputStream in = new DataInputStream(fstream);
--- a/src/main/java/org/myrobotlab/document/transformer/XPathExtractor.java
+++ b/src/main/java/org/myrobotlab/document/transformer/XPathExtractor.java
@@ -68,7 +68,7 @@ public class XPathExtractor extends AbstractStage {
     try {
       builder = factory.newDocumentBuilder();
     } catch (ParserConfigurationException e) {
-      log.warn("Parser configuration error {}", e);
+      log.warn("Parser configuration error.", e);
     }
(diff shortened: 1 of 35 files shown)


supertick:
Cool !
please remove any System.out.println and e.printStackTrace()
other than that looks good ... although I did not see OpenCV fixes/update .. the start thread/stop thread/capture/stopCapture - is a mess, if you don't have fixes for it, I'll fix it up - but I assumed it would be in the pr too ..

kwatters:
The OpenCV fixes were exactly what you had done in the VideoProcessor, nothing further there.  I'll take care of the system.outs..

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -595,13 +595,16 @@ public void reloadSession(String path, String userName, String botName, Boolean
...
+    // DO NOT SAVE THE AIML!  That overwrites any changes people would have made to it!
+    // writeAndQuit(killAimlIf);
     // kill the session
 
     if (sessions.containsKey(botName) && sessions.get(botName).containsKey(userName)) {
       // TODO: will garbage collection clean up the bot now ?
       // Or are there other handles to it?
-      sessions.remove(sessions.get(botName).get(userName));
+      // sessions.remove(sessions.get(botName).get(userName));
+      if (sessions.containsKey(botName))
```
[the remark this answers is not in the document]
kwatters: ah good point. it's an unnecessary check here. you're right.

src/main/java/org/myrobotlab/document/transformer/XPathExtractor.java
```diff
@@ -132,7 +132,7 @@ private void processXml(String xml, Document doc) throws SAXException, IOExcepti
     try {
       fstream = new FileInputStream(filename);
     } catch (FileNotFoundException e) {
-      System.out.println("XPATH Extractor config file not found: " + filename);
+      log.warn("XPATH Extractor config file not found: {}" , filename);
       e.printStackTrace();
       return null;
```
supertick: Stacktrace will go to the log if you 'do not' put a auto formatter bracket for it .. like this
log.warn("XPATH Extractor config file not found: {} " , filename, e);

src/main/java/org/myrobotlab/document/transformer/XPathExtractor.java
```diff
@@ -172,17 +172,15 @@ private void processXml(String xml, Document doc) throws SAXException, IOExcepti
         }
       }
     } catch (IOException e) {
-      System.out.println("IO Exception reading from file " + filename);
-      e.printStackTrace();
+      log.warn("IO Exception reading from file {} : {}" , filename, e);
       // return what we can...
```
supertick: this way only gets e.toString() or e.getMessage()  ... if you don't put the auto formatting bracket you get the whole stack trace .. e.g.

log.warn("IO Exception reading from file {} : " , filename, e);

kwatters: doh!

supertick:
Welcome Lloyd !

Issue #263, opened 2018-09-15
HttpClient versions - stop the madness


supertick:
Currently Runtime meta data is defined as :

meta.addDependency("org.apache.httpcomponents", "httpclient", "4.5.2");

ProgramAB is
    meta.addDependency("org.apache.httpcomponents", "httpclient", "4.5.2");
    meta.addDependency("org.apache.httpcomponents", "httpcore", "4.4.6");

If httpclient is critical to the very core of mrl - then perhaps all version references should be removed - and the Runtime get a "full" description

    meta.addDependency("org.apache.httpcomponents", "httpclient", "4.5.2");
    meta.addDependency("org.apache.httpcomponents", "httpcore", "4.4.6");

Anyone disagree ?

wilcovantoorn:
It Would be the most logical approach.

Op za 15 sep. 2018 om 17:01 schreef GroG <notifications@github.com>:

kwatters:
who ever gets around to it, http client/core should be excluded from all services and included in either runtime or in httpclient.  (my preference would be runtime.. for simplicity)

PR #287, opened 2018-10-04
HttpClient versions - stop the madness #263

supertick:
both http dependencies are in Runtime and become part of the "one jar" so they are not needed as dependencies in other services, and to normalize things I remove them from a bunch of other services

### DIFF
--- a/src/test/java/org/myrobotlab/deeplearning4j/SolrDataSetIteratorTest.java
+++ b/src/test/java/org/myrobotlab/deeplearning4j/SolrDataSetIteratorTest.java
@@ -63,7 +63,7 @@ public class SolrDataSetIteratorTest {
     dl4j = (Deeplearning4j)Runtime.start("dl4j", "Deeplearning4j");
   }
 
-  @Test
+  @Ignore
   public void testSolrTransferLearningVGG16() throws IOException, NoAvailableBackendException, SolrServerException {
     LoggingFactory.init("INFO");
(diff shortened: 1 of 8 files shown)


kwatters:
yeah. that test should be ignored.

src/test/java/org/myrobotlab/deeplearning4j/SolrDataSetIteratorTest.java
```diff
@@ -1,129 +1,129 @@
...
+import org.nd4j.linalg.dataset.api.preprocessor.VGG16ImagePreProcessor;
+import org.nd4j.linalg.factory.Nd4jBackend;
+import org.nd4j.linalg.factory.Nd4jBackend.NoAvailableBackendException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.myrobotlab.logging.LoggingFactory;
+import org.myrobotlab.service.Deeplearning4j;
+import org.myrobotlab.service.Runtime;
+
+// @Ignore
```
kwatters: This should be ignored..

src/test/java/org/myrobotlab/deeplearning4j/SolrDataSetIteratorTest.java
```diff
@@ -1,129 +1,129 @@
...
+public class SolrDataSetIteratorTest {
```
kwatters: This line should be ignored..

PR #288, opened 2018-10-04
Update issue templates

supertick:


supertick:
hmm - interesting ... bug and feature templates are "committed" in source - (this is from github)

LunDev:
What about noWorkies?

supertick:
A noWorky URL in a bug report, or rename a Bug to NoWorky ?
I suspect you can edit this PR - let me know if you cannot.

LunDev:
would suggest that bug reports should (if applicable) either contain a reference to a noWorky or an uploaded myrobotlab.log
Only problem is privacy, logs may contain private information that shouldn't be shared with the world
(btw one of the initial reasons for the issuetracker project I shared a link to some time ago)

PR #296, opened 2018-10-06
fixed screaming inmoov and fixed unbridled frame rate - now ~30fps

supertick:


src/main/java/org/myrobotlab/jme3/InMoov3DApp.java
```diff
@@ -725,6 +718,13 @@ public void simpleUpdate(float tpf) {
 
     }
 
+    // to achieve 30 fps - we sleep 33 ms - the time the user took to update
+    deltaMs = System.currentTimeMillis() - startUpdateTs;
+    sleepMs = 33 - deltaMs;
```
LunDev: Sorry, could you please rephrase _the time the user took to update_?

supertick: updated

LunDev: oh, I didn't wanted you to change the code, I was (and still am) completly fine with that.
I was just curious how you got to the value of 33ms. I just didn't see the connection between 33ms and 30 fps when I made that comment.
Now it's perfectly clear - dunno where my mind was this afternoon - 30 **f**rames **p**er **s**econd, 30 frames per 1000 milliseconds, 33(.!p3) milliseconds per frame

PR #300, opened 2018-10-08
LocalSpeech fix

moz4r:


src/main/java/org/myrobotlab/service/LocalSpeech.java
```diff
@@ -30,7 +31,7 @@
   private static final long serialVersionUID = 1L;
 
   public final static Logger log = LoggerFactory.getLogger(LocalSpeech.class);
-  
+  private String ttsPath = "\"" + System.getProperty("user.dir") + File.separator + "tts" + File.separator + "tts.exe" + "\"";
```
kwatters: add a setter/getter for ttsPath so people can override it if they install tts.exe into a different location.

src/main/java/org/myrobotlab/service/LocalSpeech.java
```diff
@@ -142,6 +142,22 @@ protected void loadVoices() {
...
   }
+  
+  /**
+   * override default tts.exe path
+   * 
+   * @param ttsPath
+   *          - full path to windows tts.exe executable
+   * TODO - override also other os
+   */
+  public void setPath(String ttsPath) {
```
kwatters: This should be name  setTtsPath and getTtsPath to follow the java convention.  Setters and getters are named for the variable that they set and get.

PR #309, opened 2018-10-09
OpenCV filters mods :

moz4r:
Idea, is : 
- If there is a thread running by filter, terminate it , if filter is removed..
- Fix lastSoureKey for filters, it was broken !

Todo : understand why there is still some things in memory after killed the filter ...

### DIFF
--- a/src/main/java/org/myrobotlab/opencv/VideoProcessor.java
+++ b/src/main/java/org/myrobotlab/opencv/VideoProcessor.java
@@ -63,7 +63,7 @@ public class VideoProcessor implements Runnable, Serializable {
   transient public BlockingQueue<Object> blockingData = new LinkedBlockingQueue<Object>();
   private transient FrameGrabber grabber = null;
   public transient Thread videoThread = null;
-  private transient LinkedHashMap<String, OpenCVFilter> filters = new LinkedHashMap<String, OpenCVFilter>();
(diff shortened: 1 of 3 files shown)


src/main/java/org/myrobotlab/opencv/VideoProcessor.java
```diff
@@ -62,7 +63,7 @@
   transient public BlockingQueue<Object> blockingData = new LinkedBlockingQueue<Object>();
   private transient FrameGrabber grabber = null;
   public transient Thread videoThread = null;
-  private transient Map<String, OpenCVFilter> filters = new LinkedHashMap<String, OpenCVFilter>();
+  private transient LinkedHashMap<String, OpenCVFilter> filters = new LinkedHashMap<String, OpenCVFilter>();
   private transient List<OpenCVFilter> addFilterQueue = new ArrayList<OpenCVFilter>();
```
supertick: Map<String, OpenCVFilter> filters = new LinkedHashMap<String, OpenCVFilter>() 
Using Map is preferred.  Map is an "interface" and LinkedHashMap is the implementation. If the code uses the interface, a different implementation can easily be switched

moz4r:
tks for info, will add some stuff to this pr : 
- terminate every orphan threads, not only yolo filter

PR #313, opened 2018-10-11
minimal Isready() checks for some services

moz4r:


src/main/java/org/myrobotlab/service/Polly.java
```diff
@@ -356,7 +356,8 @@ public static void main(String[] args) {
 
   @Override
   public boolean isReady() {
-    return polly != null ? true : false;
+    setReady(polly != null ? true : false);
```
kwatters: the isReady() method should only return   this.ready ...  the logic for   calling   this.ready = true  should be implemented in the code where the polly service is created and started.. not here

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -181,7 +181,7 @@
   /**
    * overload this if your service needs other environmental or dependencies to be ready
    */
-  protected boolean isReady = true;
+  protected boolean ready = true;
```
kwatters: Hmm.. it's a bigger refactor to do this, but  "ready" should probably be defaulted to "false" ..  And it's up to each service to decide when it's ready to set it to true.  (I think this is a bigger discussion that GroG should weigh in on.)

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -462,57 +467,58 @@ public String publishSpeechRequested(String toSpeak) {
...
-   
+
     for (String speak : spokenParts) {
-  
+
       AudioData audioData = null;
       if (speak.startsWith("#") && speak.endsWith("#")) {
-        audioData = new AudioData(System.getProperty("user.dir") + File.separator + "audioFile" + File.separator + "voiceEffects" + File.separator + speak.substring(1, speak.length() - 1) + ".mp3");
+        audioData = new AudioData(
+            System.getProperty("user.dir") + File.separator + "audioFile" + File.separator + "voiceEffects" + File.separator + speak.substring(1, speak.length() - 1) + ".mp3");
```
kwatters: The path for where the cached mp3 files should not be computed here, it should be pulled out as a variable for this class and referenced here.  That should have proper setters / getters on it so that people can configure it as needed.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -901,7 +906,24 @@ public void stop() {
   public String setAudioEffects(String audioEffects) {
     return audioEffects;
   }
-  
+
+  // TODO check also internet for cloud services, not only api
+  @Override
+  public boolean isReady() {
+    String[] kn = getKeyNames();
```
kwatters: This logic should not be here.  when the service is started, is when it should decided that the service is ready.  This should only call the default behavior of returning this.ready  (from the base class.)

src/main/java/org/myrobotlab/service/NaturalReaderSpeech.java
```diff
@@ -54,7 +55,8 @@ static public ServiceType getMetaData() {
     meta.setSponsor("kwatters");
     meta.addPeer("httpClient", "HttpClient", "httpClient");
     meta.addCategory("speech", "sound");
-    meta.setAvailable(true);
+    // moz4r : set to hidden / poc only, unstable ...
+    meta.setAvailable(false);
```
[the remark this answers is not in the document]
moz4r: Just want to hide it for experimental use only.
Because I'm afraid about license. what do you think

src/main/java/org/myrobotlab/service/Polly.java
moz4r: to be, or not to be. @Grog will awnser the question.
I think it need to be false for services that implement it, true by default

src/main/java/org/myrobotlab/framework/Service.java
moz4r: to be, or not to be. @GROG will awnser the question.
I think it need to be false for services that implement it, true by default

src/main/java/org/myrobotlab/service/Polly.java
moz4r: @kwatters said : the isReady() method should only return this.ready ... the logic for calling this.ready = true should be implemented in the code where the polly service is created and started.. not here

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -200,13 +203,14 @@ public Object getVoiceProvider() {
 
   public AbstractSpeechSynthesis(String reservedKey) {
     super(reservedKey);
-    
+    setReady(false);
     audioFile = (AudioFile) createPeer("audioFile");
-    
+
```
supertick: isReady = false;

by doing this the default for all SpeechSynthesis is false ...  maybe its a good default, I guess out of all of them Mary, Local are nearly the only ones self contained

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -423,6 +425,9 @@ public void startService() {
     subscribe(audioFile.getName(), "publishAudioStart");
     subscribe(audioFile.getName(), "publishAudioEnd");
     getVoices();
+    if (voices.size() > 0) {
+      setReady(true);
+    }
   }
```
supertick: ok

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
supertick: I agree - default could be this, but I agree on the setters & getters in the AbstractSpeechSynthesis class

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -462,57 +467,58 @@ public String publishSpeechRequested(String toSpeak) {
...
     // FIXME - in theory "speaking" means generating audio from some text
     // so starting speaking event is when the first audio is "started"
     // and finished speaking is when the last audio is finished
-  
+
     return playList;
   }
-  
+
   /**
```
supertick: Seems like a lot of whitespace modification .. is it because there are spaces on the originals or is this cr/lf mods ?

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -537,31 +543,31 @@ public Long publishGenerationTime(Long timeMs) {
...
+      }
+
+      play(audioData, block);
+    } catch (Exception e) {
+      log.error("could not generate audio", e);
+      error("%s %s", e.getClass().getSimpleName(), e.getMessage());
+    }
+
+    return audioData;
   }
```
supertick: Were there changes here ? - the git compare seems to freak out

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -871,12 +877,11 @@ protected void addVoice(String name, String gender, String lang, Object voicePro
       genderIndex.put(gender, group);
     }
   }
-  
+
```
supertick: are you using the mrl_java_formatter.xml  file in eclipse ?

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
supertick: Right, no need to override - you can check in the constructor and
isReady = true/false depending on logic there

src/main/java/org/myrobotlab/swing/abstracts/AbstractSpeechSynthesisGui.java
```diff
@@ -362,6 +370,12 @@ public void run() {
...
         lastUtterance.setText(speech.getlastUtterance());
+        if (speech.isReady()) {
+          isReadyIcon.setIcon(readyOK);
+        } else {
+          isReadyIcon.setIcon(readyNOK);
+        }
+
         restoreListeners();
       }
     });
```
supertick: cool .. nice enhancement - and in the correct place ;)  
One Abstract to rule them All !

src/main/java/org/myrobotlab/service/NaturalReaderSpeech.java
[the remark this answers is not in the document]
moz4r: yup actually worky but can break overnight, as in the past. 
Just wondering about api license

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
kwatters: I'd prefer to see this pulled out where there is a member for 

private String mp3CacheDir = System.getProperty("user.dir") + File.separator;     that has a getter/setter

and then when creating this AudioData it would use that .like the following 

new AudioData(mp3CacheDir + "audioFile" + File.separator + "voiceEffects" + File.separator + speak.substring(1, speak.length() - 1) + ".mp3")

I also don't like all the speak.substring stuff.. i'm not sure why the whole utterance isn't just hashed and used ...

src/main/java/org/myrobotlab/framework/Service.java
supertick: I changed my mind I prefer "ready" now :)

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
moz4r: Inside the constructor is not enough, There is events on startService() that can take long time for service start. 
Also we need to "recheck" things sometime. by example if user setKeys() after service start. We need to set ready=true. 
setKeys it's not necessarily the only one check, we can have many more

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
moz4r: yup spaces on the originals

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
moz4r: yes..

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
moz4r: ok for setters / getters...
about speak.substring it is related to audioeffects only. 
convention to call it is #R2D2# and file is named R2D2.mp3

src/main/java/org/myrobotlab/swing/abstracts/AbstractSpeechSynthesisGui.java
moz4r: the initial idea was to use a general abstract, for example a little pic in the tab, but that will do for the moment ...

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
moz4r: will change to ready=true

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
moz4r: no change here, except the formatter

src/main/java/org/myrobotlab/service/Polly.java
moz4r: Inside the constructor is not enough, we need to check for valid api keys, after service is started. 
invalid api keys > polly==null

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1458,7 +1458,14 @@ public boolean isRuntime() {
...
-    return isReady ;
+    return ready;
+  }
+
+  protected void setReady(Boolean ready) {
+    if (!ready.equals(this.ready)) {
+      this.ready = ready;
+      broadcastState();
+    }
   }
```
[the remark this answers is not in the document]
moz4r: bye bye setReady ( just wanted to be sure ready status was broadcasted to ui ) .

src/main/java/org/myrobotlab/service/LocalSpeech.java
```diff
@@ -83,8 +87,13 @@ public AudioData generateAudioData(AudioData audioData, String toSpeak) throws I
...
-
-    return new AudioData(localFileName);
+    File fileTest = new File(localFileName);
+    if (fileTest.exists() && fileTest.length() > 0) {
+      return new AudioData(localFileName);
+    } else {
+      error("%s returned 0 byte file !!! - it may block you", getName());
+      return null;
+    }
   }
```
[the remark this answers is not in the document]
moz4r: 0 byte mp3 file is hell ! 
oh, discovered some NPE and errors while stressing those services.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
moz4r: we can add some check later to localspeech / mimic, 
mary need some time to load and ready status is set only when loaded.

supertick:
After reviewing it in detail .. I'm good with it .. Nice work Moz4r !
I agree with the implementation, and variable name.
There still exists possibility of discussion regarding:

-  licensing for NaturalReaderSpeech 
-  if your doing the following for VoiceRss.java
```java
  @Override
  public AudioData generateAudioData(AudioData audioData, String toSpeak) throws Exception {
    if (isReady())
    {
```
then perhaps it should be promoted into AbstractSpeechSynthesis.
- guidelines when to use

Issue #308, opened 2018-10-09
info level logging


moz4r:
info level logging is too much verbose.
myself I tend to use a lot info for events. 
Today we are drowned in the news. A little bit of sorting to debug?

supertick:
Sure ...  sounds good.

moz4r:
![image](https://user-images.githubusercontent.com/18124594/46674323-1deb3e80-cbdc-11e8-85ca-1c798c3860e3.png)

supertick:
merged !

PR #320, opened 2018-10-12
some log.info to log.debug 

moz4r:
maybe subjective…

try to reduce little by little the level of information for INFO

https://github.com/MyRobotLab/myrobotlab/issues/308

### DIFF
--- a/src/main/java/org/myrobotlab/framework/Instantiator.java
+++ b/src/main/java/org/myrobotlab/framework/Instantiator.java
@@ -18,7 +18,10 @@ public class Instantiator {
     try {
       return getThrowableNewInstance(cast, classname, params);
     } catch (Exception e) {
-      log.error("getNewInstance failed for {} {} {}", cast, classname, e.getClass());
+      // if we don't have gui it throw an error...
+      if (!classname.endsWith("Gui")) {
+        log.error("getNewInstance failed for {} {} {}", cast, classname, e.getClass());
+      }
     }
     return null;
   }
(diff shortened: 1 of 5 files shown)


src/main/java/org/myrobotlab/framework/Instantiator.java
```diff
@@ -18,7 +18,10 @@ static public Object getNewInstance(Class<?>[] cast, String classname, Object...
     try {
       return getThrowableNewInstance(cast, classname, params);
     } catch (Exception e) {
-      log.error("getNewInstance failed for {} {} {}", cast, classname, e.getClass());
+      // if we don't have gui it throw an error...
+      if (!classname.endsWith("Gui")) {
+        log.error("getNewInstance failed for {} {} {}", cast, classname, e.getClass());
```
kwatters: Something wonky here.   seems like this should be written like the following:

log.error("getNewInstance failed for {} {}", cast, classname, e);

kwatters: if "classname" is null.. this will cause an error..  please remove the  surrounding "if" statement.

moz4r: agree was tired... an idea to not get errors anymore about missing gui for service and filters?

moz4r: set to warn ?

kwatters: warn is fine..

PR #328, opened 2018-10-27
new service: WebSocketConnector

LunDev:
More puzzle pieces from Eindhoven ...

This service connects to a websocket (server) connection and publishes all received messages while allowing to send messages back.

I'm not quite sure if the dependencies are added in all required places - actually, I'm sure they are not.
Therefore I will use this occasion to ask about the current process. (Where do they need to be added manually, what tools generate which definitions, ...)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1256,6 +1256,24 @@
   </dependency>
 <!-- WebGui end -->
 
+<!-- WebSocketConnector begin -->
+  <dependency>
+    <groupId>javax.websocket</groupId>
+    <artifactId>javax.websocket-api</artifactId>
+    <version>1.1</version>
+  </dependency>
+  <dependency>
+    <groupId>org.glassfish.tyrus</groupId>
+    <artifactId>tyrus-client</artifactId>
+    <version>1.1</version>
+  </dependency>
+  <dependency>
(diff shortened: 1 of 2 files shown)


supertick:
This service seems very Text related - the WebGui uses websockets and serializes all incoming and outbound messages.  Was curious the use cases which your new web socket service will be satisfying vs. using WebGui ? Nice addition !

LunDev:
The webgui starts an internal websocket server.
This service connects to an external websocket server, this way it's possible to connect to e.g. chat systems embedded in web pages, for example the shoutbox @ myrobotlab.org ! (not done that - yet)

supertick:
Ah ... a websocket client ... excellent ! ..
I thought about this and never got around to do it ... although I was looking at creators of the WebGui to support the client side too - https://github.com/Atmosphere/wasync 

I wonder how wasync compares with 
javax.websocket-api (I'm gonna guess Oracle's "spec") &
org.glassfish.tyrus & tyrus-container-grizzly (glassfish implementation) ..

I glanced over grizzly's api - looks pretty user friendly, I know many have been interested in connecting to the shoutbox for a long time (including me) :D

PR #337, opened 2018-11-06
Implement kinematics class inside bodypart based on inmoov arm service

moz4r:
It was missing, just pasted existing things from inmoov arm Inside generic bodypart, for future tests.
I never used it yet, don't know every details…

### DIFF
--- a/src/main/java/org/myrobotlab/service/BodyPart.java
+++ b/src/main/java/org/myrobotlab/service/BodyPart.java
@@ -230,104 +230,113 @@ public class BodyPart extends AbstractBodyPart implements IKJointAngleListener {
   //pasted from inmoov service, for compatibility
   @Override
   public void onJointAngles(Map<String, Double> angleMap) {
-    // We should walk though our list of servos and see if
-    // the map has it.. if so .. move to it!
-    // Peers p = InMoovArm.getPeers(getName()).getPeers("Servo");
-    // TODO: look up the mapping for all the servos in the arm.
+    // todo implement other body parts
+    if (this.getIntanceName().toLowerCase().contains("arm")) {
+      // We should walk though our list of servos and see if
+      // the map has it.. if so .. move to it!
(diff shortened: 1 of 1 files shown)


src/main/java/org/myrobotlab/service/BodyPart.java
```diff
@@ -219,4 +226,108 @@ static public ServiceType getMetaData() {
     meta.addDependency("org.apache.commons", "commons-lang3", "3.3.2");
     return meta;
   }
+
+  //pasted from inmoov service, for compatibility
+  @Override
+  public void onJointAngles(Map<String, Double> angleMap) {
```
kwatters: This method is implemented specifically for the InMoovArm.   It does not generically apply to any "BodyPart".  It should not be copy and pasted here.  Implementing IKJointAngleListener is fine.. but rather than assuming it's an inmoov arm.. log a warning that it's not implemented yet and go on.  (or throw a NotImplementedExeception if you wish.)

src/main/java/org/myrobotlab/service/BodyPart.java
```diff
@@ -219,4 +226,108 @@ static public ServiceType getMetaData() {
...
+        }
+      }
+    }
+  }
+  
+  //pasted from inmoov service, for compatibility
+  public static DHRobotArm getDHRobotArm() {
+
+    // TODO: specify this correctly and document the reference frames!
+    DHRobotArm arm = new DHRobotArm();
```
kwatters: Similar, this is specific to the inmoov arm, and not a generic body part.  Therefore this should return null and log a warning message that it's not implemented.  (Another approach would be to make it abstract and force a subclass to implement it.

moz4r:
Thanks, I can not wait to test ! I'm sure there will be some questions...
Also related to it, I wonder if standardization of gestures by angles without encoder is possible?

PR #339, opened 2018-11-13
Opencv fixes

supertick:


src/main/java/org/myrobotlab/opencv/OpenCVFilterTracker.java
```diff
@@ -131,7 +134,7 @@ public IplImage process(IplImage image, OpenCVData data) {
         float yC = (float) (boundingBox.y() + boundingBox.height()/2);
         Point2Df center = new Point2Df(xC, yC);
         pointsToPublish.add(center);
-        data.set(pointsToPublish);
+        data.put("TrackingPoints",pointsToPublish);
```
kwatters: I'd like to see TrackingPoints pulled out as a constant at the top of the class.  eventually refactored out into an enum that defines all the keys for data.

src/main/java/org/myrobotlab/service/SwingGui.java
```diff
@@ -477,7 +477,11 @@ public void setStatus(Status inStatus) {
       status.setOpaque(false);
     }
 
-    status.setText(inStatus.detail);
+    if (inStatus.detail != null && inStatus.detail.length() > 128) {
+      status.setText(inStatus.detail.substring(128));
```
kwatters: it seems like we should have a different approach to displaying these warnings so they don't resize the swing gui when long errors show up.  Also. the arbitrary length of 128 will still resize on low resolution monitors  (like the raspi touch screen.)

src/test/resources/ProgramAB/bots/pikachu/aiml/pikachu.aiml
```diff
@@ -1,10 +1,8 @@
...
-		<!-- あなたに会えてよかったXXXXX -->
-		<template>あなたに会えてよかった<star /></template>
-	</category>
-</aiml>
\ No newline at end of file
+<?xml version="1.0" encoding="UTF-8"?>
+<aiml>
+<!--  -->
+<!--  -->
+<category><pattern>私 は * です</pattern>
```
kwatters: your auto-formatting actually broke this!  there should not be spaces in the pattern tag here.  Please revert this change.

src/main/java/org/myrobotlab/document/Classification.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  public void setImage(BufferedImage image) {
+    try {
+      ByteArrayOutputStream baos = new ByteArrayOutputStream();
+      ImageIO.write(image, "png", baos);
+      baos.flush();
+      byte[] bytes = baos.toByteArray();
+      baos.close();
+      setField("image", bytes);
```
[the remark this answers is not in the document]
supertick: bytes could be anything - this "is" image data, it seems like knowing that might be helpful

src/main/java/org/myrobotlab/document/Classification.java
```diff
@@ -0,0 +1,91 @@
...
+  public Float getConfidence() {
+    return (Float) getValue("confidence");
+  }
+
+  public void setBoundingBox(int x, int y, int width, int height) {
+    setField("boundingBox", new Rectangle(x, y, width, height));
+  }
+
+  public Rectangle getBoundingBox() {
+    return (Rectangle) getValue("boundingBox");
```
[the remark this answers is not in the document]
supertick: gah .. more mangling .. 
done ..

src/main/java/PrincipalComponentAnalysis.java
```diff
@@ -0,0 +1,182 @@
+import org.bytedeco.javacpp.indexer.DoubleIndexer;
```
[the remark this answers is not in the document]
supertick: ya .. nothing should be in the default package name...
but there are a couple of outstanding items I need to attend to in Ogre relating more to networking ..
This was some javacv samples, which are all dumped into the default package (not sure what your reference is to boofcv) - deleted this one

src/main/java/SimpleOpenNI/SimpleOpenNI.java
```diff
@@ -6,17 +6,17 @@
  * This file is part of SimpleOpenNI.
  *
  * SimpleOpenNI is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
+ * it under the terms of the Apache License 2.0 as published by
```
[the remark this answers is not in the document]
supertick: nice catch - reverted

src/test/resources/ProgramAB/bots/pikachu/aiml/pikachu.aiml
supertick: reverted

src/main/java/org/myrobotlab/service/SwingGui.java
supertick: ya .. but that means someone needs to come up with an alternative. Having a stack trace go through will certainly blow it out at any resolution, so this was a stop gap until someone come up with a better plan.

src/main/java/org/myrobotlab/opencv/OpenCVFilterTracker.java
supertick: I would prefer this be put into a common abstract VideoData base class or interface which OpenCVData derives/implement not an opencv class

src/main/java/org/myrobotlab/opencv/OpenCVFilterSetImageROI.java
```diff
@@ -1,19 +1,19 @@
 /**
```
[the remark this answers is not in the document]
supertick: does something now :)

src/main/java/org/myrobotlab/opencv/OpenCVFilterResetImageRoi.java
```diff
@@ -1,19 +1,19 @@
 /**
```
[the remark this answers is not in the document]
supertick: does something now :)

src/main/java/org/myrobotlab/document/Classifications.java
```diff
@@ -0,0 +1,46 @@
+package org.myrobotlab.document;
```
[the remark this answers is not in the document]
supertick: ya - the basic attempt was to provide some form of new classifications which didn't need pub-sub .. but probably this part could use some re-design

src/main/java/org/myrobotlab/opencv/OpenCVFilter.java
```diff
@@ -127,17 +138,141 @@ public void broadcastFilterState() {
...
+  
+  public void saveToFile(String filename, IplImage image) {
+    opencv.saveToFile(filename, image);
+  }
+  
+  public Mat convertToMat(IplImage copy) {
+    return opencv.convertToMat(copy);
+  }
+  
+  public void show(final IplImage image, final String title) {
```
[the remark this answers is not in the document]
supertick: Ya .. I wanted to make "creating a filter" easy for the developer .. similarly following the design of Service.  So probably a static in ImageUtils would be good (if its thread-safe)..
But this is in the base class already - so any "real" filter would just be calling 

convertToMat(image)  ... maybe shorten it to toMat(image) ..

src/main/java/org/myrobotlab/opencv/OpenCVFilterTracker.java
kwatters: Typically for items like this, we create an enum called FieldNames that every class references for things like this.  It helps flush out a schema.

src/main/java/org/myrobotlab/service/SwingGui.java
kwatters: maybe changing the status bar to be something like a text area with scroll bars and word wrap ?  I'm fine with the truncation now, but this still probably needs to be revisited.

src/test/resources/ProgramAB/bots/pikachu/aiml/pikachu.aiml
kwatters: Thanks, Japanese text does not have spaces between words.  I have to noodle on how a unit test should work to verify that.  Also, if anyone speaks Japanese, can they verify my grammar here.

src/main/java/org/myrobotlab/opencv/OpenCVFilterResetImageRoi.java
```diff
@@ -36,48 +39,21 @@
...
-     * cfg.get(OUTPUT_IMAGE_NAME); if (globalData.containsKey(dstName)) { dst =
-     * (IplImage) globalData.get(dstName); } else { dst = image.clone();
-     * globalData.put(dstName, dst); }
-     * 
-     * } else { dst = src; }
-     * 
-     * // if (cfg.getBoolean(USE_ROI)) // { cvResetImageROI(dst); // }
-     */
-    return image; // TODO - src dst or image? consistency?
+    CvRect r = new CvRect(0, 0, image.width(), image.width());
```
kwatters: seems like this filter should just be calling  cvResetImageROI(image);   No big deal,  this should effectively be the same thing.

PR #353, opened 2018-11-22
kinect related updates

supertick:


kwatters:
Seems good, but I agree.. why pyramid down?

PR #357, opened 2018-11-23
a little kinect depth filter polish

supertick:


src/main/java/org/myrobotlab/opencv/OpenCVFilterKinectDepth.java
```diff
@@ -51,43 +52,40 @@
...
+    OpenKinectFrameGrabber kinect = null;
 
-  public OpenCVFilterKinectDepth() {
-    super();
-    if (opencv != null) {
-      cameraIndex = opencv.getCameraIndex();
+    if (kinect == null && opencv != null) {
+      try {
+        FrameGrabber grabber = opencv.getGrabber();
+        if (!grabber.getClass().equals(OpenKinectFrameGrabber.class)) {
```
kwatters: just curious why not use  "instanceof"  here?

PR #361, opened 2018-11-24
fixed recordSingleFrame / recordFrame

supertick:


src/main/java/org/myrobotlab/jme3/Jme3Servo.java
```diff
@@ -19,19 +18,32 @@
  */
 public class Jme3Servo extends Jme3Object implements VirtualServo {
```
kwatters: Why does this implement virtual servo? seems like it should implement ServoControl ...

src/main/java/org/myrobotlab/kinematics/DHLink.java
```diff
@@ -41,10 +41,10 @@
 
   public transient final static Logger log = LoggerFactory.getLogger(DHLink.class);
   
-  private double velocity;
-  private int state = Servo.SERVO_EVENT_STOPPED;
+  private double velocity; // FIXME - is this set by IK being dp/dt ? .. it should be
```
kwatters: DHLink should not have velocity or state information in it.  I'm not sure where this came from.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1249,7 +1258,7 @@ public InMoovHead startHead(String port, String type, Integer headYPin, Integer
       type = Arduino.BOARD_TYPE_MEGA;
     }
 
-    head.arduino.setBoard(type);
+    // FIXME -  !!! =>  cannot do this "here" ??? head.arduino.setBoard(type);
```
kwatters: we shouldn't allow "setBoard" for the type  perhaps remove this method all together...  when you connect to the arduino. i'm pretty sure it publishes why type of board it is..

src/main/java/org/myrobotlab/service/InMoovHead.java
```diff
@@ -118,7 +119,7 @@ public boolean connect(String port) throws Exception {
   }
   
   public boolean connect(String port,Integer headYPin,Integer headXPin,Integer eyeXPin,Integer eyeYPin,Integer jawPin,Integer rollNeckPin) throws Exception {
-    arduino.connect(port);
+ // FIXME -  !!! =>  cannot do this "here" ??? arduino.connect(port);
```
kwatters: for backwards compatibility.. you should do a   if arduino instanceof Arduino  .. .. then cast it to an arduino and tell it to connect.

src/main/java/org/myrobotlab/service/InMoovTorso.java
```diff
@@ -121,12 +122,15 @@ public boolean connect(String port) throws Exception {
       return false;
     }
 
+    /** FIXME - connections must be done externally - this
+     * needs to be a ServoController
```
kwatters: same issue as above.. test if the arduino is an instance of Arduino .. and then call connect on it.  o/w this isn't backwards compatible.

src/main/java/org/myrobotlab/service/InMoovTorso.java
```diff
@@ -264,10 +268,12 @@ public void setpins(Integer topStomPin, Integer midStomPin, Integer lowStomPin)
     this.lowStom.setPin(lowStom);
     */
 	  
+    /** FIXME - has to be done outside of
```
kwatters: not backwards compatible...   should be refactored to set the pin on the servo and attach it to the arduino.

src/main/java/org/myrobotlab/jme3/Jme3Servo.java
supertick: Yes, I think your right ... (head hurts sometimes)  - but Arduino's "HardwareServo" derives from VirtualServo - are you thinking this should be ServoControl too ?

The challenge with ServoControl is that it's really 2 interfaces - it represents what a user can do and it also represents how a ServoController can interface with it.

src/main/java/org/myrobotlab/service/InMoov.java
supertick: it can't always be trusted ..  I would like it to be external to the InMoov class ..  it would be nice to get away from "over helping" with configuration

src/main/java/org/myrobotlab/service/InMoovHead.java
supertick: ok ..  so there are many things which don't connect, and some things which connect but are not an Arduino .. 
So I think I might have an acceptable solution ..

src/main/java/org/myrobotlab/opencv/ImageFileFrameGrabber.java
```diff
@@ -12,24 +14,62 @@
...
+      // pause for the specified delay before loading the image.
+      Thread.sleep(delay);
+    } catch (InterruptedException e) {}
+    // set the file path
+     * 
+     */
+    path = imageFiles.get(grabCount).getAbsolutePath();
+    log.debug("Grabbing file {} - {}", grabCount, path);
+    // grab it.
     if (!cache.containsKey(path)) {
```
kwatters: This cache is problematic .. if you try to process a large directory with lots of image files, this cache will run us out of memory.  We should re-design this.  (perhaps not for this pull request.. but.. this is definitely not good as it can cause unbounded memory usage.)

src/main/java/org/myrobotlab/service/InMoovTorso.java
```diff
@@ -115,26 +116,21 @@ public void broadcastState() {
...
-
-    if (arduino == null) {
+    if (controller == null) {
       error("arduino is invalid");
       return false;
     }
 
-    /** FIXME - connections must be done externally - this
-     * needs to be a ServoController
+    PortConnector arduino = (PortConnector) controller;
```
kwatters: this seems like an unsafe type cast...

src/test/resources/ProgramAB/bots/pikachu/aiml/pikachu.aiml
```diff
@@ -1,10 +1,8 @@
...
-	<category>
-		<pattern>私は*です</pattern>
-		<!--  あなたを*に会うのはいいことだ   ?? -->
-		<!--  あなたを*に会えてうれしい -->
-		<!-- あなたに会えてよかったXXXXX -->
-		<template>あなたに会えてよかった<star /></template>
-	</category>
-</aiml>
\ No newline at end of file
+<?xml version="1.0" encoding="UTF-8"?>
```
kwatters: Please Revert This Change!

src/main/java/org/myrobotlab/service/InMoovTorso.java
supertick: fixed

src/main/java/org/myrobotlab/service/InMoovTorso.java
supertick: fixed

src/main/java/org/myrobotlab/opencv/ImageFileFrameGrabber.java
supertick: yup .. agree, but this pr did not change the original design .. it has always been this way.
Sure we could make it better - but its currently worky ;)

src/main/java/org/myrobotlab/service/InMoovTorso.java
supertick: fixed

src/test/resources/ProgramAB/bots/pikachu/aiml/pikachu.aiml
supertick: we need to do the data/resource {Service}/{instance} we talked about

supertick:
btw ... its taking a while, but I hope in the future my prs will be much more isolated - apologies for this "Franken-Pull-Request" some of it was fall out from "re-discovering" code I thought was lost - sorry for the chaos

src/main/java/org/myrobotlab/service/InMoov.java
supertick: fixed with instanceof checking

src/main/java/org/myrobotlab/jme3/Jme3Servo.java
```diff
@@ -19,19 +15,32 @@
...
 
+  // FIXME - probably not needed
   private int lastPosUs;
-
+  
+  
   // uber parent of all - moved up
   // transient SimpleApplication app;
+  // FIXME - should allow input of Servo & 3D Info + ref to Jme3App
+  public void someWeirdConstructor(Servo servo, Jme3App app) {
```
kwatters: shouldn't this be a ServoControl  not a Servo ?

supertick: good catch .. fixed

src/test/resources/ProgramAB/bots/pikachu/aiml/pikachu.aiml
supertick: reverted

supertick:
What's left ?

src/main/java/org/myrobotlab/jme3/Jme3Servo.java
kwatters: The way I think of it, the only thing that needs to be virtual is the ServoController...  If you connect a real device to a virtual controller, the device becomes virtual by default... (at least for the servo case, and really anything that attaches to an arduino / i2c bus.)

src/main/java/org/myrobotlab/kinematics/DHLink.java
Calamity01: It his me that add those info needed for collision detection simulation. To be able to do simulation, it need to know the speed (velocity) that this particuliar dhlink can move. I don't think those info affect anything about how IK3D behave. Maybe I should have use other naming to avoid confusion

PR #362, opened 2018-11-25
Load some services from json

moz4r:
It is important for servo service. With it, servo will load based on the last servo position.
it avoids sudden movements at startup, I broke a part without it

### DIFF
--- a/src/main/java/org/myrobotlab/service/ProgramAB.java
+++ b/src/main/java/org/myrobotlab/service/ProgramAB.java
@@ -867,6 +867,12 @@ public class ProgramAB extends Service implements TextListener, TextPublisher {
     meta.addCategory("ai","control");
     return meta;
   }
+  
+  @Override
+  public void startService() {
+    super.startService();
+    load();
+  }
 
   public static void main(String s[]) throws IOException {
     try {
--- a/src/main/java/org/myrobotlab/service/Servo.java
+++ b/src/main/java/org/myrobotlab/service/Servo.java
@@ -636,6 +636,7 @@ public class Servo extends Service implements ServoControl {
   public void startService() {
     super.startService();
     this.addServoEventListener(this);
+    load();
   }
 
   public void rest() {
@@ -1429,8 +1430,8 @@ public class Servo extends Service implements ServoControl {
       // servo.attach(arduino, 8);
       // servo.attach(
       /// servo.attach(adafruit16CServoDriver, 7);
-      servo.moveToBlocking(90);
-      servo.moveToBlocking(30);
+      //servo.moveToBlocking(90);
+      //servo.moveToBlocking(30);
 
       /// servo.enable(9);
       /// servo.moveToBlocking(90);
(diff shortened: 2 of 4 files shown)


kwatters:
This will break if the JSon serialization changes from build to build and as a result will cause services to not start.  I don't think we should do this

moz4r:
Locally tested these specific services load the configuration again as before.
What do you suggest to make it work again without using serialization, also without adding too many duplicates about the variables needed

moz4r:
I am rather in the optics to delete a maximum of things rated python land, but if it is a good practice, will do it. 
Mmm, what do you think of a fresh new function : service.startAndLoad() ?

supertick:
_"This will break if the JSon serialization changes from build to build and as a result will cause services to not start. I don't think we should do this"_  if this is the case, it needs to be fixed - json not loading should not prevent a service from starting .. regardless of what outcome is decided.

@moz4r  - are you proposing   servo = Runtime.startAndLoad("servo","Servo")   ?

kwatters:
do something like this:

service = Runtime.start("serviceName", "ServiceType")
service.load()

moz4r:
I don't like this : 
- not universal, risk of part break if someone forgot it.
- If I use 20 servo, it will cost now 40 total lines to declare the servo service... actual inmoov script is quite overloaded.

use load() from java land is just perfect, maybe with related unit test to secure.
if not possible a thing like Runtime.startAndLoad("servo","Servo") will be a bandage for now...

kwatters:
Override the load() method in the InMoov service to call load on all peers.  Then it's only 1 line.

moz4r:
We don't want to hardcode hardware no more Inside high level, servo peers are going to go away from InMoov service. Things will be attached.
I think this is not a suitable solution. Also for people not using InMoov.

This is not related only to last servo position, also about custom mapings.
Not really a problem for us,but some people used UI only to calibrate servos, and save. that's all. And was happy to discover mrl offer it.

Today mapings are not loaded anymore because there is no backward compatible solution, and replaced by default servo values, that is very bad, exept if we like magic blue smoke.
Personnaly I used it also, outside inmoov. That was fuckin handy ! And maybe it is the direction to go for InMoov calibration process.

I think it's a regression, and is blocking for next uses. If we can't find a good solution I will explicit call load() everywhere Inside python land of course, but I don't like it,at all.

moz4r:
Happy by a solution that makes everyone agree.
Runtime.loadAndStart() to respect life-cycle ?

kwatters:
how would loadAndStart() work in the following 2 scenarios:
1. there is no pre-existing json to load from
2. the pre-existing json is malformed?

moz4r:
About no pre-existing json, default values are used.
2 cases ( based on master branch ),

- variable is not related to json loading, no problem because this initiate the var, with or without json load : https://github.com/MyRobotLab/myrobotlab/blob/master/src/org/myrobotlab/service/Servo.java#L185

- we need to read variable from json, variable is null because : https://github.com/MyRobotLab/myrobotlab/blob/master/src/org/myrobotlab/service/Servo.java#L180

so we do a null check to init the var : https://github.com/MyRobotLab/myrobotlab/blob/master/src/org/myrobotlab/service/Servo.java#L620

About eventual malformation, mmm an idea to check, maybe existing things from json class to warn ? then , cure / restore / delete ?
image

supertick:
New PR ! - https://github.com/MyRobotLab/myrobotlab/pull/368

supertick:
fixed by global consensus

PR #369, opened 2018-11-28
Load gestures

moz4r:
Use class for loadGestures(), no refactor. 
Also add stopGesture() to stop a running gesture ( big red buton ! )

### DIFF
--- a/src/main/java/org/myrobotlab/inmoov/Utils.java
+++ b/src/main/java/org/myrobotlab/inmoov/Utils.java
@@ -2,7 +2,7 @@ package org.myrobotlab.inmoov;
 
 import java.io.File;
 import java.io.IOException;
-
+import org.apache.commons.io.FilenameUtils;
 import org.myrobotlab.io.FileIO;
 import org.myrobotlab.logging.LoggerFactory;
 import org.myrobotlab.service.Python;
@@ -16,21 +16,30 @@ public class Utils {
   /**
    * This method will load a python file into the python interpreter. 
    */
-  public static boolean loadPythonFile(String file) {
+  public static boolean loadFile(String file, String type) {
+    if (!FilenameUtils.getExtension(file).equalsIgnoreCase(type)) {
+      log.warn("{} is not a {} file", file, type);
+      return false;
+    }
     File f = new File(file);
     Python p = (Python) Runtime.getService("python");
     log.info("Loading  Python file {}", f.getAbsolutePath());
+    if (p == null) {
+      log.error("Python instance not found");
+      return false;
+    }
     String script = null;
     try {
       script = FileIO.toString(f.getAbsolutePath());
     } catch (IOException e) {
-      log.error("IO Error loading file");
+      log.error("IO Error loading file : {}",e);
       return false;
     }
     // evaluate the scripts in a blocking way.
     boolean result = p.exec(script, true, true);
     if (!result) {
       log.error("Error while loading file {}", f.getAbsolutePath());
+      return false;
     } else {
       log.info("Successfully loaded {}", f.getAbsolutePath());
     }
--- a/src/main/java/org/myrobotlab/inmoov/Utils.java
+++ b/src/main/java/org/myrobotlab/inmoov/Utils.java
@@ -2,7 +2,6 @@ package org.myrobotlab.inmoov;
 
 import java.io.File;
 import java.io.IOException;
-import org.apache.commons.io.FilenameUtils;
 import org.myrobotlab.io.FileIO;
 import org.myrobotlab.logging.LoggerFactory;
 import org.myrobotlab.service.Python;
@@ -16,11 +15,7 @@ public class Utils {
   /**
    * This method will load a python file into the python interpreter. 
    */
-  public static Boolean loadFile(String file, String type) {
-    if (!FilenameUtils.getExtension(file).equalsIgnoreCase(type)) {
-      log.warn("{} is not a {} file", file, type);
-      return null;
-    }
+  public static boolean loadFile(String file) {
     File f = new File(file);
     Python p = (Python) Runtime.getService("python");
     log.info("Loading  Python file {}", f.getAbsolutePath());
@@ -32,7 +27,7 @@ public class Utils {
     try {
       script = FileIO.toString(f.getAbsolutePath());
     } catch (IOException e) {
-      log.error("IO Error loading file : ",e);
+      log.error("IO Error loading file : ", e);
       return false;
     }
     // evaluate the scripts in a blocking way.
@@ -41,7 +36,7 @@ public class Utils {
       log.error("Error while loading file {}", f.getAbsolutePath());
       return false;
     } else {
-      log.info("Successfully loaded {}", f.getAbsolutePath());
+      log.debug("Successfully loaded {}", f.getAbsolutePath());
     }
     return true;
   }
--- a/src/main/java/org/myrobotlab/inmoov/Utils.java
+++ b/src/main/java/org/myrobotlab/inmoov/Utils.java
@@ -16,10 +16,10 @@ public class Utils {
   /**
    * This method will load a python file into the python interpreter. 
    */
-  public static boolean loadFile(String file, String type) {
+  public static Boolean loadFile(String file, String type) {
     if (!FilenameUtils.getExtension(file).equalsIgnoreCase(type)) {
       log.warn("{} is not a {} file", file, type);
-      return false;
+      return null;
     }
     File f = new File(file);
     Python p = (Python) Runtime.getService("python");
@@ -32,7 +32,7 @@ public class Utils {
     try {
       script = FileIO.toString(f.getAbsolutePath());
     } catch (IOException e) {
-      log.error("IO Error loading file : {}",e);
+      log.error("IO Error loading file : ",e);
       return false;
     }
     // evaluate the scripts in a blocking way.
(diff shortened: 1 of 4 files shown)


src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1761,30 +1761,21 @@ public void loadGestures() {
...
-        }
-        // evaluate the gestures scripts in a blocking way.
-        boolean result = p.exec(script, true, true);
-        if (!result) {
-          log.warn("Error while loading gesture file: {}", f.getAbsolutePath());
-        } else {
-          log.info("Successfully loaded gesture {}", f.getAbsolutePath());
-        }
-      }
+      loaded = Utils.loadPythonFile(f.getAbsolutePath());
```
kwatters: you need to check that it's a python file before trying to load it.  You're missing that check here by deleting this code.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1761,30 +1761,21 @@ public void loadGestures() {
...
     }
+    return true;
```
kwatters: You are also missing the point that this method previously would skip failed scripts.  This is a highly desirable behavior because it means it will continue to load gestures even if one of them has a syntax error in it.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1761,30 +1761,21 @@ public void loadGestures() {
...
+  }
+
+  public void stopGesture() {
+    Python p = (Python) Runtime.getService("python");
```
kwatters: Why does calling stop on the python service "stop a gesture"  What is the point of this method?

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1876,7 +1867,7 @@ private File makeGesturesDirectory(String directory) {
     dir.mkdirs();
     if (!dir.isDirectory()) {
       // TODO: maybe create the directory ?
-      log.warn("Gestures directory {} doest not exist.", directory);
+      log.warn("Gestures directory {} doest not exist, will create it.", directory);
```
kwatters: This log message is not correct.  The line above dir.mkdirs() actually tries to create the directory.  This log message is in the event that you can't create the directory... one such example could be a file permissions issue.  please revert this log message as it's incorrect as to what is happening here.

src/main/java/org/myrobotlab/inmoov/Utils.java
```diff
@@ -16,13 +16,9 @@
   /**
    * This method will load a python file into the python interpreter. 
    */
-  public static boolean loadPythonFile(String file, String intanceName) {
+  public static boolean loadPythonFile(String file) {
```
kwatters: I'm not sure this is safe in a static context.  Also, it doesn't check if the python service is up and running.. and.. it assumes that the python instance is named "python"  (though, that's a pretty safe assumption.

src/main/java/org/myrobotlab/inmoov/Utils.java
```diff
@@ -16,13 +16,9 @@
   /**
    * This method will load a python file into the python interpreter. 
    */
-  public static boolean loadPythonFile(String file, String intanceName) {
+  public static boolean loadPythonFile(String file) {
     File f = new File(file);
-
     Python p = (Python) Runtime.getService("python");
```
kwatters: Python service could be null.

src/main/java/org/myrobotlab/inmoov/Utils.java
```diff
@@ -16,13 +16,9 @@
...
+  public static boolean loadPythonFile(String file) {
     File f = new File(file);
-
     Python p = (Python) Runtime.getService("python");
-    if (!intanceName.equals("inMoov")) {
-      //p.exec("inMoov=" + intanceName, true, true);
-    }
     log.info("Loading  Python file {}", f.getAbsolutePath());
     String script = null;
     try {
```
kwatters: You should verify that it's a python file with a .py file extension.

src/main/java/org/myrobotlab/inmoov/Utils.java
kwatters: Also, this method should not return true if the underlying script execution had an error.. it should return false.  otherwise what's the point of returning anything?

src/main/java/org/myrobotlab/inmoov/Utils.java
kwatters: The IO exception isn't logged properly that occurs here.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1761,30 +1761,21 @@ public void loadGestures() {
...
    */
-  public void loadGestures(String directory) {
-    // TODO: iterate over each of the python files in the directory
+  public boolean loadGestures(String directory) {
+
+    // iterate over each of the python files in the directory
     // and load them into the python interpreter.
-    File dir = makeGesturesDirectory(directory);
+    File dir = new File(directory);
+    boolean loaded = true;
```
kwatters: The loaded variable should be set AFTER it's loaded the scripts.. not before.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1761,30 +1761,21 @@ public void loadGestures() {
...
     for (File f : dir.listFiles()) {
```
kwatters: The gestures directory might not exist, you should be logging a message for that. and aborting the loading of scripts.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1761,30 +1761,21 @@ public void loadGestures() {
    * @param directory
    *          - the directory that contains the gesture python files.
    */
-  public void loadGestures(String directory) {
-    // TODO: iterate over each of the python files in the directory
+  public boolean loadGestures(String directory) {
```
kwatters: If you're changing the return type for this, you should update the javadoc comment to explain what the return value is and means.  If it's a boolean, when is it true and when is it false.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -11,7 +11,7 @@
 import org.myrobotlab.framework.ServiceType;
 import org.myrobotlab.framework.Status;
 import org.myrobotlab.framework.interfaces.ServiceInterface;
-import org.myrobotlab.io.FileIO;
+import org.myrobotlab.inmoov.Utils;
```
kwatters: Why the need for an additional utils class?

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: https://github.com/MyRobotLab/inmoov/issues/165

src/main/java/org/myrobotlab/inmoov/Utils.java
moz4r: intanceName parameter was not related to python but inmoov instance. To fix a possible issue if user set inmoov service name other than "i01" . deleted will return to it later…
checked also for null py instance

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: Because loadFile with be used for other things that loadGesture, also this will reduce codes lines in the high level

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: reverted

moz4r:
No mistaken, errors was no blocking, but it missed lot of stuffs. with your help I think it is better now.

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: checked now

moz4r:
Done, will do an other PR for the combo

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1762,29 +1765,30 @@ public void loadGestures() {
...
-        }
-        // evaluate the gestures scripts in a blocking way.
-        boolean result = p.exec(script, true, true);
-        if (!result) {
-          log.warn("Error while loading gesture file: {}", f.getAbsolutePath());
-        } else {
-          log.info("Successfully loaded gesture {}", f.getAbsolutePath());
+    if (dir.exists()) {
+      for (File f : dir.listFiles()) {
+        Boolean fileLoaded = Utils.loadFile(f.getAbsolutePath(), extension);
```
kwatters: This is very awkward to treat a boolean value as true, false, or null.  As i mentioned before,  I think you should test the file extension before you try to load the file here.  the logic would be a lot simpler.

src/main/java/org/myrobotlab/service/InMoov.java
kwatters: if it's used by something other than InMoov why is it in the inmoov package?  Also, at the end of the day, this method just calls python.exec ..

moz4r: something other than Gestures but still related to inmoov. There is some stuffs inside python land not yet pushed here like language pack ( translated vars )

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: done

PR #373, opened 2018-11-29
fixed transition states file -> camera -> file -> camera

kwatters:
sorry, missed this because i was too eager to merge i guess.. twitchy fingers after lots of coffee.

### DIFF
--- a/src/main/java/org/myrobotlab/service/OpenCV.java
+++ b/src/main/java/org/myrobotlab/service/OpenCV.java
@@ -861,7 +861,10 @@ public class OpenCV extends AbstractVideoSource {
    * 
    * @param cameraIndex
    */
-  public void capture(int cameraIndex) {
+  public void capture(Integer cameraIndex) {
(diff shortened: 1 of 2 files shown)


moz4r:
To not forget I had also a transition issue : camera(opencv) -> stop -> camera(sarxos)

supertick:
thanks moz4r i'll check

supertick:
@moz4r 

So this works for me : 

```java
      cv.capture(0);
      cv.stopCapture();
      cv.setGrabberType("Sarxos");
      cv.capture(0);
```

cv.capture(0) will start a capture with the first camera using OpenCV as the "default" for cameras
cv.stopCapture() will nullify the frame grabber type - if you try to capture again without setting a type you explicitly want - it will use OpenCV, but if you set it explicitly with
cv.setGrabberType("Sarxos");
or any other frame grabber type - it will honor your desired frame grabber type.

If there is a better way to handle this please let me know.

PR #376, opened 2018-11-30
logs error, add more time to default face detect - possible test success

supertick:


src/main/java/org/myrobotlab/opencv/ImageFileFrameGrabber.java
```diff
@@ -72,7 +72,12 @@ public Frame grab() {
       image = cvLoadImage(path, CV_LOAD_IMAGE_UNCHANGED);
       cache.put(path, image);
     } else {
-      image = cache.get(path).clone();
+      IplImage cachedImage = cache.get(path);
```
kwatters: this "cache" should be removed.  Just load the file on demand from the file system.  It doesn't make sense to scan a whole directory and load every file into memory.  That's just asking for an OOM exception.

supertick: nice.

supertick: speed seems ok - removing ...

Issue #342, opened 2018-11-14
Virtual InMoov returns an error


hairygael:
**Describe the bug**
Virtual InMoov returns an error after starting, the widow pops up and disappears living the error notice.

**To Reproduce**
1. Launch the START_INMOOV.bat in "Virtual" mode
2. See all the tabs getting initialized
3. When it comes to initializing Virtual InMoov
4. See the error notice

**Expected behavior**
Virtual InMoov window should be launched and stable

**Desktop (please complete the following information):**
 - OS: [windows10 pro]
 - Browser chrome]
 - Version [1.1.961]

supertick:
simulator works - virtual mode is not a requirement to run simulator

PR #397, opened 2018-12-14
fix jme3 crash

moz4r:
Oh I think this will fix random jme crash related to "State was changed after rootNode.updateGeometricState() call" .

I just removed the throw exception that break space time, replaced by a System.out.println Inside Spatial.java from jme3-core

https://github.com/MyRobotLab/myrobotlab/issues/342

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -438,9 +438,9 @@
 
 <!-- JMonkeyEngine begin -->
   <dependency>
-    <groupId>org.jmonkeyengine</groupId>
+    <groupId>org.myrobotlab.jme3</groupId>
     <artifactId>jme3-core</artifactId>
-    <version>3.2.0-stable</version>
+    <version>3.2.0-custom</version>
     <scope>provided</scope>
   </dependency>
   <dependency>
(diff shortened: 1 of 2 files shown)


kwatters:
Don't replace the default JME dependency.  We don't want to maintain a custom version of this.  There needs to be a better way to fix this.  If it's a bug in JME it should be fixed and contributed to the JME project so we can pull it in as a dependency.


Do you have the full stack trace for this exception?  It's difficult to know what's going on here without seeing the full error message.   Maybe it's in a no-worky somewhere?

kwatters:
Do you have a copy of the full stack trace/error message?  I'd like to see when this is being called from MRL.

moz4r:
This is the only one error message I found.. before JME3 close, just after i01.startVinmoov() 
I will try to reproduce Inside eclipse for more debug, but for now occure,with the standalone jar

kwatters:
Yeah. this sounds like a threading issue on the MRL side, not an issue with JME.

supertick:
I have lots of jmonkey features & fixes  - I agree this is a threading issue .  Mods on the scene graph are happening before its fully initialized

Issue #395, opened 2018-12-12
autodisable function not responding


hairygael:
**Describe the bug**
Seems when there is crash with python, it removes the autodisable function for further gestures on servo.

**To Reproduce**
1. Start INMOOV_START
2. Run the gesture "Giving", which will throw an error related to a python piece being not defined.
3. Run "Rest" gesture
4. See the servos being not disabled after the proper timing.

**Expected behavior**
Even with python error, if you execute another gesture which doesn't throw an error, the servo should then be disabled.

**Desktop (please complete the following information):**
 - OS: windows 10 pro
 - Browser:chrome, safari
 - Version 1.1.56

moz4r:
Oh yes. When a gesture start, startedGesture() is called.
At the end finishedGesture() , only if there is no crash between...
Welcome ( soon ) to i01.execGesture() that will take care off crash and report it without break things.

hairygael:
Thanks for responding!

Gael Langevin
Creator of InMoov
InMoov Robot <http://www.inmoov.fr>
@inmoov <http://twitter.com/inmoov>

moz4r:
That is a sample to change a little how we can launch gestures, to talk about : 

https://github.com/MyRobotLab/inmoov/commit/33ff862acbf50d97f2bc20d0c3b50c751ec18345

You will see giving will not contain anymore started/stopped gesture, 2 advantages less code lines, and gestures chain without warning ( giving can call fullspeed by exemple )
The whole chain will be considered as 1 gesture.

execGesture will call for startedGesture() at the begining ( today is hardcoded and disable "autoDisable", we can enhance this )
and stopedGesture() at the end that end stuffs.
If there is an error in the py code, you will be vocally warned about, and also logged for more details. But the gesture will "finish" now.

We can call `i01.execGesture("giving()")` from script
or, `<oob><mrl><service>i01</service><method>execGesture</method><param>giving()</param></mrl></oob>`

hairygael:
Hello Anthony,
That seems fine to me. Lots of AIML to check and to adapt. :)
Maybe the warning for the python error could be an option?
When working on coding, we maybe don't want to get recalled about the
warning every now and then.
As a side note: Just tested version 1.1.66 and got a python error which
froze the cmd.exe, but MRL was still responding.

Gael Langevin
Creator of InMoov
InMoov Robot <http://www.inmoov.fr>
@inmoov <http://twitter.com/inmoov>

moz4r:
[...]

Issue #387, opened 2018-12-03
Remove logic that strips special characters from programab input.


kwatters:
ProgramAB is designed to take free form input from end users as is.  We should not sanitize this input.  This ticket is to remove the logic that strips punctuation and other characters from the input.  All logic for handling this should be taken care of in the bot specific normal.txt and denormal.txt.

PR #400, opened 2018-12-16
Inmoov align

moz4r:
following the last PR, I needed to move more things to fix thread problems, Lot off py lines will be deleted after that. I think more things rock solid now.
**This need some last testing**, sorry if we find new bugs. good news is we can autotest every gestures now.

- Fix gestures that used opencv like https://github.com/MyRobotLab/myrobotlab/issues/403
- Deleted inmoov2 draft ( will create a new branch later )
- Definitly fix yolo gesture ( gael found a race condition between ear and python classifier thread, now the publisher is inside java land )  https://github.com/MyRobotLab/inmoov/issues/180
- Mouth control refactored to attach servocontrol instead off arduino & hardcoded pins
- Implemented inmoov language pack functions to java land, that will solve lot off errors from py land https://github.com/MyRobotLab/inmoov/issues/181
- Added startBrain() to launch chatbot
- Added a new i01.execGesture() -> so if a gesture crash java land can report it and execute finishedGesture() that do stuff like autodisable .Also for now user is warned about gesture crash, maybe we can use it for autotesting. https://github.com/MyRobotLab/myrobotlab/issues/395
- Fix swing window overexpand after filter added ( just a trick, maximize window ... )
- Prepare inmoov service for GUI ( basic swing UI and webgui maybe ), that is not for Nixie..
- Removed ear->chatbot interception to use normalise / denormalise for french chatbot https://github.com/MyRobotLab/myrobotlab/issues/387
- Updated inmoov zip ( waiting for this pr to update inmoov repo )
- fix https://github.com/MyRobotLab/myrobotlab/issues/401
- I forgot some fixes
- Oh we can override every start function like startBrain() by using inmoov attach methods. Or like before by declaring services from python land, so nothing will change, that's the goal.


src/main/java/org/myrobotlab/inmoov/Utils.java
```diff
@@ -41,4 +51,57 @@ public static boolean loadFile(String file) {
     return true;
   }
 
+  /**
+   * This method will try to launch a python comman from java land
+   */
+  public static String execPy(String command) {
```
kwatters: just use the python method directly.  No reason to also have this method. This is just extra code that will need to be mainatined.

src/main/java/org/myrobotlab/inmoov/Utils.java
```diff
@@ -41,4 +51,57 @@ public static boolean loadFile(String file) {
...
+    try {
+      return python.evalAndWait(command);
+    } catch (Exception e) {
+      log.error("execPy : {}", e);
+    }
+    return null;
+  }
+
+  //list services from meta category (pasted from RuntimeGui.java)
+  public static List<String> getServicesFromCategory(final String filter) {
```
kwatters: functionality like this should be on Runtime, not here.  Also what's the point of this. why is this method here.  Again.. more code just means more to maintain.  Please keep code changes to a minimum. especially as we near the release.

src/main/java/org/myrobotlab/inmoov/Utils.java
```diff
@@ -41,4 +51,57 @@ public static boolean loadFile(String file) {
...
+          // log.debug("serviceType : " + serviceType.getName());
+          servicesFromCategory.add(serviceType.getSimpleName());
+        }
+      }
+
+    }
+    return servicesFromCategory;
+  }
+  
+  public static File makeDirectory(String directory) {
```
kwatters: doesn't belong here.  This is a file utils sort of method. In either respect, this is an unnecessary method as you can just call   file.makedirs()  ..  This code is bloat and extra code that isn't needed.

src/main/java/org/myrobotlab/inmoov/Vision.java
```diff
@@ -68,6 +86,11 @@ public OpenCVFilter setActiveFilter(String filterName) {
       instance.opencv.addFilter(filterName);
     }
     instance.opencv.setActiveFilter(filterName);
+    SwingGui gui = (SwingGui) Runtime.getService("gui");
```
kwatters: You should not be getting a handle to the gui here.  What's the point of this?  If you want to get the filter, that's defined by the opencv service, not the swing gui!

src/main/java/org/myrobotlab/inmoov/Vision.java
```diff
@@ -110,4 +132,47 @@ public boolean isTracking() {
   public boolean isReady() {
     return ready;
   }
+
+  /**
+   * sub onClassification to a analyze yolo filter, and save some code lines from main service
+   */
+  public void onClassification(TreeMap<String, List<Classification>> classifications) {
```
kwatters: Seems like the wrong place for this method.  Typically only services should have things invoked and published on them.  Vision is not a service.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -123,7 +127,9 @@ public InMoov(String n) {
   transient public OpenCV opencv;
   public Vision vision;
   transient public SpeechRecognizer ear;
+  public SpeechRecognition speechRecognition;
```
kwatters: SpeechRecognition class should not exist.  Please remove.   Why are you changing this ?  This was all worky before.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -123,7 +127,9 @@ public InMoov(String n) {
...
   transient public SpeechSynthesis mouth;
+  public SpeechSynthesizer speechSynthesizer;
```
kwatters: please remove this class.  it's unnecessary and should not be here.  If there is missing functionality from the speech synthesis interface class, we should add the appropriate functionality to be generic.  not specific to inmoov.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -143,6 +149,7 @@ public InMoov(String n) {
   transient public Arduino neopixelArduino;
   transient public UltrasonicSensor ultrasonicSensor;
   transient public ProgramAB chatBot;
+  transient public HtmlFilter htmlFilter;
```
kwatters: Why do you need this a member method?  if all you are doing is invoking the static strip html, you should not have a handle to it as a member to this class.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -154,9 +161,39 @@ public InMoov(String n) {
...
 
   public void attach(Attachable attachable) {
+    //opencv
     if (attachable instanceof OpenCV) {
       opencv = (OpenCV) attachable;
+      subscribe(opencv.getName(), "publishClassification");
+      //SpeechRecognizer
+    } else if (attachable instanceof SpeechRecognizer) {
+      if (attachable.getClass().getSimpleName().equals("WebkitSpeechRecognition")) {
+        WebGui webgui = (WebGui) Runtime.create("webgui", "WebGui");
```
kwatters: you should not be changing the parameters of the webgui here nor should you be creating it.  It's already created if it's being passed into this method.  Please remove.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -154,9 +161,39 @@ public InMoov(String n) {
...
+        webgui.autoStartBrowser(false);
+        webgui.startService();
+        webgui.startBrowser("http://localhost:8888/#/service/" + attachable.getName());
```
kwatters: you should not force the browser to start here.  This method is for attaching.  Please remove this code.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -154,9 +161,39 @@ public InMoov(String n) {
...
+      }
+      ear = (SpeechRecognizer) attachable;
+    } else if (attachable instanceof ProgramAB) {
+      if (htmlFilter == null) {
```
kwatters: html filter should not be here.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -154,9 +161,39 @@ public InMoov(String n) {
...
+        htmlFilter = (HtmlFilter) Runtime.start(this.getIntanceName() + ".htmlFilter", "HtmlFilter");
```
kwatters: you don't need to start this here.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -154,9 +161,39 @@ public InMoov(String n) {
...
+      }
+      chatBot = (ProgramAB) attachable;
+    }
+    //sub attach
+    if (chatBot != null && ear != null) {
+      chatBot.attach((Attachable) ear);
```
kwatters: nope.. all of this stuff should be in python.. and not here.  Do not handle all of this logic here.  It's way too much of a change before the nixie release.  Changes of this scale are for Ogre at this point and NOT for Nixie.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -154,9 +161,39 @@ public InMoov(String n) {
...
+    }
+    if (chatBot != null && mouth != null) {
+      //TODO attach...
+      chatBot.addTextListener(htmlFilter);
```
kwatters: this should be done in python as it was before.  Changes like this are for InMoov2 in ogre not Nixie!

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -342,21 +384,47 @@ public void loadGestures() {
   }
 
   /**
-   * This method will look at all of the .py files in a directory. One by one it
+   * This method will try to launch a python command
+   * with error handling
+   */
+  public String execGesture(String gesture) {
```
kwatters: too big a change for nixie.  stuff like this should be inmoov2 and in ogre.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -342,21 +384,47 @@ public void loadGestures() {
...
+  /**
+   * This blocking method will look at all of the .py files in a directory. One by one it
    * will load the files into the python interpreter. A gesture python file
    * should contain 1 method definition that is the same as the filename.
    * 
    * @param directory
    *          - the directory that contains the gesture python files.
    */
-  public void loadGestures(String directory) {
+  public Boolean loadGestures(String directory) {
```
kwatters: Should return a native "boolean"  not a Boolean here.  Booleans can be null.  This method,  if  it returns anything should only be true or false.  not "null" as is possible with a Boolean as a return type.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -342,21 +384,47 @@ public void loadGestures() {
...
 
     // iterate over each of the python files in the directory
     // and load them into the python interpreter.
     String extension = "py";
     Integer totalLoaded = 0;
     Integer totalError = 0;
-    File dir = makeGesturesDirectory(directory);
+    File dir = Utils.makeDirectory(directory);
```
kwatters: unnecessary helper function.   File.mkdirs() ...

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -372,6 +440,7 @@ public void loadGestures(String directory) {
       }
     }
     info("%s Gestures loaded, %s Gestures with error", totalLoaded, totalError);
+    return ((totalError == 0) ? true : false);
```
kwatters: so. any errors means this function returns false...what reads the return value of this fuction?

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -526,6 +584,7 @@ public void startedGesture(String nameOfGesture) {
     if (gestureAlreadyStarted) {
       warn("Warning 1 gesture already running, this can break spacetime and lot of things");
     } else {
+      log.info("Starting gesture : {}", nameOfGesture);
       gestureAlreadyStarted = true;
```
kwatters: seems like a lot of booleans here to maintain state.  This logic is too complex and should be simplified.  Having so many state variables is very difficult to test and leads to bugs.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -526,6 +584,7 @@ public void startedGesture(String nameOfGesture) {
...
       RobotCanMoveRandom = false;
```
kwatters: variable names should start with a lower case letter by java conventions.

src/main/java/org/myrobotlab/service/SwingGui.java
```diff
@@ -340,8 +340,12 @@ public void run() {
         send(name, "publishState");
 
         if (getName().equals(name) && guiServiceGui == null) {
-          guiServiceGui = (SwingGuiGui) newGui;
-          guiServiceGui.rebuildGraph();
+          try {
+            guiServiceGui = (SwingGuiGui) newGui;
+            guiServiceGui.rebuildGraph();
+          } catch (Exception e) {
```
kwatters: Why do we need this try/catch here.. we didn't have it before?  Don't change this .. please revert.

src/main/java/org/myrobotlab/swing/InMoovGui.java
```diff
@@ -64,6 +64,8 @@
   private final JPanel panel = new JPanel();
   private final JCheckBox Flip = new JCheckBox("Flip Camera");
   private final JCheckBox PyramidDown = new JCheckBox("PyramidDown");
+  private final JLabel lblNewLabel_1 = new JLabel("InMoov SwingGui is not yet ready... ");
```
kwatters: please use good variable names.  NewLabel_1 is not a good variable name.

src/main/java/org/myrobotlab/swing/InMoovGui.java
```diff
@@ -78,6 +80,11 @@ public InMoovGui(final String boundServiceName, final SwingGui myService) {
     ImageIcon generalIcon = Util.getImageIcon("InMoov.png");
     inmoovPane.addTab("General", generalIcon, generalPanel);
     generalPanel.setLayout(new GridLayout(3, 2, 0, 0));
+    lblNewLabel_1.setForeground(Color.RED);
+    
+    generalPanel.add(lblNewLabel_1);
```
kwatters: If these are only local to this method the labels should not be added as members to the class.

moz4r:
those changes are bugfix ! As I said I had to change things in depth to fix little bugs... because some foundations were not solid .... I will answer your comments, thank you for the review

src/main/java/org/myrobotlab/inmoov/Utils.java
moz4r: Just to save some code lines makeDirectory() is shared by 2 functions..

src/main/java/org/myrobotlab/inmoov/Vision.java
moz4r: The point is not to get a filter, but the current gui . To maximize the jframe after a filter Added. That will resolve a bug that overexpand the window after a filter is added

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: That is not unnecessary, that is for SwingUI and and error management, but will revert to return later.. Like you said maybe by mods speech synthesis interface

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1361,73 +1442,110 @@ public void startAll(String leftPort, String rightPort) throws Exception {
...
 
   // gestures begin ---------------
 
-  public SpeechSynthesis startMouth() throws Exception {
-    mouth = (SpeechSynthesis) startPeer("mouth");
-    speakBlocking("starting mouth");
-
-    if (ear != null) {
-      ear.addMouth(mouth);
+  public SpeechSynthesis startMouth() {
```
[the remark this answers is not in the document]
moz4r: it will work even better if inadvertently the service ear is launched after the service mouth ( not deleted , moved after services are attached )

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -996,7 +1066,7 @@ public InMoovHead startHead(String port, Integer headYPin, Integer headXPin, Int
   public InMoovHead startHead(String port, String type, Integer headYPin, Integer headXPin, Integer eyeXPin, Integer eyeYPin, Integer jawPin, Integer rollNeckPin)
       throws Exception {
     // log.warn(InMoov.buildDNA(myKey, serviceClass))
-    speakBlocking(String.format("starting head on %s", port));
+    speakBlocking(languagePack.get("STARTINGHEAD")+" "+port);
```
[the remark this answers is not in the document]
moz4r: I'm not sure I changed this ( just the text ) , also I think speablocking instead of speak is important to avoid concurent speech events, or speaking for events that are already done because fast launched... ( also mute is possible at startup )

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1226,13 +1296,14 @@ public void powerDown() {
     disable();
 
     if (ear != null) {
-      ear.lockOutAllGrammarExcept("power up");
+      ear.lockOutAllGrammarExcept("wake up");
```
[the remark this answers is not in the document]
moz4r: It was already hard coded, we are not using this inside py land, I wanted also to change this function to merge stuff but seem it is not for today. I will revert the original word ..

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1246,6 +1317,7 @@ public void powerUp() {
     python.execMethod("power_up");
   }
 
+  //TODO FIX/CHECK this, migrate from python land
   public void publishPin(Pin pin) {
```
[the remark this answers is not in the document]
moz4r: I changed nothing, just puted a comment because we are not using it inside py land

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1311,17 +1383,26 @@ public void setMute(boolean mute) {
     this.mute = mute;
   }
 
-  public List<AudioData> speakBlocking(Status test) {
-    if (test != null && !mute) {
-      return speakBlocking(test.toString());
+  public List<AudioData> speakBlocking(String toSpeak) {
```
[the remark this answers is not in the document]
moz4r: what method, speakBlocking ? it was already here. And I think it is good for gestures, if we don't have a "mouth" it will not crash

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1311,17 +1383,26 @@ public void setMute(boolean mute) {
...
     return null;
   }
 
-  public List<AudioData> speakBlocking(String toSpeak) {
+  public List<AudioData> speakAlert(String toSpeak) {
+    speakBlocking(languagePack.get("ALERT"));
+    return speakBlocking(toSpeak);
+  }
+
+  public List<AudioData> speak(String toSpeak) {
```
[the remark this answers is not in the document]
moz4r: it is necessarie for gestures, if we don't have a "mouth" it will not crash the whole gesture

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1361,73 +1442,110 @@ public void startAll(String leftPort, String rightPort) throws Exception {
...
+    chatBot.setPredicate("questionfirstinit", "");
+    chatBot.setPredicate("tmpname", "");
+    chatBot.setPredicate("null", "");
+    //load last user session
+    if (!chatBot.getPredicate("name").isEmpty()) {
+      if (chatBot.getPredicate("lastUsername").isEmpty() || chatBot.getPredicate("lastUsername").equals("unknown")) {
+        chatBot.setPredicate("lastUsername", chatBot.getPredicate("name"));
+      }
+    }
+    chatBot.setPredicate("parameterHowDoYouDo", "");
```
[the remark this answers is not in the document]
moz4r: that's related to HowDoYouDo gesture, to reset a counter at startup. 
maybe we can find later other thing cleaner inside the gesture itself

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1361,73 +1442,110 @@ public void startAll(String leftPort, String rightPort) throws Exception {
...
+      mouthControl.attach((Attachable) mouth);
       mouthControl.setmouth(10, 50);
     }
     return mouthControl;
   }
 
   public boolean startOpenCV() {
-    if (opencv == null)  {
-      opencv = (OpenCV)startPeer("opencv");
+    speakBlocking(languagePack.get("STARTINGOPENCV"));
```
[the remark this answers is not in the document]
moz4r: awsered, we have already speakblocking events, we can setMute() to speedup. that is just a missing event

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1519,29 +1637,30 @@ public void preShutdown() {
...
-    }
   }
 
   @Override
   public void startService() {
     super.startService();
+    //we need auto load : if user launch inmoov from runtime, or start from script
+    //without load : this will overwrite EVERY previous personal configs !!
+    //OR : we need to disable autosave EVERYWHERE..
+    load();
```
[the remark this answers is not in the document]
moz4r: Kwatters… That's a fact if we launch inmoov service from runtime GUI load **will not be called**, and every parameters will be overwriten at shutdown… because we don't want gui for now, I will revert to avoid infinite loop talk, maybe we can talk about it later

src/main/java/org/myrobotlab/service/NaturalReaderSpeech.java
```diff
@@ -59,7 +59,7 @@ static public ServiceType getMetaData() {
     meta.addPeer("httpClient", "HttpClient", "httpClient");
     meta.addCategory("speech", "sound");
     // moz4r : set to hidden / poc only, unstable ...
-    meta.setAvailable(false);
+    meta.setAvailable(true);
```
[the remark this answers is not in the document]
moz4r: not sure :) but why not

src/main/java/org/myrobotlab/service/SwingGui.java
moz4r: Ok, but I saw an error one time because of this.. will revert

src/main/java/org/myrobotlab/swing/InMoovGui.java
moz4r: agree..

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: ok let's do it Inside ogre ...

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: we can do it as it was before, just added a shorcut ... but reverted ...

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: No we need it ! not sure about when is the next release... that is optional , gestures are not using it yet..

src/main/java/org/myrobotlab/inmoov/Utils.java
```diff
@@ -12,6 +13,9 @@
 
   public final static Logger log = LoggerFactory.getLogger(Utils.class);
 
+  transient private static Runtime myRuntime = (Runtime) Runtime.getInstance();
```
kwatters: I think if you have this here, it will cause runtime to be instantiated any time this Utils class is loaded into the classpath.  I don't think we want this behavior.

src/main/java/org/myrobotlab/inmoov/Vision.java
kwatters: Let's keep callback/subscribe methods to Service classes if we can.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -154,8 +158,20 @@ public InMoov(String n) {
...
+      //SpeechRecognizer
+    } else if (attachable instanceof SpeechRecognizer) {
+
+      ear = (SpeechRecognizer) attachable;
+    } else if (attachable instanceof ProgramAB) {
+      chatBot = (ProgramAB) attachable;
+    }
+    if (mouth != null && ear != null) {
+      //TODO attach...
+      ear.addMouth(mouth);
```
kwatters: normally in python you would be doing this.  I don't think this is the right place for it.  This means that any time any service is attached after the ear and mouth have been attached,  this ear.addMouth method is going to get called.

src/main/java/org/myrobotlab/service/InMoov.java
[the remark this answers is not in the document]
kwatters: There's a lot of additional code here for situations where the Mouth is null.   This is ok, but we should really address the issue where a mouth might not have been attached yet and the inmoov is asked to "speak" ..

src/main/java/org/myrobotlab/service/InMoov.java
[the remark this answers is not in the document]
kwatters: I think we need to address the startup order and attachment order of services in the inmoov, but perhaps that's an ogre thing.

src/main/java/org/myrobotlab/service/InMoov.java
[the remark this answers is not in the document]
kwatters: good point about mute.

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: agree, that's why I started to rename some.. there is a risk to break stuff for now

src/main/java/org/myrobotlab/service/InMoov.java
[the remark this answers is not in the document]
moz4r: error notification about missing mouth added

src/main/java/org/myrobotlab/service/InMoov.java
moz4r: If we can offer some shortcuts ... I prefer not to remove the method for the moment because it was already launched while startMouth(). I reviewed the code so that it would be executed only 1 time. it would be great to have a global x.isAttached(y) also ..

moz4r:
seem ok, gestures from py land updated also

src/main/java/org/myrobotlab/service/MouthControl.java
```diff
@@ -18,79 +20,31 @@
...
-    // TODO: mouth should probably implement speech synthesis.
-    // in a way of speaking, one day, people may be able to read the lips
-    // of the inmoov.. so you're synthesising speech in a mechanical way.
-    // similar to sign language maybe?
-    subscribe(mouth.getName(), "publishStartSpeaking");
-    subscribe(mouth.getName(), "publishEndSpeaking");
   }
 
-  // FIXME make interface
-  public boolean connect(String port) throws Exception {
```
kwatters: changing/removing this method name will probably break a bunch of scripts. but I think it's better to start moving towards the attach methods.

moz4r:
thanks ! time to do some test now ...

PR #405, opened 2018-12-18
prevent infinite loop without disabling autolisten

moz4r:


src/main/java/org/myrobotlab/service/WebkitSpeechRecognition.java
```diff
@@ -51,6 +51,7 @@
 
   private boolean listening = false;
   private boolean speaking = false;
+  private boolean commError = false;
```
kwatters: what reads this value ?

src/main/java/org/myrobotlab/service/WebkitSpeechRecognition.java
```diff
@@ -198,22 +199,26 @@ public void startListening() {
   @Override
   public void stopListening() {
     log.debug("Stop listening event seen.");
+    commError = false;
```
kwatters: this variable should be defined locally to this method if it's not used outside of this method.

src/main/java/org/myrobotlab/service/WebkitSpeechRecognition.java
moz4r: what value ? :)

PR #407, opened 2018-12-19
Removed vocal alert about gesture error

moz4r:


hairygael:
Hello,
I am wondering, what is a gesture doing in the start up sequence?
i01.execGesture("BREAKITdaVinci()");

moz4r:
Hi :) not really in the startup sequence but the main() , it the place where we can put test code. I do it on purpose made a mistake for testing.

hairygael:
Ah ok
:)
I question because there use to be hardcoded gestures in java land at some
point, and I was wondering if that was a remaining relicat

Gael Langevin
Creator of InMoov

PR #412, opened 2018-12-21
fix UI OverExpand ( optional )

moz4r:
Maybe not the better way, this will fix UI overexpand
Also it can be useful to get always fullscreen..

opencv.fixUIOverExpand=true


supertick:
Ahoy moz4r - please use this https://github.com/MyRobotLab/myrobotlab/pull/413 in InMoov

Issue #343, opened 2018-11-14
SwingGui resizes when open cv begins to capture.


kwatters:
The opencv capture should not trigger a resize of the swing gui.  Instead we should have the VideoWidget handle auto resizing so the swing gui doesn't get so large it goes off the screen.

supertick:
blech - swing is evil, the JavaCV's CanvasFrame is currently used for native display, webgui for remote display

PR #413, opened 2018-12-23
SwingGui.maximize()

supertick:
use SwingGui.maximize() from InMoov if you want the gui to maximize


moz4r:
auto-adjusts will be great to not stress the jframe, this would avoid commands like pack () or maximize ()

moz4r:
https://github.com/MyRobotLab/myrobotlab/issues/343

PR #414, opened 2018-12-31
276 whitespace format

kwatters:
we can re-do this again if we want to tweak things.. this is just an initial cut at a full whitespace refactor , converting to unix style line endings, and increasing the mrl formatter to make lines wrap at 132 chars (instead of 80)


kwatters:
doh! wrong base branch..

PR #417, opened 2019-01-02
276 whitespace format

kwatters:


supertick:
What files are needed to keep it this way ? formatter details ? git details ? Are they checked in ? - e.g. a sample git config ?

kwatters:
This PR kept the line length for comments at 80.  Also, the java formatter doesn't have a way to enforce CR/LF being unix style..   
You can tell Eclipse to default to unix style endings based on this page:
https://stackoverflow.com/questions/1886185/eclipse-and-windows-newlines

PR #424, opened 2019-01-31
jem and servo updates

supertick:


src/main/java/org/myrobotlab/jme3/Interpolator.java
```diff
@@ -1,21 +1,97 @@
...
+  // FIXME - use ms for rate vs just seconds ?!
+  public void generateMoves() {
+    // for (Move move : futureMoves.values()) {
+    for (String name : futureMoves.keySet()) {
+      Move move = futureMoves.get(name);
+      // find the time into our move (total) vs last ..
+      long deltaTime = System.currentTimeMillis() - move.startTs; // milliseconds
+
+      // total difference in our move - degrees per second
+      float interPos = move.startPos + (move.direction * deltaTime * move.speed / 1000);
```
kwatters: It'd be really nice to have acceleration here by adding   0.5 * acceleration * deltaTime * deltaTime to this equation.

supertick: uh .. we can't use -1 (which it currently is) so what is the average acceleration value ?

src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -548,14 +544,14 @@ public synchronized void moveTo(double pos) {
 
     lastActivityTime = System.currentTimeMillis();
 
-    if (lastPos != pos) {
+   // if (lastPos != pos) {
```
kwatters: this introduces a slight change in behavior for the auto-disable  (probably warranted.. just calling out that there might be some testing to do around auto-disable/enable functionality here.)

src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
```diff
@@ -67,7 +67,7 @@
    * @throws Exception
    *           e
    */
-  void attachServoController(ServoController controller) throws Exception;
+  void attachServoController(Attachable controller) throws Exception;
```
kwatters: Ok with this change, however, it seems like you should only be allowed to pass in a ServoController here...  which also says that "Simulator" and "ServoController" should be the same interface for purproses of a JME3 engine?

src/main/java/org/myrobotlab/service/interfaces/Simulator.java
```diff
@@ -14,13 +12,19 @@
  *
  */
 
+// FIXME !!! refactor this stuff out - create an AbstractSimulator ... 
 public interface Simulator extends NameProvider {
```
kwatters: Seems like this should implement ServoController ...

src/main/java/org/myrobotlab/service/interfaces/Simulator.java
```diff
@@ -14,13 +12,19 @@
...
   // Generalized create ??? a good thing - router for ObjectFactory??
   // public Object create(ServiceInterface service);
 
+  /*
   public VirtualServo createVirtualServo(String name);
 
   public VirtualMotor createVirtualMotor(String name);
+  */
+  
+  ServoController getServoController();
```
kwatters: If this interface implements ServoController this method isn't needed.

src/main/java/org/myrobotlab/jme3/Interpolator.java
kwatters: 0.0 .. no acceleration.

src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -836,34 +835,47 @@ public double getPos() {
...
+      return;
+    }
+    
+    // find servo controller - 
+    if (attachable instanceof ServoController) {
+      sc = (ServoController)attachable;
+    } else if (attachable instanceof Simulator) {
+      sc = ((Simulator)attachable).getServoController();
+    } else {
+      log.error("{} unsupported type {}", attachable.getName(), Attachable.class.getCanonicalName());
```
kwatters: A log message here is fine, however, this method signature says it throws an exception.  I think if you pass in a bogus service, you should throw something here.. not just log it and silently return..

src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
supertick: I did not want to bloat Simulators by making them implement every possible controller we want to virtualize ... I also didn't want to crowd our Services with "Implementation" services who's sole purpose is to provide a ServoController for a very specific Simulator ...
E.g. I don't thing we need a full blown JMEServoController service ... additionally I don't want the Simulator itself to implement every sing method specified in the ServoController interface.

The attach routes based on type to the appropriate details to attach a simulator to a servo (through a lightweight JmeServoController which is just a class not a service in the org.myrobotlab.jme package space)

src/main/java/org/myrobotlab/service/Servo.java
supertick: I don't mind either way ... you want it to go BOOM bigger .. fine by me

src/main/java/org/myrobotlab/service/interfaces/Simulator.java
supertick: it is .. unless you can figure a way to support the gui and "attaching" correctly without it ..
I see this as a small cost compared to begin pushing all methods of all interfaces into the simulator :(

src/main/java/org/myrobotlab/service/interfaces/Simulator.java
supertick: i hope not .. 
Right now it "has a" ServoController but "is a" Simulator - and when I get MotorController working and all sorts of other virtualized sensors or control components - I don't want the simulator to "be a" "everything"

src/main/java/org/myrobotlab/service/Servo.java
supertick: honestly I don't understand how this worked previously .. pos is not guaranteed to be set - what value is this ?

src/main/java/org/myrobotlab/jme3/Interpolator.java
supertick: I'm afraid to set it to 0.0 .. someone set it to -1 .. it looks like it might be looked at by other services (Arduino & Adafruit16)

PR #430, opened 2019-02-02
multi - ik3d and little cleanup

supertick:


src/main/java/org/myrobotlab/jme3/MainMenuState.java
```diff
@@ -37,7 +37,7 @@
   transient JMonkeyEngine jme = null;
   transient Jme3App app;
   transient Container main;
-  transient Label breadCrumbs;
+  transient Label breadCrumbsx;
```
kwatters: Is this a typo? did you really want to rename this? or maybe you want to delete this variable?

src/main/java/org/myrobotlab/service/InverseKinematics3D.java
```diff
@@ -363,7 +374,7 @@ public void onPoint(Point point) {
     log.info("Attempting to move to {}", point);
 
     // TODO: scale / translate & rotate...
-    moveTo(point);
+    moveTo(currentArm, point);
```
kwatters: This is probably problematic.. we should make note that onPoint interface isn't aware of which arm it should be publishing points to...

PR #434, opened 2019-02-05
multi joint servo control

supertick:


supertick:
Nope .. its not related ..
This is mapping JMonkey Geometries to a service .. previously its been only 1:1 .. but now we can handle 1:N

PR #436, opened 2019-02-26
Resource dir

kwatters:
centralizing the resource directory into a single method (I think it is comprehensive.)  you can override the resource directory by specifying the resource.dir system property.  It does touch on some of the serviceData.json generation logic, so that should be sanity checked.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1300,7 +1300,7 @@
   <dependency>
     <groupId>com.google.code.gson</groupId>
     <artifactId>gson</artifactId>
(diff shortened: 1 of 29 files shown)


pom.xml
```diff
@@ -1300,7 +1300,7 @@
   <dependency>
     <groupId>com.google.code.gson</groupId>
     <artifactId>gson</artifactId>
-    <version>2.8.0</version>
+    <version>2.8.5</version>
```
supertick: was the pom regenned ?

PR #439, opened 2019-03-04
Worke

supertick:
lots of emoji and fsm updates


src/main/java/org/myrobotlab/fsm/api/Event.java
```diff
@@ -0,0 +1,70 @@
+/**
```
kwatters: Was there no jar that we could pull all of these classes in from?  I'd prefer if we had a dependency for this, rather than a copy & paste..

supertick: It's not a copy & paste.  Yes there is a library, but it falls short of being capable of implementing my desired use case.   I still believe easy-states is an excellent example of minimal & clean implementation of a FSM, however, there were certain design implementations I couldn't get around.  One of the largest was the idea that a new Event type had to be implemented with a new class.  

The required use case I needed was a service which could completely described, run, receive inputs, and publish outputs of arbitrary complexity without the need to add new code.

PR #442, opened 2019-03-05
fixed undocking problem

supertick:


src/main/java/org/myrobotlab/swing/widget/DockableTab.java
```diff
@@ -232,10 +233,17 @@ public void run() {
...
         // icon
-        URL url = getClass().getResource(Util.getResourceDir() + "/mrl_logo_36_36.png");
-        Toolkit kit = Toolkit.getDefaultToolkit();
-        Image img = kit.createImage(url);
-        undocked.setIconImage(img);
+        // URL url = getClass().getResource(Util.getResourceDir() +
+        // "/mrl_logo_36_36.png");
+        URL url = null;
+        try {
+          url = new URL("file:///" + Util.getResourceDir() + File.separator + "mrl_logo_36_36.png");
```
kwatters: hmm.. i think you want to getResourceDir here with forward slashes, and not the system dependant ones..

supertick: heh .. your right .. I first switched that one .. and it did not work ..
only after did I add the file:/// reference .. changing it back to forward slash ..

PR #447, opened 2019-03-07
moved webgui's client to WebGui/app

supertick:
will be removing web refactoring branch - although it can still be reached as download here 
https://drive.google.com/file/d/19717imBu0zUvFAktc5qk9K8hdrN2Dxzf/view?usp=sharing


src/test/resources/ProgramAB/bots/lloyd/aiml/mrl_added.aiml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kwatters: This shouldn't be checked in.

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -786,11 +786,11 @@ public String getDataInstanceDir() {
   }
 
   public String getResourceDir() {
-    return "data" + File.separator + getClass().getSimpleName();
+    return "resource" + File.separator + getClass().getSimpleName();
```
kwatters: we should reference "resource" from Util.getResourceDir()  no ?

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -786,11 +786,11 @@ public String getDataInstanceDir() {
   }
 
   public String getResourceDir() {
-    return "data" + File.separator + getClass().getSimpleName();
+    return "resource" + File.separator + getClass().getSimpleName();
   }
 
   public String getResourceInstanceDir() {
-    return "data" + File.separator + getClass().getSimpleName() + File.separator + getName();
+    return "resource" + File.separator + getClass().getSimpleName() + File.separator + getName();
```
kwatters: same comment about using the Util.getResourceDir() method instead of it being hard coded here.

supertick: fixed ... added Util.getDataDir() for consistency too

PR #451, opened 2019-03-09
Worke

supertick:
ImageDisplay fixes


src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -971,7 +967,7 @@ public void start() {
       subscribe(runtime.getName(), "registered");
       subscribe(runtime.getName(), "released");
 
-      if (autoStartBrowser) {
+      if (autoStartBrowser && GraphicsEnvironment.isHeadless()) {
```
LunDev: Did you mean
```suggestion
      if (autoStartBrowser && !GraphicsEnvironment.isHeadless()) {
```
or did I misunderstood something?
`GraphicsEnvironment.isHeadless()` returns _true if this environment cannot support a display, keyboard, and mouse; false otherwise_
[JavaDoc](https://docs.oracle.com/javase/7/docs/api/java/awt/GraphicsEnvironment.html#isHeadless())

PR #454, opened 2019-03-12
Worke junit

supertick:
lots of releaseService and stopService thread fixes, add/purge task race condition fixed, Peer Release Worky !  WooHoo !


src/main/java/org/myrobotlab/framework/Instantiator.java
```diff
@@ -18,7 +18,7 @@ static public Object getNewInstance(Class<?>[] cast, String classname, Object...
     try {
       return getThrowableNewInstance(cast, classname, params);
     } catch (Exception e) {
-      log.warn("getNewInstance failed for {} {} {}", cast, classname, e.getClass().getCanonicalName());
+      log.info("getNewInstance failed for {} {} {}", cast, classname, e.getClass().getCanonicalName());
```
kwatters: This probably is an error or at least a warn level message.  If this comes up, I would think there is a bug somewhere that you want to know about?

src/main/java/org/myrobotlab/net/BareBonesBrowserLaunch.java
```diff
@@ -29,7 +31,13 @@
   static final String errMsg = "Error attempting to launch web browser";
 
   public static void openURL(String url) {
-    try { // attempt to use Desktop library from JDK 1.6+
+    try { 
+      
+      if (!GraphicsEnvironment.isHeadless()) {
+        log.warn("graphical browser requested but currently no display");
```
kwatters: Is this logic backwards?  And if the env is headless, wouldn't you want to just return and not spawn the browser?

src/main/java/org/myrobotlab/net/CommunicationManager.java
```diff
@@ -67,7 +67,7 @@ final public void send(final Message msg) {
 
     ServiceInterface sw = Runtime.getService(msg.getName());
     if (sw == null) {
-      log.error("could not find service {} to process {} from sender {} - tearing down route", msg.name, msg.method, msg.sender);
+      log.info("could not find service {} to process {} from sender {} - tearing down route", msg.name, msg.method, msg.sender);
```
kwatters: This probably is a warning level ..  someone is trying to send a message to an unknown service, right?  that indicates a bug somewhere.

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -989,28 +989,26 @@ public void startService() {
...
     }
 
     start();
   }
 
   public void stop() {
     if (nettosphere != null) {
-
-      log.info("stopping nettosphere");
+      log.warn("==== nettosphere STOPPING ====");
```
kwatters: This is info level

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -989,28 +989,26 @@ public void startService() {
...
-      sleep(1000);
+      /*
+       * (new Thread("stopping nettophere") { public void run() {
+       * 
+       * log.error("==== nettosphere stopping  ===="); //
+       * nettosphere.framework(). // nettosphere.framework().destroy();
+       * nettosphere.stop(); log.error("==== nettosphere STOPPED ===="); }
+       * }).start(); sleep(1000);
+       */
+      log.warn("==== nettosphere STOPPED ====");
```
kwatters: this should also be info level.    (only if the stop() method blows up. should we print a warn/stack trace..)  but it doesn't look like stop() can throw anything.

src/test/java/org/myrobotlab/test/AbstractTest.java
```diff
@@ -0,0 +1,167 @@
...
+    boolean isVirtual = true;
+    String isVirtualProp = System.getProperty("junit.isVirtual");
+
+    if (isVirtualProp != null) {
+      isVirtual = Boolean.parseBoolean(isVirtualProp);
+    }
+    return isVirtual;
+  }
+
+  public static void main(String[] args) {
```
kwatters: a unit test shouldn't have a main method.

src/test/java/org/myrobotlab/service/AbstractServiceTest.java
```diff
@@ -1,49 +0,0 @@
...
- * A base generic test class that can start a service, call a test method on it,
- * and then release the service The idea is you can subclass this and implement
- * the createService() and testService() methods to instrument a unit test for
- * any service in MRL.
- * 
- * @author kwatters
- *
- */
-@Ignore
-public abstract class AbstractServiceTest {
```
kwatters: Seems like having a base class for testing services specifically would be useful.  We have lots of services to test.  Having a base class for those tests would be desirable.

src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,17 @@
+<configuration debug='true'>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
```
kwatters: we should sanity check that stderr gets captured also.

src/test/resources/logback.xml
```diff
@@ -0,0 +1,26 @@
+<configuration>
+	<appender name="STDOUT"
```
kwatters: does this capture stderr?

src/main/java/org/myrobotlab/framework/Instantiator.java
supertick: primarily this is used to "attempt" to get gui's, but when a gui is not been developed, the framework displays a empty frame with "X has no display" 
It will return null in this case, and the calling code will handle it appropriately, there is also a throwable version

src/main/java/org/myrobotlab/net/CommunicationManager.java
supertick: when a service is released, it currently does not scan through all services, and all subscriptions to see if there are any hanging references.
When a message is routed to a service which does not exist, the error will cause the msg sendor to tear down the route ..

Having services removed or added is potentially common, and the framework should allow graceful (or graceful as possible) releasing of services.

Msg routing can be handled at a higher level, for example, if you depend on a service and get a onReleased msg from the framework, the service can do what it thinks is appropriate..

src/main/java/org/myrobotlab/service/WebGui.java
supertick: temporary debug - the logs are nice and clear, but still nettosphere is not behaving correctly,
when i get the issue fully resolved I'll move it to info

src/test/java/org/myrobotlab/service/AbstractServiceTest.java
supertick: Sure .. it should extend AbstractTest ..

src/test/resources/logback-test.xml
supertick: pretty sure it does .. check the jenkins log or run mvn locally

supertick: OOP ! .. good catch ;)
I'll add it

PR #467, opened 2019-03-26
Inmoov unit test

kwatters:
this tests 45% of the inmoov source code.  we can always improve upon it with better assertions.


hairygael:
Cool !

Gael Langevin
Creator of InMoov

PR #470, opened 2019-03-27
removed servo orchestrator

supertick:


LunDev:
Rest in Peace, ServoOrchestrator.
I will not forget you.
Maybe you will be resurrected someday, harder, better, faster, stronger than never before. 
I will think about doing this - I promise!
Now go, and please find peace in heaven - or at least in our git history.
Goodbye!

PR #476, opened 2019-03-31
Servo fixes

supertick:


src/test/resources/ProgramAB/bots/pikachu/aiml/pikachu.aiml
```diff
@@ -1,10 +1,8 @@
...
-		<!--  あなたを*に会うのはいいことだ   ?? -->
-		<!--  あなたを*に会えてうれしい -->
-		<!-- あなたに会えてよかったXXXXX -->
-		<template>あなたに会えてよかった<star /></template>
-	</category>
-</aiml>
\ No newline at end of file
+<?xml version="1.0" encoding="UTF-8"?>
+<aiml>
+<!--  -->
```
kwatters: please revert this change.

kwatters: programab should no longer overwrite this file when the test runs..

.gitignore
```diff
@@ -98,3 +98,4 @@ Thumbs.db
 /models/
 /Common/
 /bulletjme.dll
+testbots/*
```
kwatters: this entry shouldn't be necessary

src/test/resources/ProgramAB/bots/pikachu/aiml/pikachu.aiml
supertick: reverted to develop version

.gitignore
supertick: removed

src/main/java/org/myrobotlab/service/Emoji.java
```diff
@@ -61,111 +61,11 @@
...
-    log.info(emotionalState.getFsmMap());
-
-    // loopback on state publishing...
-    // subscribe(getName(),"publishState");
-    try {
-      emotionalState.attach(this);
-    } catch (Exception e) {
-      error(e);
-    }
+    startPeer("fsm");
```
kwatters: we really need to call this 3 times?

src/test/java/org/myrobotlab/document/transformer/AbstractStageTest.java
```diff
@@ -13,17 +13,18 @@
...
-    stage.processDocument(doc);
-    validate(doc);
+  @Before
+  public void init() {
+    // LoggingFactory.init("WARN");
+  }
+
+  @Test
+  public void test() {
+    if (printMethods)System.out.println(String.format("Running %s.%s", getSimpleName(), getName()));
```
kwatters: Is this necessary now?

src/test/java/org/myrobotlab/integration/InMoovScriptTest.java
```diff
@@ -115,6 +119,7 @@ public void testInMoovMinimalHead() throws IOException {
   // Test the inmoov minimal arm script.
   @Test
   public void testInMoovMinimalTorso() throws IOException {
+    if (printMethods)System.out.println(String.format("Running %s.%s", getSimpleName(), getName()));
```
kwatters: this sort of debug message should be refactored into the base class... or removed altogether.

src/main/java/org/myrobotlab/service/interfaces/PinDefinition.java
```diff
@@ -5,9 +5,11 @@
 import com.pi4j.io.gpio.GpioPinDigitalMultipurpose;
 
 public class PinDefinition extends SensorDefinition implements Serializable {
```
kwatters: i think we should rename this class to be "Pin"

src/main/java/org/myrobotlab/service/interfaces/EncoderControl.java
```diff
@@ -11,8 +11,21 @@
 
   // TODO: publish encoder data that includes the name/id of the device.
   public void onEncoderData(EncoderData data);
```
kwatters: shouldn't an encoder control publishEncoderData , not subscribe to it?

src/main/java/org/myrobotlab/service/interfaces/Microcontroller.java
```diff
@@ -25,4 +25,8 @@
    * @return list of board types
    */
   public List<BoardType> getBoardTypes();
+  
+  public void reset();
```
kwatters: awesome.... yes, we should be able to reset microcontrollers...

PR #485, opened 2019-04-06
Oscope fixes

supertick:


src/main/java/org/myrobotlab/service/Arduino.java
```diff
@@ -977,8 +978,8 @@ public Integer getMrlPinType(PinDefinition pin) {
         } else if (i > 53) {
           pinName = String.format("A%d", i - 54);
           pindef.setAnalog(true);
-          pindef.setDigital(false);
-          pindef.canWrite(false);
+          pindef.setDigital(true);
```
kwatters: I'm confused, shouldn't setDigital be false here?  It's an analog pin...

supertick: No - you can do a digitalWrite to analog pins - and since this is currently controlling what the UI pin interface can do - we want the option to do that given to the user.

https://www.arduino.cc/en/Tutorial/AnalogInputPins

kwatters: ok.. so digital write can work on either an analog or a digital pin, and the "setDigital" true is really saying that the pin supports digitalWrite

PR #488, opened 2019-04-07
beginning of servo fixes and servo2

supertick:


src/main/java/org/myrobotlab/service/IntegratedMovement.java
```diff
@@ -551,11 +551,11 @@ public void setDHLink(String arm, String name, double d, double theta, double r,
     }
   }
 
-  public void setDHLink(String arm, ServoControl servo, double d, double theta, double r, double alpha) {
+  public void setDHLink(String arm, Servo servo, double d, double theta, double r, double alpha) {
```
kwatters: why the change from ServoControl to Servo here?

src/main/java/org/myrobotlab/service/IntegratedMovement.java
```diff
@@ -551,11 +551,11 @@ public void setDHLink(String arm, String name, double d, double theta, double r,
     }
   }
 
-  public void setDHLink(String arm, ServoControl servo, double d, double theta, double r, double alpha) {
+  public void setDHLink(String arm, Servo servo, double d, double theta, double r, double alpha) {
     setDHLink(arm, servo, d, theta, r, alpha, servo.getMinInput(), servo.getMaxInput());
   }
 
-  public void setDHLink(String arm, ServoControl servo, double d, double theta, double r, double alpha, double minAngle, double maxAngle) {
+  public void setDHLink(String arm, Servo servo, double d, double theta, double r, double alpha, double minAngle, double maxAngle) {
```
kwatters: Same question. why Servo instead of ServoControl here?

src/main/java/org/myrobotlab/service/Servo2.java
```diff
@@ -0,0 +1,1022 @@
...
+ * 
+ *         outputY - is the values sent to the firmware, and should not
+ *         necessarily be confused with the inputX which is the input values
+ *         sent to the servo
+ *         
+ *         FIXME - inherit from AbstractMotor ..
+ * 
+ */
+
+public abstract class Servo2 extends Service implements ServoControl {
```
kwatters: If this is abstract, we should call this class AbstractServo ...

src/main/java/org/myrobotlab/service/interfaces/EncoderControl.java
```diff
@@ -10,7 +10,7 @@
   public void setController(EncoderController controller);
 
   // TODO: publish encoder data that includes the name/id of the device.
-  public void onEncoderData(EncoderData data);
+  // public void onEncoderData(EncoderData data);
```
kwatters: shouldn't this be renamed to publishEncoderData(EncoderData data)  ?

src/main/resources/resource/Arduino/generate/arduinoMsgs.schema
```diff
@@ -141,5 +141,5 @@
 # Amt203=0 / As5048A=1 Encoder methods
 > encoderAttach/deviceId/type/pin
 > setZeroPoint/deviceId
-< publishEncoderPosition/deviceId/b16 position
+< publishEncoderData/deviceId/b16 position
```
kwatters: I didn't see the MrlComm version number get incremented with this change..  seems like we should bump the version if we've made a chance like this.

src/main/java/org/myrobotlab/service/IntegratedMovement.java
supertick: it should be ServoControl - but I was struggling elsewhere, as Servo implemented some critical  methods that were not in the interface ... I'll correct it .

src/main/java/org/myrobotlab/service/Servo2.java
supertick: its not ... I have yet to make the abstract - this is the beginning of an implementation
I'll get Servo worky again as it did ... but I'm not happy with Servo's implementation - So I'm making a Servo2 - they will share the same ServoControl interface, and after I'm happy with that, I'll push common code into an Abstract

src/main/java/org/myrobotlab/service/IntegratedMovement.java
supertick: Ah .. another reason is larger reason is all the "IK" awareness Servo has .. I "**don't**" want that propagated to the interface ... so this IK class has to use Servo directly

src/main/java/org/myrobotlab/service/interfaces/EncoderControl.java
supertick: no the EncoderController does the publishing

src/main/resources/resource/Arduino/generate/arduinoMsgs.schema
supertick: Oooh definitely ...  nice catch !

src/main/java/org/myrobotlab/service/Servo2.java
kwatters: but.. the class is defined as "public abstract class Servo2..."   if the class isn't supposed to be abstract you should remove the "abstract" in the class definition.

src/main/java/org/myrobotlab/service/interfaces/EncoderControl.java
kwatters: ok.. but ServoController isn't responsible for publishing the data.. that's done by Servo currently.  So this would be a change from that pattern.  Seems like the EncoderControl should be the one publishing..   this makes the Controller more transparent..  

.. as a user..  I would like to listen to servo events from the servo.  

is different than

.. as a user..  I would like to have a servo and I will get events from the controller. 

The first statement seems much simpler and easier to explain to people.

src/main/java/org/myrobotlab/service/Servo2.java
supertick: Ha !  I think I spent 5 minutes trying to figure out when I added more functions to the interface it inherits why it didn't tell me it had to implement them ... Ahahaha - I think I was trying to do too many things at once..   I do want to create an abstract at some point but not yet ..
Thanks for pointing it out.

src/main/java/org/myrobotlab/service/interfaces/EncoderControl.java
supertick: I agree ..

I think I went overboard in hating ServoEvents.  I knew a piece was missing (real sensor data).

Now, I'm starting to warm up to the idea to ServoEvents... but it takes EncoderControl, EncoderControllers & EncoderListeners to support them.

I agree EncoderControl should do the publishing - EncoderControllers do the callback into the Control.   And listeners use and handle callbacks from the control.

Yep this has been the pattern, I got too enthusiastic with my hacking away ;)

Will correct ...  again thanks for the catch

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -0,0 +1,764 @@
...
+
+    } catch (Exception e) {
+      log.error("main threw", e);
+    }
+  }
+
+  /**
+   * acceleration of servo - not really implemented
+   */
+  protected Double acceleration;
```
kwatters: should be a primitive and defaulted to 0.0.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -0,0 +1,764 @@
...
+
+  /**
+   * maximum speed
+   */
+  protected Double maxSpeed;
+
+  /**
+   * status if the servo thinks its moving ..
+   */
+  protected boolean moving;
```
kwatters: seems difficult to compute as know as a boolean.. this should be removed as the logic for computing that a servo is moving would require testing if it's sweeping.. or if it's moving to a position with a set velocity..

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -0,0 +1,764 @@
...
+
+    // update pos if non-null value supplied
+    if (pos != null) {
+      targetPos = pos;
+      lastPos = targetPos;
+    }
+
+    // update speed if non-null value supplied
+    if (speed != null) {
+      setSpeed(speed);
```
kwatters: velocity..

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -0,0 +1,764 @@
...
+        ((ServoController) Runtime.getService(controller)).detach(this);
+      } else {
+        send(controller, "detach", this);
+      }
+    }
+  }
+
+  @Override
+  public void detach(ServoController controller) {
+    if (controller == null) {
```
kwatters: inconsistent disable behavior for the detach() methods.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -0,0 +1,764 @@
...
+      if (si.isLocal()) {
+        ((ServoController) Runtime.getService(controller)).servoEnable(this);
+      } else {
+        send(controller, "servoEnable", this);
+      }
+    }
+  }
+
+  @Override
+  public void enable(Integer pin) {
```
kwatters: these should be deprecated..  enabled method shouldn't take any args.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -0,0 +1,764 @@
...
+    this.rest = rest;
+  }
+
+  @Override
+  public void setRest(Integer rest) {
+    this.rest = rest.doubleValue();
+  }
+
+  @Override
+  public void setSpeed(Double degreesPerSecond) {
```
kwatters: I think we need to maintain the backwards compatibility for the old skewl setSpeed . our new proper method is setVelocity which takes a double that specifies  the angular velocity of the servo that we will try to achieve.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -0,0 +1,764 @@
...
+
+  @Override
+  public void setSpeed(Integer degreesPerSecond) {
+    setSpeed(degreesPerSecond.doubleValue());
+  }
+
+  @Override
+  @Deprecated /* use setSpeed */
+  public void setVelocity(Double speed) {
+    setSpeed(speed);
```
kwatters: as my many other comments call out..  this will confuse the hell out of a lot of people...  we went through so much back and forth on this.  we finally had consensus across the community about the behaviors of setSpeed and setVelocity.   many scripts rely on the behavior of setSpeed ...

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -0,0 +1,764 @@
...
+    super.stopService();
+    disable();
+  }
+
+  @Override
+  public void sync(ServoControl sc) {
+    if (sc == null) {
+      log.error("{}.sync(null)", getName());
+    }
+    syncedServos.add(sc.getName());
```
kwatters: test to not add yourself to a synced servo list.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -0,0 +1,764 @@
...
+  public void unsync(ServoControl sc) {
+    if (sc == null) {
+      log.error("{}.unsync(null)", getName());
+    }
+    syncedServos.remove(sc.getName());
+  }
+
+  @Override
+  public void waitTargetPos() {
+    // TODO Auto-generated method stub
```
kwatters: seems like this is a method that needs to block until a servo reaches the last requested position.

src/main/java/org/myrobotlab/service/interfaces/HttpDataListener.java
```diff
@@ -1,8 +1,9 @@
 package org.myrobotlab.service.interfaces;
 
+import org.myrobotlab.framework.interfaces.NameProvider;
 import org.myrobotlab.service.data.HttpData;
 
-public interface HttpDataListener {
+public interface HttpDataListener extends NameProvider {
```
kwatters: if NameProvider is an interface.. this class should implement it .. not extend it.

src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
```diff
@@ -120,150 +168,249 @@
...
+   *          - min input
+   * @param maxX
+   *          - max input
+   * @param minY
+   *          - min output
+   * @param maxY
+   *          - max output
    */
-  double getAcceleration();
+  void map(Double minX, Double maxX, Double minY, Double maxY);
```
kwatters: I would much prefer just see a simple set /get Mapper(Mapper m) in the interface rather than having to pass 4 doubles.

src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
```diff
@@ -120,150 +168,249 @@
...
 
   /**
-   * unsync a servo
+   * control message publishing moveTo
    * 
    * @param sc
+   * @return
    */
-  void unsync(ServoControl sc);
+  ServoControl publishMoveTo(ServoControl sc);
```
kwatters: shouldn't this publish some sort of a servoEvent not a the actual servo itself...

src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
```diff
@@ -120,150 +168,249 @@
...
-   *          A default position for the servo. Defaulted to 90 unless
-   *          explicitly set. Position the servo will move to when method
-   *          servo.rest() is called
+   * 
+   * @param eventType
+   * @param currentPosUs
+   * @return
    */
-  void setRest(double rest);
+  ServoData publishServoData(ServoStatus eventType, Double currentPosUs);
```
kwatters: this should take servo data .. not eventtype and currentposus.

src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
```diff
@@ -120,150 +168,249 @@
...
+   * Set the acceleration of the servo
    * 
-   * @param invert
-   *          - true is to invert
+   * @param acceleration
    */
-  void setInverted(boolean invert);
+  void setAcceleration(Double acceleration);
+  
+  void setAcceleration(Integer acceleration);
```
kwatters: not sure why we would allow for an integer to be passed as acceleration

src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
```diff
@@ -120,150 +168,249 @@
...
+   * move their full range
+   * 
+   * @param min
+   *          min value
+   * @param max
+   *          max value
    * 
    */
-  void waitTargetPos();
+  void setMinMax(Double min, Double max);
```
kwatters: min/max what?  input? output? very confusing.

src/main/java/org/myrobotlab/swing/ServoGui.java
```diff
@@ -679,8 +679,8 @@ public void run() {
         }
 
         // In the inverted case, these are reversed
-        slider.setMinimum((int) servo.getMin());
-        slider.setMaximum((int) servo.getMax());
+        slider.setMinimum(servo.getMin().intValue());
```
kwatters: if getMin() returns null.  this could blow a npe.)

src/main/java/org/myrobotlab/swing/ServoGui.java
```diff
@@ -726,8 +726,8 @@ public void run() {
 
         mapInputSlider.setLowValue((int) servo.getMinInput());
         mapInputSlider.setHighValue((int) servo.getMaxInput());
-        mapOutputSlider.setLowValue((int) servo.getMinOutput());
-        mapOutputSlider.setHighValue((int) servo.getMaxOutput());
+        mapOutputSlider.setLowValue(servo.getMinOutput().intValue());
```
kwatters: this is odd syntax... semi-colons between args?

src/test/java/org/myrobotlab/service/HttpClientTest.java
```diff
@@ -0,0 +1,99 @@
...
+   * 
+   */
+  /*
+   * @Override //FIXME - add blocking queue - to verify public void
+   * onHttpResponse(String data) { log.info(data); queue.add(data); }
+   */
+
+  @Override
+  public void testService() throws Exception {
+    HttpClient http = (HttpClient) Runtime.start("http", "HttpClient");
```
kwatters: this should be created in the createService() method.  this.service is the httpClient that you created already.

src/test/java/org/myrobotlab/service/HttpClientTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    // this is how a listener might subscribe
+    // TODO - put dynamically subscribing into framework
+    // with interface inspection ??
+
+    // FIXME - add make the attach !
+    http.attach(catcher);
+
+    catcher.clear();
+    String data = http.get("https://postman-echo.com/get?foo1=bar1&foo2=bar2");
```
kwatters: this test requires an internet connection to pass making it unreliable.

src/test/java/org/myrobotlab/service/PidTest.java
```diff
@@ -4,17 +4,22 @@
...
-  @Test
-  public void testCompute() {
+  @Override
+  public Service createService() {
+    return (Service) Runtime.start("pid", "Pid");
+  }
+
+  @Override
+  public void testService() throws Exception {
     Pid pid = (Pid) Runtime.start("pid", "Pid");
```
kwatters: this.service has the pid that was created in the createService method.

src/test/java/org/myrobotlab/test/AbstractTest.java
```diff
@@ -14,6 +17,7 @@
 import org.junit.Rule;
 import org.junit.rules.TestName;
 import org.myrobotlab.framework.Platform;
+import org.myrobotlab.framework.interfaces.Attachable;
```
kwatters: seems like only services are attachable.. this attachable stuf shouldn't be in the base AbstractTest class.. it should be up one layer in AbstractServiceTest.

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1434,6 +1434,9 @@ final public Object invokeOn(Object obj, String method, Object... params) {
       // if so we need to search
 
       // FIXME - WHY ISN'T METHOD CACHING USED HERE !!!
+      if (method.equals("attach")) {
```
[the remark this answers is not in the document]
supertick: removed

src/main/java/org/myrobotlab/service/Adafruit16CServoDriver.java
```diff
@@ -140,7 +140,7 @@ public void run() {
 
   public double publishServoEvent(ServoControl servo, Integer eventType, double currentOutput) {
```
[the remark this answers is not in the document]
supertick: agreed - change publishServoData with ServoData as the name of class to contain details,
also changed ServoListener to ServoDataListener if that is correct

src/main/java/org/myrobotlab/service/DiyServo.java
```diff
@@ -929,20 +923,14 @@ public boolean getAutoDisable() {
   }
 
   @Override
-  public boolean moveToBlocking(double pos) {
+  public void moveToBlocking(double pos) {
```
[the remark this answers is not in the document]
supertick: ok changed to Double since this method is in the org.myrobotlab.org.service.interfaces package - done

src/main/java/org/myrobotlab/encoder/TimerEncoder.java
```diff
@@ -0,0 +1,24 @@
...
+  // resolution
+  
+  // start - stop
+  
+  // move from x to y at velocity z
+  
+  // use acceleration - slope - max velocity
+  
+  public void attach(Attachable attachable) throws Exception {
+
```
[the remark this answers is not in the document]
supertick: it will .. I have to finish it in order to provide the a transportable "waitForPos()"  for all controllers  moveToBlocking will depend on it ... ya needs code :)

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1434,9 +1438,9 @@ final public Object invokeOn(Object obj, String method, Object... params) {
       // if so we need to search
 
       // FIXME - WHY ISN'T METHOD CACHING USED HERE !!!
-      if (method.equals("attach")) {
-        log.info("here");
-      }
+      // if (method.equals("moveTo")) {
```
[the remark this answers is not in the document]
supertick: done

src/main/java/org/myrobotlab/jme3/Jme3ServoController.java
```diff
@@ -118,7 +118,7 @@ public void servoMoveTo(ServoControl servo) {
       log.error("servoMoveTo({})", servo);
       return;
     }
-    float velocity = (float) servo.getVelocity();
+    double velocity = servo.getSpeed();
```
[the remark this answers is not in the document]
supertick: getVelocity setVelocity is the wrong word..
this is "net-new" implementation .. Servo can have both, but new stuff should have the right method name with the correct implementation.

not changing

src/main/java/org/myrobotlab/kinematics/DHLink.java
```diff
@@ -325,11 +325,11 @@ public void setVelocity(double velocity) {
   }
 
   public Integer getState() {
```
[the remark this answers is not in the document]
supertick: removed

src/main/java/org/myrobotlab/kinematics/IMEngine.java
```diff
@@ -583,8 +583,8 @@ public void updateLinksPosition(ServoData data) {
     for (DHLink l : arm.getLinks()) {
       if (l.getName().equals(data.name)) {
         l.addPositionValue(data.pos);
-        l.setState(data.state);
-        l.setVelocity(data.velocity);
+        // l.setState(data.state);
+        l.setVelocity(data.speed);
```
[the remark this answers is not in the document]
supertick: internal I'm still of the belief setSpeed is correct - previously setSpeed was the wrong implementation (mine) but has since been corrected.  velocity is the incorrect name.

src/main/java/org/myrobotlab/swing/ServoGui.java
supertick: ?

src/test/java/org/myrobotlab/service/HttpClientTest.java
supertick: I'm not sure I understand the benefit of creating services in  @Before setup although we do it all the time.   

I think potentially there is a benefit of starting all services necessary at the beginning of a test.  The test is potentially fully atomic ... no special setup needs to be run in order to do the test, yet if this used the static reference - the method would throw an NPE

I removed the reference in setup for the reasons above.

AbstractTest already did releasing of services, additionally it keeps track of warrant threads and reports on them.

So, I ended up removing everything except the test itself.

src/test/java/org/myrobotlab/service/HttpClientTest.java
supertick: fixed

src/test/java/org/myrobotlab/service/PidTest.java
supertick: not sure what benefit there is changing the regular junit cycle into a different one.

src/test/java/org/myrobotlab/test/AbstractTest.java
supertick: I'm not sure if only Services are attachable...

supertick:
Regarding setVelocity setSpeed

- setVelocity is still in ServoControl
- I didn't change the implementation of either
- setSpeed has the same implementation as its previous release which is effectively setVelocity - most implementations of setSpeed just call setVelocity

src/main/java/org/myrobotlab/service/Amt203Encoder.java
```diff
@@ -25,16 +25,17 @@
  * @author kwatters
  *
  */
-public class Amt203Encoder extends AbstractEncoder implements EncoderControl {
+public class Amt203Encoder extends AbstractPinEncoder implements EncoderControl {
```
kwatters: As mentioned in the shoutbox.  Amt203Encoder isn't really pin based.. it's SPI based.  I don't think this is a good name for the base class.  The base class should just be AbstractEncoder.

src/main/java/org/myrobotlab/service/Arduino.java
```diff
@@ -66,6 +66,8 @@
 import org.myrobotlab.service.interfaces.UltrasonicSensorControl;
 import org.myrobotlab.service.interfaces.UltrasonicSensorController;
 
+import akka.event.AddressTerminatedTopic;
```
kwatters: What is this import needed for?

src/main/java/org/myrobotlab/service/As5048AEncoder.java
```diff
@@ -2,18 +2,16 @@
...
-import org.myrobotlab.service.interfaces.EncoderController;
 
 /**
  * AS5048A - SPI based 14 bit magnetic absolute position encoder.
  * 
  * @author kwatters
  *
  */
-public class As5048AEncoder extends AbstractEncoder implements EncoderControl {
+public class As5048AEncoder extends AbstractPinEncoder implements EncoderControl {
```
kwatters: Same comment as before.  I think the base class should just be AbstractEncoder.    This is an SPI based encoder.  It's not based on a Pin.

src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -1631,6 +1631,12 @@ public void setSpeed(Double speed) {
     // TODO Auto-generated method stub
     
   }
+
+  @Override
+  public List<String> refreshEncoders() {
```
kwatters: This method should not be here.  If we want a way to get all "encoders" we should ask Runtime for them.

src/main/java/org/myrobotlab/service/TimeEncoder.java
```diff
@@ -31,28 +51,39 @@ public TimeEncoder(String n) {
...
     meta.addCategory("general");
     return meta;
   }
 
   @Override // from EncoderController
-  public void attach(EncoderControl control, Integer pin) throws Exception {
+  public EncoderData publishEncoderData(EncoderData data) {
     // TODO Auto-generated method stub
-    
+    return null;
```
kwatters: this should return the data. not null.

src/main/java/org/myrobotlab/service/TimeEncoder.java
```diff
@@ -31,28 +51,39 @@ public TimeEncoder(String n) {
...
   }
 
-  @Override // from EncoderController
-  public EncoderData publishEncoderData(EncoderData data) {
+  public void attach(EncoderController controller) throws Exception {
     // TODO Auto-generated method stub
-    return null;
+
```
kwatters: needs implementation.

src/main/java/org/myrobotlab/service/TimeEncoder.java
```diff
@@ -31,28 +51,39 @@ public TimeEncoder(String n) {
...
   }
 
-  @Override // from EncoderController
-  public void setZeroPoint(EncoderControl encoder) {
+  @Override
+  public void disable() {
```
kwatters: needs implementation.

src/main/java/org/myrobotlab/service/TimeEncoder.java
```diff
@@ -31,28 +51,39 @@ public TimeEncoder(String n) {
...
     // TODO Auto-generated method stub
-    
+
+  }
+
+  @Override
+  public void enable() {
```
kwatters: needs implementation.

src/main/java/org/myrobotlab/service/TimeEncoder.java
```diff
@@ -31,28 +51,39 @@ public TimeEncoder(String n) {
...
+    // TODO Auto-generated method stub
+
+  }
+
+  @Override
+  public Boolean isEnabled() {
+    // TODO Auto-generated method stub
+    return null;
```
kwatters: needs implementation should not return null.  actually this method really shouldn't return a Boolean.  In general , return values from our interfaces ,  I feel should be primitives , especially in the case of boolean.

src/main/java/org/myrobotlab/service/abstracts/AbstractPinEncoder.java
```diff
@@ -6,41 +6,41 @@
 import org.myrobotlab.service.interfaces.EncoderControl;
 import org.myrobotlab.service.interfaces.EncoderController;
 
-public class AbstractEncoder extends Service implements EncoderControl {
+public class AbstractPinEncoder extends Service implements EncoderControl {
```
kwatters: poor name for the class.  Amt and AS 203 encoders are not "pin" based.  They are typically I2C or SPI based.  That doesn't map to a concept of a Pin in reality.  more like a bus address instead, especially for the i2c variants.

src/main/java/org/myrobotlab/service/abstracts/AbstractPinEncoder.java
```diff
@@ -78,37 +78,32 @@ public void setZeroPoint() {
...
-  @Override
   public void setPin(Integer address) {
     this.pin = String.format("%d", address);
   }
 
   @Override
   public void disable() {
-    // TODO Auto-generated method stub
-    
+    enabled = false;
```
kwatters: This isn't correct.. This should actually send a message to the Amt203 or As5048 encoders in MrlComm land to make sure the devices are disabled.  Otherwise data will continue to flow over the serial port from the arduino and could clog up communication.    this should be should be abstract.

src/main/java/org/myrobotlab/service/abstracts/AbstractPinEncoder.java
```diff
@@ -78,37 +78,32 @@ public void setZeroPoint() {
...
   }
 
   @Override
   public void enable() {
-    // TODO Auto-generated method stub
-    
+    enabled = true;
```
kwatters: this doesn't actually tell the underlying encoder to be enabled.  as mentioned in other comments.  this should be abstract

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -2034,6 +2037,80 @@ synchronized public void startService() {
...
+      Method method = this.getClass().getMethod("getMetaData");
+      ServiceType st = (ServiceType) method.invoke(null);
+      peers = st.getPeers();
+    } catch (Exception e) {
+    }
+
+    if (peers == null) {
+      return;
+    }
+
```
kwatters: Why aren't we just calling Runtime.start here?  I'm not fully grokin' the need for the logic below.

src/main/java/org/myrobotlab/service/HobbyServo.java
```diff
@@ -114,33 +113,22 @@ public long getLastActivityTime() {
...
-    
-  }
-
-  @Override
+  @Deprecated /* bleh - speed is the right method name */
   public void setSpeed(Double speed) {
-    // TODO Auto-generated method stub
+    // These 2 values can be tweaked for a slightly different curve that
+    // fits the observed data.
+    double slope = 3.25;
```
kwatters: seems like this should be pulled out into a static helper method and referenced as a place holder until we can remove all references to old style speed control.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1785,9 +1759,10 @@ public void loadCalibration(String calibrationFilename) {
   }
 
   public void saveCalibration() {
-    saveCalibration(CALIBRATION_FILE);
+    // saveCalibration(CALIBRATION_FILE);
```
kwatters: I use this.. let's not remove it.

src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -2409,6 +2287,10 @@ public static void main(String[] args) throws Exception {
     i01.execGesture("BREAKITdaVinci()");
   }
 
+  public void startVinMoov() {
```
kwatters: This is way beyond the scope of servo fixes.  why are we changing this in the servo fix pull request?

src/main/java/org/myrobotlab/service/InMoovArm.java
```diff
@@ -261,22 +246,6 @@ public void waitTargetPos() {
   // FIXME - releasePeers()
   public void release() {
     disable();
```
kwatters: disable is different than release is it not?  seems like this really should just be removed and we should use releaseService here?

src/main/java/org/myrobotlab/service/InMoovEyelids.java
```diff
@@ -210,14 +200,6 @@ public void waitTargetPos() {
   // FIXME - releasePeers()
   public void release() {
     disable();
```
kwatters: we should remove this method.. and just use enable/disable.  if the release() method is intended to be releaseService.. then references should be updated to call releaseService instead.

src/main/java/org/myrobotlab/service/InMoovEyelids.java
```diff
@@ -236,23 +218,19 @@ public boolean save() {
   @Override
   public void startService() {
     super.startService();
-    eyelidleft.startService();
-    eyelidright.startService();
+    eyelidleft.setRest(0.0);
```
kwatters: seems like there should be somewhere else in the life cycle to put this. we shouldn't be adding hard coded values in the startService method here.

src/main/java/org/myrobotlab/service/InMoovHand.java
```diff
@@ -128,12 +110,7 @@ public void bird() {
   @Override
   public void broadcastState() {
     // notify the gui
-    thumb.broadcastState();
-    index.broadcastState();
-    majeure.broadcastState();
-    ringFinger.broadcastState();
-    pinky.broadcastState();
-    wrist.broadcastState();
+    Runtime.broadcastStates();
```
kwatters: why is this runtime broadcast state? shouldn't it be super.broadcastState()  .. perhaps this should just be removed from here and rely on the base class.

src/main/java/org/myrobotlab/service/InMoovHand.java
```diff
@@ -544,20 +503,21 @@ public void startLeapTracking() throws Exception {
...
+  public void startService() { 
     super.startService();
-    thumb = (Servo) startPeer("thumb");
-    thumb.startService();
-    index.startService();
-    majeure.startService();
-    ringFinger.startService();
-    pinky.startService();
-    wrist.startService();
+
```
kwatters: TODO: remove this hard coded stuff out of the startService method.

src/main/java/org/myrobotlab/service/InMoovHead.java
```diff
@@ -102,12 +79,7 @@ public boolean enable() {
   @Override
   public void broadcastState() {
     // notify the gui
-    rothead.broadcastState();
-    rollNeck.broadcastState();
-    neck.broadcastState();
-    eyeX.broadcastState();
-    eyeY.broadcastState();
-    jaw.broadcastState();
+    Runtime.broadcastStates();
```
kwatters: same comment as above.. why does this need to be overridden here?  seems like the base class should handle this method.

src/main/java/org/myrobotlab/service/InMoovHead.java
```diff
@@ -287,12 +257,6 @@ public void waitTargetPos() {
 
   public void release() {
```
kwatters: should remove the release method if it really should be releaseService.. and if it really should be releaseService.. it should releaseService.

src/main/java/org/myrobotlab/service/InMoovHead.java
```diff
@@ -438,22 +373,49 @@ public void setSpeed(Double headXSpeed, Double headYSpeed, Double eyeXSpeed, Dou
...
     }
 
-    jaw.startService();
-    eyeX.startService();
-    eyeY.startService();
-    rothead.startService();
-    neck.startService();
-    rollNeck.startService();
-  }
+    if (jaw == null) {
```
kwatters: all this hard coded configuration stuff should be moved elsewhere form startService. from a lifecycle perspective i would see it as 3 separate phases.   1. create service .   2. configure service .. 3. start service.    
having this code here forces 3 and 2 to occur no matter what.

src/main/java/org/myrobotlab/service/InMoovTorso.java
```diff
@@ -102,18 +102,10 @@ public void setAutoDisable(Boolean param) {
...
-    lowStom.setOverrideAutoDisable(param);
-  }
-
   @Override
   public void broadcastState() {
     // notify the gui
-    topStom.broadcastState();
-    midStom.broadcastState();
-    lowStom.broadcastState();
+    Runtime.broadcastStates();
```
kwatters: same question as before.. why isn't this handled in the base class.

src/main/java/org/myrobotlab/service/IntegratedMovement.java
```diff
@@ -559,7 +559,7 @@ public void setDHLink(String arm, ServoControl servo, double d, double theta, do
     if (engines.containsKey(arm)) {
       IMEngine engine = engines.get(arm);
       DHLink dhLink = new DHLink(servo.getName(), d, r, MathUtils.degToRad(theta), MathUtils.degToRad(alpha));
-      servo.addIKServoEventListener(this);
+      // servo.addIKServoEventListener(this);
```
kwatters: This is a breaking change.  At this point, the IntegratedMovement service should subscribe to the servo events instead.  (same data is passed in ServoData)

src/main/java/org/myrobotlab/service/IntegratedMovement.java
```diff
@@ -573,7 +573,7 @@ public void setDHLink(String arm, ServoControl servo, double d, double theta, do
       dhArm.addLink(dhLink);
       engine.setDHRobotArm(dhArm);
       engines.put(arm, engine);
-      servo.subscribe(getName(), "publishAngles", servo.getName(), "onIMAngles");
+      // servo.subscribe(getName(), "publishAngles", servo.getName(), "onIMAngles");
```
kwatters: need a way to publish the solved angles from IK/IM service to the appropriate servo.  it should probably invoke moveTo .. intead of onIMAngles.  (what ever that method was..)

src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
```diff
@@ -29,10 +29,11 @@
 import java.util.Set;
 
 import org.myrobotlab.framework.interfaces.Attachable;
+import org.myrobotlab.framework.interfaces.StateSaver;
 import org.myrobotlab.math.Mapper;
 import org.myrobotlab.service.interfaces.ServoData.ServoStatus;
 
-public interface ServoControl extends AbsolutePositionControl, Attachable {
+public interface ServoControl extends AbsolutePositionControl, Attachable, StateSaver, org.myrobotlab.framework.interfaces.StatePublisher {
```
kwatters: do we need the fully qualified class name here? why not just import?  do we have a class name conflict with StatePublisher?

src/main/java/org/myrobotlab/service/InMoovHand.java
```diff
@@ -591,7 +591,7 @@ static public ServiceType getMetaData() {
     meta.addPeer("pinky", "Servo", "Pinky servo");
     meta.addPeer("wrist", "Servo", "Wrist servo");
     meta.addPeer("arduino", "Arduino", "Arduino controller for this arm");
-    meta.addPeer("leap", "LeapMotion", "Leap Motion Service");
+    meta.addPeer("leap", "LeapMotion", "Leap Motion Service", false);
```
kwatters: instead of adding this additinal boolean to addPeer, i'd prefer to see the leap motion service get fixed so it doesn't kill the jvm.  we should consider removal of this additional boolean here..

src/main/java/org/myrobotlab/service/InMoovHead.java
```diff
@@ -97,12 +97,14 @@ public boolean connect(String port, Integer headYPin, Integer headXPin, Integer
         error("controller for head is not connected");
       }
     }
-    ((Servo) neck).attach(controller, headYPin, neck.getRest(), neck.getVelocity());
-    ((Servo) rothead).attach(controller, headXPin, rothead.getRest(), rothead.getVelocity());
-    ((Servo) jaw).attach(controller, jawPin, jaw.getRest(), jaw.getVelocity());
-    ((Servo) eyeX).attach(controller, eyeXPin, eyeX.getRest(), eyeX.getVelocity());
-    ((Servo) eyeY).attach(controller, eyeYPin, eyeY.getRest(), eyeY.getVelocity());
-    ((Servo) rollNeck).attach(controller, rollNeckPin, rollNeck.getRest(), rollNeck.getVelocity());
+
```
kwatters: I guess the assumption is that if you want to use a DiyServo here, it would just ignore the attach method?

src/main/java/org/myrobotlab/service/Amt203Encoder.java
supertick: When I met it - it required a pin .. there were references to either - encoder.setPin(3); or encoder.pin = 3; ... 
I guess we both agree this shouldn't really be in the interface as is part of configuration. Its possible an encoder could have x pins to properly configure or no pins (as the case with TimeEncoder) ..  so I would think you would agree it shouldn't be in AbstractEncoder - hence the move

src/main/java/org/myrobotlab/service/Arduino.java
supertick: long gone .. dunno how that got in there ..

src/main/java/org/myrobotlab/service/As5048AEncoder.java
supertick: Same as above too ...
You can get rid of AbstractPinEncoder .. change the name to AbstractI2CEncoder or remove it completely - but previously the AbstractEncoder was not devoid of pin references ..
Now, its gone ready to be created without pin references

src/main/java/org/myrobotlab/service/Servo.java
supertick: I'd be ok with removing it ...   same **should** go for refreshControllers too which has been part of ServoControl  .. if encoder goes controllers should go  (means some UI work at least...)

src/main/java/org/myrobotlab/service/TimeEncoder.java
supertick: barely begun - I've just created the class ;)

src/main/java/org/myrobotlab/service/TimeEncoder.java
supertick: Ya .. I know

src/main/java/org/myrobotlab/service/TimeEncoder.java
supertick: oye

src/main/java/org/myrobotlab/service/abstracts/AbstractPinEncoder.java
supertick: There was confusion I suspect between the interface and the classes - I think it was pretty incomplete.  It looks like I did the original check-in (dunno where from but i'm sure I copy/pasted or just guessed on the original class) - You in-turn did the first implementations of it.  Obviously, it needs refining.  Feel free to change the name back, and remove any pin references, I was only trying **not** to break your existing implementations.

src/main/java/org/myrobotlab/service/abstracts/AbstractPinEncoder.java
supertick: same as above

src/main/java/org/myrobotlab/service/abstracts/AbstractPinEncoder.java
supertick: same as above - my sloppy first interface definition, meets your first implementation(s)

src/main/java/org/myrobotlab/framework/Service.java
supertick: this is starting peers - there's a possibility of  not having getMetaData, nor having any peers ..  if this service has no peers, we are done ... 

I'd say that's better than an NPE

src/main/java/org/myrobotlab/service/HobbyServo.java
supertick: I supposed.
HobbyServo is "net new" - so I'd prefer that it get one step closer to the  

- correct implementation
- correct method name

The next step in that sequence would be setSpeed has the correct implementation - my preference would be to remove this incorrect albiet backwards-compatible implementation behind and basically be the same as setVelocity.

I know there will be about 8 steps to get back to correct implementation with correct name .. but shouldn't we take the first step with "new" stuff ?

src/main/java/org/myrobotlab/service/InMoov.java
supertick: fine .. although mixing python generation with a Java service that is not related to Python at all gives me bumps ... I think it should use json in a declarative fashion to get back to a worky state

src/main/java/org/myrobotlab/service/InMoov.java
supertick: if you run down the rabbit hole, you'll find servo, servoevents, IK and jme intertwined.
I'm untangling them.

src/main/java/org/myrobotlab/service/InMoovArm.java
supertick: whoa .. not that i see it I don't know how the hell this came about..
it should be 
public void releaseService {
     super.releaseService();
     disable();

before .. I wrote a long comment about the problems of using polymorphism for framework lifecycle.   Its problematic for the developer to always remember to call super.releaseService() ..
So I said "blah blah mebbe the framework should always call some other method name which effectively does super.releaseService() and **THEN**  call this function.  That way developers don't need to remember to call the super .. but I then realized this is just whack ...

Ahahah  5 years ago - there was never a releaseService and this NEVER got called .... ever .. 

fixing after 5 years ..

supertick: pushed

src/main/java/org/myrobotlab/service/InMoovHead.java
supertick: agree in general
Create service means creating peers or not ?  I prefer not.
If you taking the trouble to create a service and not using the default peers or config. Then you should be handed a thin skeleton of the composite service.  Allowing you to set and modify to your delight WITHOUT having to release services that were auto-magically created for you.

To not burden the consumer of releasing pre-existing services, i thought late binding to peers would be a good thing (ie peers created and configured at startService if not already set by previous customizations)

Since peers are not created automagically untill startService - these would all be NPEs

I guess you're arguing to create and configure it all, then make the user tear it down & release it - if they want to do customizations ..

I'm on the fence, but creation does mean fully created - which means I'm prolly leaning towards your side.

src/main/java/org/myrobotlab/service/InMoovTorso.java
supertick: the question and answer was never raised - 
does a broadcast of the service mean broadcast of the peers .. it "could" its just never been implemented.

src/main/java/org/myrobotlab/service/IntegratedMovement.java
supertick: IM service polluted several other services in its implementation (Servo, JME, InMoov, ...)
Priority is to fix up the other services and contain the problem.
Fixing IM is lower priority.

src/main/java/org/myrobotlab/service/IntegratedMovement.java
supertick: absolutely ...
this should have used "command" messages, instead it used status message.  Specifically, status messages that were only generated on speed control from an Arduino. 
It needs to change to command messages.

src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
supertick: yup

src/main/java/org/myrobotlab/service/InMoovHand.java
supertick: Peer lifecycle is relatively new.. I was willing to give it flexibility.
We could certainly remove it, and the more strict life-cycle definition would make me happy.

But, I wanted to keep a card up my sleeve if I had to play it.
The killing of the jvm is a static load and exit on Leap - its just "poor code design"
If we have the correct natives, and reference to the java.library.path - all is happy.
But, personally, I'd prefer looking at a log file and seeing it error vs. someone just describing, "MRL just doesnt start"

src/main/java/org/myrobotlab/service/InMoovHead.java
supertick: that's all been cleaned up...
(Servo) cast has been there forever, and would explode if a different ServoControl was used.

PR #512, opened 2019-06-28
Agent auto update

supertick:


src/main/resources/MANIFEST.MF
```diff
@@ -0,0 +1,31 @@
+Manifest-Version: 1.0
```
kwatters: Should this file be checked in?  I figured it would be generated.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1144,145 +1128,98 @@ static public boolean loadAll() {
...
-    System.out.println("-logToConsole                              # redirects logging to console");
-    System.out.println("-logLevel <DEBUG | INFO | WARNING | ERROR> # log level");
-    System.out.println("-service <name1 Type1 name2 Type2 ...>     # create and start list of services, e.g. -service gui SwingGui");
-    // System.out.println("example:");
-    // String helpString = "java
-    // -Djava.library.path=./libraries/native/x86.32.windows
-    // org.myrobotlab.service.Runtime -service webgui WebGui gui SwingGui
-    // -logLevel INFO -logToConsole";
-    // System.out.println(helpString);
+    new CommandLine(new CmdOptions()).usage(System.out);
```
kwatters: nice.

src/main/java/org/myrobotlab/service/Arduino.java
```diff
@@ -443,31 +448,69 @@ public void attachMotorControl(MotorControl motor) throws Exception {
...
 
     servo.attach(this);
   }
 
+  /**
+   * 
+   * @param dm
+   */
+  public void reattach(DeviceMapping dm) {
+    Attachable attachable = dm.getDevice();
```
kwatters: This looks hard coded to only support reattaching servos and hobby servos?  or is this some sort of override because both Servo and HobbyServo reference the same MrlServo device.

It seems like any MrlDevice should be able to be re-attached.. not just Servo/HobbyServo

src/main/resources/MANIFEST.MF
supertick: checked in one is only for ide debugging - runtime is always generated

src/main/java/org/myrobotlab/service/Runtime.java
supertick: yes a lovely piece of software - picocli

src/main/java/org/myrobotlab/service/Arduino.java
supertick: agreed - although the details on reattach might depend on device. I've created a ticket for this #514

src/main/resources/MANIFEST.MF
kwatters: just update the email addr below and we can merge this request.  we have tickets to track the items of concern.

PR #520, opened 2019-07-03
Worke

supertick:


src/main/java/org/myrobotlab/service/Agent.java
```diff
@@ -1220,8 +1254,8 @@ public static void main(String[] args) {
         agent.shutdown();
       }
 
-      if (options.manifest) {
-        Map<String, String> manifest = Runtime.getManifest();
+      if ("".equals(options.version)) {
```
kwatters: seems like a StringUtils.isEmpty(options.version) would be better here,  I wonder could options.version be null?

.gitignore
```diff
@@ -104,3 +104,4 @@ Thumbs.db
 /backup-routes.py
 /worke.py
 /branches/
+/shared/
```
kwatters: what's this folder used for?

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -175,10 +175,11 @@
   // FIXME SecurityProvider
   protected AuthorizationProvider authProvider = null;
 
-  private Status lastError = null;
-  private Status lastStatus = null;
-  long lastStatusTs = 0;
-  long statusBroadcastLimitMs = 1000;
+  protected Status lastError = null;
```
kwatters: this is complex enough that we might want to pull it out into a small little ServiceStatus class, for the "last*" variables that you have here.

src/main/java/org/myrobotlab/service/Agent.java
```diff
@@ -505,10 +509,10 @@ public synchronized void restart(String id) throws IOException, URISyntaxExcepti
     log.info("restarting process {}", id);
     kill(id); // FIXME - kill should include prepare to shutdown ...
     sleep(2000);
-    spawn(id);
+    spawnId(id);
   }
 
-  private void spawn(String id) {
+  public void spawnId(String id) {
```
kwatters: Why is this method public, and if we do want it to be public, we really need some javadoc on it.

src/main/java/org/myrobotlab/service/Agent.java
```diff
@@ -519,6 +523,18 @@ private void spawn(String id) {
       log.error("spawn({}) threw ", id, e);
     }
   }
+  
+  public Process spawn() throws IOException, URISyntaxException, InterruptedException {
```
kwatters: add javadoc for public methods.

src/main/java/org/myrobotlab/service/Agent.java
```diff
@@ -904,8 +932,14 @@ public Process spawn(CmdOptions inOptions) throws IOException, URISyntaxExceptio
     Platform platform = Platform.getLocalInstance();
     String exeName = platform.isWindows() ? "javaw" : "java";
     pd.javaExe = String.format("%s%sbin%s%s", System.getProperty("java.home"), fs, fs, exeName);
-
-    pd.jvm = new String[] { "-Djava.library.path=libraries/native", "-Djna.library.path=libraries/native", "-Dfile.encoding=UTF-8" };
+    
+    String jvmArgs = "-Djava.library.path=libraries/native -Djna.library.path=libraries/native -Dfile.encoding=UTF-8";
```
kwatters: it'd be nice to be able to override j(ava|ni).library.path  variables...  (as a future enhancement.)  seems like you should be able to override/append to the library paths (sometimes that's useful.)

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1684,7 +1694,7 @@ static public void stopCli() {
   // the current concept is ok - but it does not work ..
   // make it work if necessary prefix everything by -agent-<...>
   // FIXME - implement --help -h !!! - handle THROW !
```
kwatters: seems like this FIXME is done now.. clean up the comment.  (also, what's the Command annotation? is that an example usage syntax?)

.gitignore
supertick: a common place to share data that you want saved over updates ... i don't want to check my keys in again :P

src/main/java/org/myrobotlab/service/Agent.java
supertick: yes, it can be null, but not when the flag was supplied by the user and  the arity=0..1 is in its definition @Option(names = { "-v", "--version" }, arity = "0..1"

.gitignore
supertick: taking this back out ... makes sense as a user I'd want my keys in the data directory

src/main/java/org/myrobotlab/service/Runtime.java
supertick: ok removing ...
Command annotation is part of picocli its whats printed out first on help

PR #522, opened 2019-07-07
Worke

supertick:
fix with -a and -f conflict


pom.xml
```diff
@@ -6,7 +6,7 @@
 		mvn -DskipTests -Dbuild.number=71 -DGitBranch=develop help:system
 		
 		# fast build
-		mvn -DskipTests -Dbuild.number=71 -DGitBranch=develop package -o
```
kwatters: Seems like we should have some sort of default for build.number=SNAPSHOT or something like that.. so we know it's not an official release/build version?

src/main/java/org/myrobotlab/service/Agent.java
```diff
@@ -98,16 +98,18 @@
   boolean autoCheckForUpdate = false;
 
   Set<String> possibleVersions = new TreeSet<String>();
+  
+  final static String REMOTE_BUILDS_URL_HOME = "http://build.myrobotlab.org:8080/job/myrobotlab-multibranch/";
```
kwatters: we should make this a default value, and allow users to specify a system property to override this url...  (maybe?)

PR #523, opened 2019-07-08
Worke

supertick:
ability for mrl to recompile itself and restart


pom.xml
```diff
@@ -113,6 +113,15 @@
     <artifactId>commons-cli</artifactId>
     <version>1.4</version>
   </dependency>
+   <!-- Core Library -->
```
kwatters: I don't see this dependency referenced from any service.  nor do I see it in the pom template.  It needs to be in at least one of those locations o/w a re-gen of the pom and this will disappear.

PR #533, opened 2019-07-19
acks back in with only 20ms wait

supertick:


supertick:
Your right calamity.  For some reason I thought returning was cancelling.
if (ackEnabled){
  if (waiting) { return } ...

This means changing generator code, which is fine - but needs more time
Its on my short list anyway, because, everything that can be generated should be re-genned when for final check-in

supertick:
Nice Catch !

supertick:
the boardinfo heartbeats need to be implemented in the virtual arduino too

PR #534, opened 2019-07-21
fix the onDisconnect that got broken and called after 1 second

Calamity01:


src/main/resources/resource/Arduino/MrlComm/MrlComm.cpp
```diff
@@ -545,6 +544,7 @@ unsigned int MrlComm::getCustomMsg() {
  }
 
  void MrlComm::onDisconnect() {
+  msg->publishError("disconnected"); // mostly for debug purpose
```
kwatters: I don't think this will work.. if the serial port is disconnected.. you can't publish the error back.
i worry that this would break.. causing the following code to never be  executed.

src/main/resources/resource/Arduino/MrlComm/Msg.cpp
```diff
@@ -806,3 +806,7 @@ void Msg::begin(HardwareSerial& hardwareSerial){
 byte Msg::getMethod(){
 	return ioCmd[0];
 }
+
+long Msg::getLastHeartbeat(){
```
kwatters: I don't think we should have this method here.  you should never ask for the last heartbeat.. the heartbeat should just be published.

src/main/resources/resource/Arduino/MrlComm/MrlComm.cpp
Calamity01: the onDisconnect was always called even if it was not disconnected. So I put that line to see if it's called at an inapropriate time.
The message will be put in the serial buffer just as many device will output to the serial buffer. I don't think it would break anything. But that line can be removed with no problem. it's just debugging thing

src/main/resources/resource/Arduino/MrlComm/Msg.cpp
Calamity01: there was 2 lastHeartbeatUpdate in the code. one in mrlComm.cpp and one in Msg.cpp. The one in Msg.cpp is getting updated, but not the one in mrlComm.

since the onDisconnected method is in mrlComm.cpp, it need to have access to the value that is been updated

Calamity01:
with the getBoardInfo now driven by arduino, this fix won't work

PR #535, opened 2019-07-26
fixed unit tests

supertick:


src/main/java/org/myrobotlab/codec/Api.java
```diff
@@ -142,21 +158,214 @@ static public Message uriToMsg(String uri) {
...
+   * @return
+   */
+  static public String getApiKey(AtmosphereResource r) {
+    String requestUri = r.getRequest().getRequestURI();
+    return getApiKey(requestUri);
+  }
 
-    int pos = requestUri.indexOf(Api.PARAMETER_API);
+  static public String getApiKey(String uri) {
+    int pos = uri.indexOf(Api.PARAMETER_API);
```
kwatters: this seems a little fragile... ok for now..  we might want to consider a URL class to handle this parsing for us.

src/main/java/org/myrobotlab/codec/Api.java
```diff
@@ -142,21 +158,214 @@ static public Message uriToMsg(String uri) {
...
+        }
+      } else {
+        // API_TYPE_MESSAGES
+        // DEPRECATE - FOR LOGGING ONLY REMOVE
+        log.warn("<< {}", CodecJson.encode(msg)); // FIXME if logTraffic
+        codec.encode(out, msg);
+      }
+
+    } catch (Exception e2) {
+      log.error("respond threw", e);
```
kwatters: we should be logging "e2" here not "e"

src/main/java/org/myrobotlab/codec/Api.java
```diff
@@ -142,21 +158,214 @@ static public Message uriToMsg(String uri) {
...
+    return process(webgui, apiKey, r.getRequest().getRequestURI(), r.getResponse().getOutputStream(), data);
+    
+    } catch(Exception e) {
+      handleError(webgui, apiKey, r, e);
+      return null;
+    }
+  }
+  
+  // MAX COMPLEXITY WITH SIMPLEST DATA TYPES
+  protected Object process(MessageSender sender, String apiKey, String uri, OutputStream out, String json) throws Exception {
```
kwatters: This method doesn't do anything, only returns null..  is it needed?

src/main/java/org/myrobotlab/codec/ApiMessages.java
```diff
@@ -153,16 +107,18 @@ public Object process(MessageSender sender, OutputStream out, Message requestUri
...
 
     } else {
+      // FIXME - this is terrible ! .. changing to a different api ??? wtf ???
       // First GET /api/messages - has data == null !
       // use different api to process GET ?
       // return hello ?
       // FALLBACK
-      ApiFactory api = ApiFactory.getInstance();
-      String newUri = requestUri.uri.replace("/messages", "/service");
+      apiKey = "service";
```
kwatters: The assignment of apiKey here seems like a side effect ,  not sure I understand the concern or detail of what's going on,  so I guess it's good to have the FIXME / TODO in the code for cleaning this up.

src/main/java/org/myrobotlab/io/StreamGobbler.java
```diff
@@ -0,0 +1,29 @@
...
+
+  public void run() {
+    try {
+      InputStreamReader isr = new InputStreamReader(is);
+      BufferedReader br = new BufferedReader(isr);
+      String line = null;
+      while ((line = br.readLine()) != null)
+        System.out.println(type + ">" + line);
+    } catch (IOException ioe) {
+      ioe.printStackTrace();
```
kwatters: should have a proper logger, not stdout.

src/main/java/org/myrobotlab/serial/Port.java
```diff
@@ -147,8 +147,10 @@ public void run() {
         }
       }
       log.info("{} no longer listening - last byte {} ", portName, newByte);
-    } catch (Exception e) {
-      log.error("port reading thread threw", e);
+    } catch (InterruptedException e) {
+      log.info("port {} interrupted - stopping listener", portName);
```
kwatters: not logging the exception stack trace here.  it should.

src/main/java/org/myrobotlab/service/TestCatcher.java
```diff
@@ -249,7 +241,7 @@ public void checkMsg(long timeout, String method, Object... checkParms) throws I
    * "would have" been created to make this direct callback
    * 
    */
-  public void addData(String method, Object... parms) {
+  public void addDatax(String method, Object... parms) {
```
kwatters: If this method isn't used anymore, it should be removed.

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -266,15 +266,15 @@ static public ServiceType getMetaData() {
 
   transient Map<String, Map<String, Panel>> desktops;
 
-  transient ApiFactory api = null;
+  transient ApiFactory apix = null;
```
kwatters: If this is no longer used, the reference here should be removed.

src/main/java/org/myrobotlab/serial/Port.java
supertick: its a specific exception from an interrupt - it was intentional, not an error - the stack trace is not needed

src/main/java/org/myrobotlab/service/TestCatcher.java
supertick: aye .. agreed - found a unified way to check for appropriate messages.
I've removed it locally.

src/main/java/org/myrobotlab/service/WebGui.java
supertick: aye .. already removed locally - will be in next pr

src/main/java/org/myrobotlab/io/StreamGobbler.java
supertick: Usually, i'd agree - but this was potentially part of Client.java which is supposed to be the very minimal java app capable of interfacing with mrl.  Picocli and wasync are the "only" dependencies - and its a console app - so I've been on the fence to including logging vs stdin/stdout.   At the current time locally Client doesnt use this file - and I'm not sure if it will stay.

src/main/java/org/myrobotlab/codec/Api.java
supertick: removed locally - will be in next pr

src/main/java/org/myrobotlab/codec/ApiMessages.java
supertick: Ya.  There is more cleanup needed in Api - currently working on this

src/main/java/org/myrobotlab/codec/Api.java
supertick: Maybe, log the exception of trying to report the exception ?   Sounds pretty chatty, but okay, added e2 locally.

src/main/java/org/myrobotlab/codec/Api.java
supertick: URL class (and URI) handle parsing of the full url, but nothing I know of in the jvm will handle "parts" of the path.  Mrl apis are very structured in that the api is determined by the path-part after the constant "/api/"
e.g. "/api/"{api-type}

PR #538, opened 2019-08-06
Worke

supertick:


pom.xml
```diff
@@ -302,6 +302,15 @@
   </dependency>
 <!-- Elasticsearch end -->
 
+<!-- Esp8266_01 begin -->
```
kwatters: this shouldn't be here anymore.

supertick: Agreed .. its gone !

PR #553, opened 2019-09-25
Inmoov hobbyservo

supertick:
current pull request - fine to reject but it will only GROW BIGGER !!!!
(don't be afraid !)


src/main/java/org/myrobotlab/service/Sweety.java
```diff
@@ -880,7 +880,7 @@ public void startService() {
     adaFruit16cRight = (Adafruit16CServoDriver) Runtime.start("I2cServoControlRight", "Adafruit16CServoDriver");
     chatBot = (ProgramAB) Runtime.start("chatBot", "ProgramAB");
     htmlFilter = (HtmlFilter) Runtime.start("htmlFilter", "HtmlFilter");
-    mouth = (MarySpeech) Runtime.start("mouth", "MarySpeech");
+    mouth = (AbstractSpeechSynthesis) Runtime.start("mouth", "MarySpeech");
```
kwatters: this should be the interface, not the abstract class...

supertick: agreed

supertick: fixed in next commit

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
```diff
@@ -256,21 +259,31 @@ synchronized public void install(String location, String[] serviceTypes) {
...
       // mark all libraries as installed
 
-      List<?> err = report.getAllProblemMessages();
+      List<?> err= report.getAllProblemMessages();
 
+      boolean error = false;
       if (err.size() > 0) {
         for (int i = 0; i < err.size(); ++i) {
           String errStr = err.get(i).toString();
+          if (!errStr.startsWith("WARN:  symlinkmass")){
```
kwatters: This seems really specific..  are there other sorts of warning messages that we might miss?

supertick: This is in response to this - https://issues.apache.org/jira/projects/IVY/issues/IVY-1609?filter=allissues
Which was improper errors coming back - We check errors, but this is not an error (shouldn't even be a warning if not invoked).  I'll comment it out, since there were a couple serious blockers to move to ant ivy 2.5.x

.gitignore
```diff
@@ -77,3 +77,5 @@
 /backup-routes.py
 /worke.py
 /myrobotlab*.jar
+/libbulletjme.so
```
kwatters: these shouldn't be added to the git ignore..

src/main/java/Sample.java
```diff
@@ -273,7 +273,7 @@ public Sample(String brokerUrl, String clientId, boolean cleanSession, boolean q
     } catch (MqttException e) {
       e.printStackTrace();
       log("Unable to set up client: " + e.toString());
-      System.exit(1);
+      return;
```
kwatters: maybe this file should be deleted entirely?

src/main/java/org/myrobotlab/arduino/VirtualMsg.java
```diff
@@ -1570,6 +1570,7 @@ public boolean readMsg() throws Exception {
           if (newByte > 64) {
             // TODO - send error back
             byteCount = 0;
+            System.out.println("TO BIG");
```
kwatters: we should avoid system.out.printlns....  and use log.debug/info ...  (System.out.println is good for debugging locally)

src/main/java/org/myrobotlab/framework/Platform.java
```diff
@@ -410,4 +406,9 @@ public static void main(String[] args) {
       // log.info("Exception: ", e);
     }
   }
+
+  public boolean getVmVersion() {
+    // TODO Auto-generated method stub
+    return false;
```
kwatters: What is this method for?  seems like it should return a string version.. or maybe an integer  number for a version..   remove this method if it's not used.

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -805,13 +803,58 @@ public String getDataInstanceDir() {
...
+    byte[] content = null;
+    try {
+      content = Files.readAllBytes(Paths.get(filename));
+    } catch (IOException e) {
+      error(e);
+    }
+    return content;
+  }
+
+  public String getResourceAsString(String resourceName) {
```
kwatters: this method should be explicit about the encoding that the resource will be decoded as.  ( utf-8? )  or the system default?  I worry about internationalization issues with this method.  we should force utf-8  (unless for some reason we don't want to do that?)

src/main/java/org/myrobotlab/lang/ServoLang.java
```diff
@@ -27,10 +27,18 @@ public String toPython(Servo s) {
     // save the servo map
     sb.append(name + ".map(" + s.getMin() + "," + s.getMax() + "," + s.getMinOutput() + "," + s.getMaxOutput() + ")\n");
     // if there's a controller reattach it at rest
+    // FIXME - there is the initial position vs rest - they potentially are very different
```
kwatters: re-attaching a controller should always specify the last known position from java-land as the initial attach position/angle...  only if a servo service is created from scratch, the initial position can be defaulted to it's rest position.

src/main/java/org/myrobotlab/net/Http.java
```diff
@@ -117,5 +124,17 @@ public static void get(String theUrl, String outFile) throws IOException {
...
       out.close();
   }
+  
+  public static void getFile(String url) throws IOException {
+    getFile(url, null);
+  }
+
+  public static void getFile(String url, String outFile) throws IOException {
+    if (outFile == null) {
+      outFile = url.substring(url.lastIndexOf("/") + 1);
```
kwatters: this url parsing is not actually correct..  it should be the last index of /  before the first index of ? or a #    (if a question mark or hash tag are in the url.)   Probably best to actually create a URL object in java.. to parse the url as a proper url and get the path...

src/main/java/org/myrobotlab/service/AudioFile.java
```diff
@@ -222,20 +221,14 @@ public void playResource(String filename) {
...
 
   public void playResource(String filename, Boolean isBlocking) {
-    log.warn("Audio File playResource not implemented yet.");
-    // FIXME AudioData needs to have an InputStream !!!
-    // cheesy - should play resource from the classpath
-    playFile(filename, isBlocking);
-    // TODO: what/who uses this? should we use the class loader to
-    // playFile(filename, isBlocking, true);
-
+    playFile(getResourceDir() + File.separator + filename, isBlocking);
```
kwatters: this changes the semantics of how this method behaves.. please add a javadoc comment to say that this will play a file relative to the resource directory .

src/main/java/org/myrobotlab/service/AudioFile.java
```diff
@@ -275,10 +268,12 @@ public AudioData waitFor(String filename, Object waitForMe) {
   }
 
   public void stop() {
+    AudioProcessor ap = processors.get(currentTrack);
     // dump the current song
-    processors.get(currentTrack).isPlaying = false; // FIXME -- necessary ????
+   
     // pause the next one if queued
-    processors.get(currentTrack).pause(false);
+    ap.pause(false); // FIXME me shouldn't it be true ?
```
kwatters: why do we pass a boolean into the pause method at all?   it should probably be refactored to simply be  ap.pause()  ...

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
```diff
@@ -1431,6 +1444,9 @@ public void onPointCloud(PointCloud pc) {
   public void onRegistered(Service service) throws Exception {
     // new service - see if we can virtualize it
     log.info("{}.onRegistered({})", getName(), service);
+    if (service.getName().contentEquals("i01.head.jaw")) {
```
kwatters: This is some debug logging / test code that should probably be cleaned up/commented out

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
```diff
@@ -2189,6 +2205,36 @@ public String toJson(Node node) {
...
+  public void onJointAngles(Map<String, Double> angleMap) {
+    for (String name : angleMap.keySet()) {
+      ServiceInterface si = Runtime.getService(name);
+      if (si instanceof Servo) {
+        ((Servo) si).moveTo(angleMap.get(name));
+      }
+    }
+  }
+
+  public void setDefaultServoSpeed(Double speed) {
```
kwatters: I'm surprised this method is here.. do we need this? is it some sort of helper function that might be better suited on some other service?

src/main/java/org/myrobotlab/service/Polly.java
```diff
@@ -123,7 +123,7 @@ private AmazonPolly getPolly() {
     String secret = getKey(AMAZON_POLLY_USER_SECRET);
 
     if (key == null || secret == null) {
-      log.error("this service requires 2 keys - you can set them once with Security.setKeys(AMAZON_POLLY_USER_KEY, AMAZON_POLLY_USER_SECRET)");
+      error("this service requires 2 keys");
```
kwatters: technically speaking this service requires a key and a secret..

.gitignore
supertick: they should be checked in ?  I doubt it ...

src/main/java/org/myrobotlab/service/RasPi.java
```diff
@@ -91,6 +91,7 @@ static public ServiceType getMetaData() {
     meta.addCategory("i2c", "control");
     meta.setSponsor("Mats");
     meta.addDependency("com.pi4j", "pi4j-core", "1.2-SNAPSHOT");
```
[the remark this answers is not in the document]
supertick: there are 3 versions 1.2 1.3 & 1.4 which all have snapshots and no release version the last I checked

src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -283,19 +280,26 @@ public Python(String n) {
       ServiceType st = sdt.get(i);
       // FIXME - cache in "data" dir Or perhaps it should be pulled into
       // resource directory during build time and packaged with jar
-      String url = String.format("https://raw.githubusercontent.com/MyRobotLab/pyrobotlab/%s/service/%s.py", p.getBranch(), st.getSimpleName());
+      String url = String.format("%s/%s.py", st.getSimpleName(), st.getSimpleName());
```
kwatters: I don't think this is a url anymore.. probably should rename the local variable to be filename ...

src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -370,7 +374,7 @@ synchronized public void createPythonInterpreter() {
     }
     log.info("Python System Path: {}", sys.path);
 
-    String selfReferenceScript = "from org.myrobotlab.service import Runtime\n" + "from org.myrobotlab.service import Python\n"
+    String selfReferenceScript = "from org.myrobotlab.framework import Platform\n" + "from org.myrobotlab.service import Runtime\n" + "from org.myrobotlab.service import Python\n"
```
kwatters: how about just do   "from org.myrobotlab.service import *"    ?

src/main/java/Sample.java
supertick: agreed ... dropped there chasing squirrels most likely ... deleted

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1878,6 +1874,14 @@ public Runtime(String n) {
     log.info("java.vm.name [{}]", System.getProperty("java.vm.name"));
     log.info("java.vm.vendor [{}]", System.getProperty("java.vm.vendor"));
     log.info("java.specification.version [{}]", System.getProperty("java.specification.version"));
+    
+    
+    String vmVersion = System.getProperty("java.specification.version");
+    vmVersion = "11";
```
kwatters: This is just a warning , right?   I use java11 with MRL ...  java8 is pretty much dead.   So I guess, I don't like this message here... we should be able to run under java8 or java11..  (and we do..)  consider removing this.

src/main/java/org/myrobotlab/arduino/VirtualMsg.java
supertick: Agreed .. this was temporary ... I was trying to isolate the root cause of the "no ack" error..
System.out - was a quick way to focus the logging to change it to error level ... filtering out all msgs except errors and this bit .. but I have removed it ..

src/main/java/org/myrobotlab/service/Security.java
```diff
@@ -417,6 +307,22 @@ public Boolean allowExportByType(String type, Boolean access) {
     return allowExportByType.put(CodecUtils.type(type), access);
   }
 
+  // default group permissions - for new user/group
+  // anonymous
+  // authenticated
+
+  private String byteArrayToHexString(byte[] b) {
```
kwatters: maybe consider using apache commons codec here instead?  
Hex.encodeHexString( bytes )

src/main/java/org/myrobotlab/service/Security.java
```diff
@@ -429,119 +335,145 @@ public void createDefaultGroups() {
...
+    SecretKeySpec sks = new SecretKeySpec(key, Security.AES);
+    return sks;
+  }
+
+  public String getStoreFileName() {
+    return storeFileName;
+  }
+
+  private byte[] hexStringToByteArray(String s) {
+    byte[] b = new byte[s.length() / 2];
```
kwatters: same.. consider apache commons codec here instaed...

src/main/java/org/myrobotlab/service/Serial.java
```diff
@@ -399,14 +399,15 @@ public void open(String inPortName, int rate, int dataBits, int stopBits, int pa
...
-    if (inPortName.toLowerCase().startsWith("tcp://")) {
+    if (inPortName != null && inPortName.toLowerCase().startsWith("tcp://")) {
       try {
         info("connecting tcp");
         connectTcp(inPortName);
         return;
       } catch (Exception e) {
         // not a big fan of re-throwing exceptions,
         // but I'll make an exception here
+        error("connect ctp threw");
```
kwatters: there's a typo in this error / log message  "ctp" should be "tcp"

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -656,6 +674,8 @@ public void handle(AtmosphereResource r) {
         Message msg = CodecUtils.fromJson(bodyData, Message.class);
         msg.setProperty("uuid", uuid);
 
+        System.out.println(String.format(">>> %s", msg));
```
kwatters: comment out or use a proper log message

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -979,6 +996,22 @@ public String redirect(String url) {
...
+    WebGui _self = this;
+
+    // done so a thread "from" webgui can restart itself
+    // similar thread within stop
+    // From a web request you cannot block on a request to stop/start self
+    new Thread() {
+      public void run() {
+        try {
+          while (nettosphere != null) {
+            Thread.sleep(500);
```
kwatters: is there a way to avoid using Thread.sleep(..)  here? if not, no big deal.

src/main/java/org/myrobotlab/framework/Platform.java
supertick: yup ... I agree .. I started it, primarily because of the whole java 11 causing issues with missing dependencies and the split up of the jvm and the lawyers field day ... :P
But I apparently got distracted from the implementation - some of this info exists in the manifest anyway - I've updated my short list to take care of this ...

src/main/java/org/myrobotlab/framework/Service.java
supertick: sure .. fixed

src/test/java/org/myrobotlab/integration/InMoovScriptTest.java
```diff
@@ -2,116 +2,113 @@
...
   }
 
-  // Test the inmoov minimal arm script.
-  @Test
-  public void testInMoovMinimalArm() throws IOException {
-    if (printMethods)
-      System.out.println(String.format("Running %s.%s", getSimpleName(), getName()));
-    if (!isHeadless()) {
-      Runtime.start("gui", "SwingGui");
+  public static void addScript(String filename, String root) throws IOException {
```
kwatters: I don't like relying on external resources in unit tests.  What is the intention of this?  to test what the current latest on develop is?  that means the mrl build will fail if there's a new error introduced into the scripts in the pyrobotlab repo... I don't think that's right.

src/main/java/org/myrobotlab/lang/ServoLang.java
supertick: I think we agree ... servos are dumb .. if InMoov want's to remember last run of a process and where all the servo positions are .. great .. let it add that functionality ... but servos are dumb
I'm good with that and will ignore the FIXME .. I'm working on Arduino + Servo now (and always WebGui) .. and will wait until I move to InMoov to look into the details of last know position

src/test/java/org/myrobotlab/service/ServiceScriptTest.java
```diff
@@ -62,45 +56,57 @@ public void tearDown() throws Exception {
...
         "PickToLight", "PythonProxy", "Sprinkler", "_TemplateProxy", "SwingGui" };
     Set<String> blacklist = new HashSet<String>(Arrays.asList(b));
 
-    pythonScripts = getPyRobotLabServiceScripts(branch);
+    pythonScripts = getPyRobotLabServiceScripts();
     log.info("testing {} scripts", pythonScripts.size());
 
     for (String name : pythonScripts.keySet()) {
+      
+      name = "AudioFile";
```
kwatters: Um... this is wrong...  why are we always overriding and forcing "AudioFile" to be tested here.

src/main/java/org/myrobotlab/net/Http.java
supertick: people don't supply urls - they supply strings .. so there must be something which accepts a string even if it does get turned into a url .. I'm fine with the URL parameter being the function of "maximum complexity" which all other overloads just do type conversions to finally get to that method with the URL ...   I'd not necessarily agree with your ? or # statement however because a file can be the "result" of a query to some cgi or angular function - I think the thing I needed at the time was the outFile

supertick: Added to my short list "Http.getFile use URL/URI parsing" .. but low priority at the moment

src/main/java/org/myrobotlab/service/AudioFile.java
supertick: Sure, I added a javadoc, but I think you are confused.  The original playFile is unchanged, the reason I **_added_** this file was to make the service script which has been broken for months if not years work.  playFile exists as it has forever .. playResource I actually implemented correctly, with the correct getResourceDir path ... previously is just logged this !

log.warn("Audio File playResource not implemented yet.");

:D

src/main/java/org/myrobotlab/service/AudioFile.java
supertick: well you need a resume() then ..
6 to 1/2 dozen ...

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
supertick: yeah ... pointcloud resolving to jmonkey polygon is way in the future .. 
I need to focus on low level (Arduino & Servo) then move on to InMoov (which will be big) and then JMonkey ... I agree - but consider this low priority and still needing debugging .. meaning anyone wanting to work on it directly can and also clean up the debugging
I'd be glad to do it .. but not there yet - and all this is certainly better that the previous JMonkey App which was completely hardcoded to InMoov (while now it dynamically loads any model you dump into the assets directory - "huge improvement")

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
supertick: JMonkey has a concept of a virtual servo driven by servo controls ... so it needs a concept of what the "virtual" speed is - not sure .. don't remember the details .. not in that area yet ... :)

src/main/java/org/myrobotlab/service/Polly.java
supertick: heh ... sure .. Security service has a different concept of keys and values vs amazons concept of key and secret ...  
Anyway both are protected as "keys" in the Security service but I agree - its in the context of Polly so ... (changed it with correct verbiage)

src/main/java/org/myrobotlab/service/Python.java
supertick: sure.. agree .. changed ..

src/main/java/org/myrobotlab/service/Python.java
supertick: I likey ! :)
Ahahahah ... crap .. Ya great idea - but it doesn't work ... raise a Jython issue perhaps ;)

src/main/java/org/myrobotlab/service/Runtime.java
supertick: well .. all the build, and all my test machines are currently running java8 
But ya .. java 11 is the future -  I've had at least 3 people send me noworkys (one being Nick_S) and when I told them to uninstall java11 and install java 8 - they are worky ... 
So until we have more worky java11 it should stay .. and for longer term - we should have a tested jvm version, and if mrl see's that its not on that version it should minimally warn

src/main/java/org/myrobotlab/service/Security.java
supertick: its in Security which is core ... same response applies

src/main/java/org/myrobotlab/service/Serial.java
supertick: ahahah ... 
my work ctp is common acronym "common telematics platform" ... muscle memory ..
fixed

src/main/java/org/myrobotlab/service/WebGui.java
supertick: ya .. was using it as high priority logging with log set to error level .. fixed in all instances inside webgui

src/main/java/org/myrobotlab/service/WebGui.java
supertick: The de-allocation of nettosphere is not in a good place..
I got a couple of prs accepted into nettosphere - but I think there is still some race condition problems

src/test/java/org/myrobotlab/integration/InMoovScriptTest.java
supertick: All service scripts have moved to the myrobotlab repo NO LONGER the pyrobotlab
(its a good thing ...)
all that crap of trying to keep on the same branch and get resources across repos is done... however the Test have not all been updated..
Its lots of work and I'm not done with it ...

src/test/java/org/myrobotlab/service/ServiceScriptTest.java
supertick: heh ... 
Question should be .. "Why are we always ignoring or disabling service scripts"  
Heh .. ya that's right all service script testing has been ignored or disabled for a very long time .. I'm in the process of activating it all back up ... but I can't fix all of the bad scripts at once ..
This is diagnostics ... and you can be assured that the build will fail when I pull out the sole service I have been testing, and let it test all it should be - but I can't fix it all at once ...
I'll pull it out and let builds fail until we need a build ..

FIXED - builds will now fail and build nodes will now kill themselves until action is taken ... need more elves !

supertick:
Really appreciate your review and comments kwatters as always !

supertick:
FYI - I don't really care if its merged, I know I have a shittonne of work to do .. but its pretty unsquirreled focused stuff .. all refactoring .. all bug fixing ... all inmoov cleanup (when i get to inmoov)

src/main/java/org/myrobotlab/service/AudioFile.java
kwatters: resume makes sense to me ... seems intuitive.

PR #554, opened 2019-10-02
Api refactor

supertick:


build.xml
```diff
@@ -0,0 +1,1075 @@
+<!--
```
kwatters: Why is there a build.xml file here?  I assume this was a mistake to checkin.

src/main/java/org/myrobotlab/framework/MethodCache.java
```diff
@@ -113,7 +115,7 @@ public static MethodCache getInstance() {
       if (instance == null) {
         instance = new MethodCache();
         instance.excludeMethods.add("main");
-        instance.excludeMethods.add("getMetaData");
+        instance.excludeMethods.add("getMetaData"); // why ?
```
kwatters: not sure why.. don't understand why we wouldn't want folks to be able to invoke getMetaData ...  seems like we can remove this exclude

build.xml
supertick: yup ... removed

src/main/java/org/myrobotlab/framework/MethodCache.java
supertick: agreed ... removed from exclusion

src/main/resources/resource/WebGui/.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
kwatters: is this intended to be checked in?

src/test/java/org/myrobotlab/client/InProcessCliTest.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+  
+  public String getResponse() {
+    String ret =  new String(bos.toByteArray());
+    log.info("cd => {}", ret);
+    // clear();
+    return ret;
+  }
+  
+  @Test
```
kwatters: you should removed empty test methods.

src/test/java/org/myrobotlab/codec/ApiCliTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.myrobotlab.logging.LoggerFactory;
+import org.myrobotlab.logging.LoggingFactory;
+import org.myrobotlab.test.AbstractTest;
+import org.slf4j.Logger;
+
+public class ApiCliTest extends AbstractTest {
+  
+  public final static Logger log = LoggerFactory.getLogger(ApiCliTest.class);
+
+  @BeforeClass
```
kwatters: remove the empty methods.

src/main/java/org/myrobotlab/client/InProcessCli.java
```diff
@@ -85,15 +308,12 @@ public void stop() {
...
-      // Logger logger = (Logger)
-      // LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
-      // logger.setLevel(Level.INFO);
 
       // Logger.getRootLogger().setLevel(Level.INFO);
 
       Runtime.getInstance();
 
-      InProcessCli client = new InProcessCli(System.in, System.out);
+      InProcessCli client = new InProcessCli(NameGenerator.getName(), "test", System.in, System.out);
```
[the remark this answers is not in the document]
supertick: Not interested at the moment in outputting to stderr

src/main/java/org/myrobotlab/client/InProcessCli.java
```diff
@@ -77,6 +115,191 @@ public void run() {
...
+      "/{serviceName}/" - list methods of service
+      "/{serviceName}/{method}" - invoke method
+      "/{serviceName}/{method}/" - list parameters of method
+      "/{serviceName}/{method}/p0/p1/p2" - invoke method with parameters
+      "/{serviceName}/{method}/p0/p1/p2/"
+       * 
+       * </pre>
+       */
+
+      String[] parts = data.split("/");
```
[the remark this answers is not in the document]
supertick: Yes, but if this is to be solved then there should be a good understanding of what the cli language "is"
Currently I can write this :
runtime/getUptime
it has been a problem specifying file locations .. 
e.g. 
python/execFile//root/test/test1.py

perhaps we can keep both concepts with something like this 
python/execFile/"root/test/test1.py"
but implementing it is not high on my priority

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1393,77 +1414,186 @@ static public boolean saveAll() {
...
+        if ("runtime".equals(serviceName) && "getHelloResponse".equals(msg.method)) {
+          params[0] = uuid;
+        }
+        ret = method.invoke(si, params);
+
+        // propagate return data to subscribers
+        si.out(msg.method, ret);
+
+        // sender is important - this "might" be right ;)
+        String sender = String.format("%s@%s", si.getName(), getId());
```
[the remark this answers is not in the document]
supertick: YES ! ... exactly
...
fixed

src/main/resources/resource/WebGui/.vscode/launch.json
supertick: yes-ish  .... it specifies how debugging could proceed in vscode - something a newbie would not want to figure out

src/test/java/org/myrobotlab/codec/ApiCliTest.java
supertick: done

src/main/java/org/myrobotlab/client/InProcessCli.java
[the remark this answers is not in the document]
kwatters: quotes would be fine...  i guess some sort of more robust parsing here is a "nice to have" for the time being.

src/main/java/org/myrobotlab/client/InProcessCli.java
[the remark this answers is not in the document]
kwatters: i'm mostly concerned about the scenario where stderr output buffer gets written and filled and causes strange behaviors in services (like we saw with dl4j & opencv)  perhaps this is no longer attached to an actual spawned process and as a result, we don't have to worry about stderr now?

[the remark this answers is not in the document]
supertick: builder.redirectErrorStream(true)  is still there ... so stderr goes to stdout and stdout is consumed by a StreamGobbler.  So, not much has changed there.  The thing that changed was we do not open  process.getOutputStream()  (remote process's stdin from the Agent). 

We had challenges with the Agent's handles of remote process streams.
InProcessCli has a different purpose - it does not handle a remote stream, it simply reads from it own process stdin and writes to stdout (it currently does not write to stderr and I don't see any reason to at the moment).

**--interactive-mode** is only used when you want to run Runtime without the Agent **AND** want cli control through stdin (so basically the old way, but without the Agent). This can be useful during debugging, but --interactive-mode is never sent by the Agent, so the remote process's stdin is never open during regular runtime.  This solves the most recent unyielding problem of opencv 1.5.1 being stuck.

All remote control now is done through websockets.
The Agent starts a listening webgui on port 8887 (without starting the browser), then spawns a new instance of MRL with a unique ID (new NameGenerator thingy).  By default the MRL instance will try to connect to port 8887, when/if it does, the Agent accepts the connection and then will have control of the instance.

This offers considerable advantages over previous control, and is similar in design to Apache Storm's Supervisor / Worker architecture.

https://www.tutorialspoint.com/apache_storm/apache_storm_cluster_architecture.htm

Agent's by default will only listen on localhost:8887 but a node is a node is a node.  With this new addressing scheme and all the refactoring work I did on gateways and routing it will be possible to connect everything in a distributed mesh, and still be able to manage from a cental endpoint ...

PR #556, opened 2019-10-23
windows: check os bitness & compare with jvm bitness

LunDev:
This will attempt to recognize the bitness of the OS during startup and find mismatches between OS & JVM.
Successfully tested on 64bit Windows 7/8/8.1/10 and with 64bit JVM 8/9/10/11 & 32bit JVM 8/9 (see below for detailed information). (For JVM 9 an early version had to be used since 32bit Java 9 was apparently never intended to be a thing.)
winXP isn't really needed. On the one hand it's unsupported since the 8th of April 2014, on the other hand 64bit winXP wasn't really a thing - and you can't mismatch bitness on 32bit systems.
For Linux, it is a similar story. I've tried for about an hour to get a working 32bit JVM on a 64bit system and gave up. If it's that hard to mismatch bitness, I'm not worried. (Results from 64bit Linux are below.)
About macOS ... well, I've no idea.

64bit Windows 7/8/8.1/10, 64bit JVM 8/9/10/11

Test Results
===
These are shortened excerpts taken from standard logging. Everything is level INFO unless noted otherwise.

64bit Windows 10, 64bit JVM 8
---
```
PROCESSOR_ARCHITECTURE=AMD64
PROCESSOR_ARCHITEW6432 not defined
os.name [Windows 10] getOS [windows]
os.arch [amd64] getArch [x86]
os.version [10.0]
java.vm.name [Java HotSpot(TM) 64-Bit Server VM]
java.vm.vendor [Oracle Corporation]
java.specification.version [1.8]
guessed os bitness [64]
```

64bit Windows 10, 64bit JVM 11
---
```
PROCESSOR_ARCHITECTURE=AMD64
PROCESSOR_ARCHITEW6432 not defined
os.name [Windows 10] getOS [windows]
os.arch [amd64] getArch [x86]
os.version [10.0]
java.vm.name [Java HotSpot(TM) 64-Bit Server VM]
java.vm.vendor [Oracle Corporation]
java.specification.version [11]
guessed os bitness [64]
```
(Other results with 64bit Windows versions and 64bit JVMs were excluded due to being boring.)

64bit Windows 10, 32bit JVM 8
---
**Please note the last line - yay!** - which is logging level WARN btw.
```
PROCESSOR_ARCHITECTURE=x86
PROCESSOR_ARCHITEW6432=AMD64
os.name [Windows 10] getOS [windows]
os.arch [x86] getArch [x86]
os.version [10.0]
java.vm.name [Java HotSpot(TM) Client VM]
java.vm.vendor [Oracle Corporation]
java.specification.version [1.8]
guessed os bitness [64]
detected possible bitness mismatch between os & jvm
```

64bit Linux, 64bit JVM 8
---
This is just for comparison.
```
PROCESSOR_ARCHITECTURE not defined
PROCESSOR_ARCHITEW6432 not defined
os.name [Linux] getOS [linux]
os.arch [amd64] getArch [x86]
os.version [4.9.0-7-amd64]
java.vm.name [OpenJDK 64-Bit Server VM]
java.vm.vendor [Oracle Corporation]
java.specification.version [1.8]
```

### DIFF
--- a/src/main/java/org/myrobotlab/framework/Platform.java
+++ b/src/main/java/org/myrobotlab/framework/Platform.java
@@ -49,7 +49,8 @@ public class Platform implements Serializable {
   // non-changing values
   String os;
   String arch;
-  int bitness;
+  int osBitness;
+  int jvmBitness;
   String lang = "java";
   String vmName;
   String vmVersion;
@@ -90,7 +91,7 @@ public class Platform implements Serializable {
 
       platform.startTime = new Date();
 
-      // os
+      // === OS ===
       platform.os = System.getProperty("os.name").toLowerCase();
       if (platform.os.indexOf("win") >= 0) {
         platform.os = OS_WINDOWS;
@@ -103,17 +104,9 @@ public class Platform implements Serializable {
       platform.vmName = System.getProperty("java.vm.name");
       platform.vmVersion = System.getProperty("java.specification.version");
 
-      // bitness
-      String model = System.getProperty("sun.arch.data.model");
-      if ("64".equals(model)) {
-        platform.bitness = 64;
-      } else {
-        platform.bitness = 32;
-      }
-
-      // arch
+      // === ARCH ===
       String arch = System.getProperty("os.arch").toLowerCase();
-      if ("i386".equals(arch) || "i686".equals(arch) || "i586".equals(arch) || "amd64".equals(arch) || arch.startsWith("x86")) {
+      if ("i386".equals(arch) || "i486".equals(arch) || "i586".equals(arch) || "i686".equals(arch) || "amd64".equals(arch) || arch.startsWith("x86")) {
         platform.arch = "x86"; // don't care at the moment
       }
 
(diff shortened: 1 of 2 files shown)


supertick:
Ahoy MaVo !  ...
Awesome that you have decided to play again ;) 
Great idea, but I think we should encapsulate what you've done into something useful for all services in org.myrobotlab.framework.Platform.   

If were going through the trouble of attempting to standardize this Java crazy ... we should put it into an object that other services can depend on ... and the logging can use this object too..

LunDev:
"Play" - as in "not wanting to do the work I should do" ^^

Now, knowing what framework.Platform is, it definitely makes more sense to do it there.
I've splitted the existing attribute `bitness` into `osBitness` and `jvmBitness`, where `jvmBitness` basically does what `bitness` did before. All places that used `bitness` where updated to use `jvmBitness` (all logging, except for one: Agent.setEnv(...), while I'm pretty sure there shouldn't be any changes in existing behavior, I'm not certain. This pr is not about only adding a check anymore.)

If osBitness is now becoming a feature of Platform, it might be worth to look into adding this for linux (at least for the major distros if each one has their own way) and macos as well.

PR #563, opened 2019-11-03
Example scripts

kwatters:
i figure having something is better than nothing.. they all need more work, but that will be incremental and over time i guess.


src/main/resources/resource/Deeplearning4j/Deeplearning4j.py
```diff
@@ -7,3 +7,12 @@
...
 deeplearning4j = Runtime.start('deeplearning4j','Deeplearning4j')
+
+# load the VGG16 model from the zoo
+deeplearning4j.loadVGG16()
+
+# run an image file through the model and get the classifications / confidence
+classifications = deeplearning4j.classifyImageFileVGG16("myimage.jpg")
+
+# print them out... it's a dictionary/map
+print classifications
```
supertick: its 2.7 Jython but shouldn't we always print with parens since it supports it ?

src/main/resources/resource/OpenNi/OpenNi.py
```diff
@@ -6,4 +6,20 @@
 # very minimal script to start
 
 openni = Runtime.start("openni", "OpenNi")
+python = Runtime.getService("python")
+
```
supertick: Runtime.start('python', 'Python') is always safe to call .. if it exists it just gets it ... if it doesn't it creates it

src/main/resources/resource/OpenNi/OpenNi.py
```diff
@@ -6,4 +6,20 @@
...
+# define a method for your callback..  here it passed the full openni data object that contains the skeleton info.
+def onOpenNIData(data):
+  # python handy method for printing all members on an object.
+  print "Full data object"
+  print dir(data)
+  print "Skeleton object"
+  print dir(data.skeleton)  
+  # example of printout of the head x,y,z coordinates.
+  print str(data.skeleton.head.x) + " " + str(data.skeleton.head.y) + " " + str(data.skeleton.head.z)
+
```
supertick: print ( ... ) ?

supertick:
Cool ! ... Thanks 
I could have sworn that I had a template that created any missing files ... hmmm mebbe its in the Test service.  Ahh .. there it is - https://github.com/MyRobotLab/myrobotlab/blob/develop/src/main/java/org/myrobotlab/service/Test.java#L380-L388

 It be nice to have a standardized header that provided this info, additionally had commented global parts  e.g. how to make stuff virtual
# Platform.setVirtual(True)

This might make super-noobies more apt to begin tinkering ...

src/main/resources/resource/Deeplearning4j/Deeplearning4j.py
kwatters: old habits.  i'll clean it up.

src/main/resources/resource/OpenNi/OpenNi.py
kwatters: implicitly, the python service has already been started, this is more to serve as an example that you can call getService to get a service by it's name if it's already been started.

src/main/resources/resource/Deeplearning4j/Deeplearning4j.py
kwatters: updated.

PR #566, opened 2019-11-03
Borg i2C LCD display driver

moz4r:
image

Hi ! This is to drive things like this : https://www.amazon.fr/gp/product/B01N3B8JMN/ref=ppx_yo_dt_b_asin_title_o03_s00?ie=UTF8&psc=1

Pasted from https://github.com/Poduzov/PI4J-I2C-LCD

Will post soon about the robot !


src/main/java/org/myrobotlab/service/LcdDisplay.java
```diff
@@ -0,0 +1,232 @@
...
+
+  public boolean getBackLight() {
+    return backLight;
+  }
+
+  /*
+   * Display string on LCD, by line 
+   */
+  public void display_string(String string, int line) {
+    screenContent.put(line, string);
```
supertick: heh ... this is Python style .. remember were in Java-Land now ;) should be displayString(String string, int line) or perhaps since the type is string (and its obvious a string) then just display(String string, int line) ...

src/main/java/org/myrobotlab/service/LcdDisplay.java
```diff
@@ -0,0 +1,232 @@
...
+    lcd_write((byte) 0x03);
+    lcd_write((byte) 0x02);
+    lcd_write((byte) (LCD_FUNCTIONSET | LCD_2LINE | LCD_5x8DOTS | LCD_4BITMODE));
+    lcd_write((byte) (LCD_DISPLAYCONTROL | LCD_DISPLAYON));
+    lcd_write((byte) (LCD_CLEARDISPLAY));
+    lcd_write((byte) (LCD_ENTRYMODESET | LCD_ENTRYLEFT));
+  }
+
+  /*
+   * Send byte to PCF controller
```
supertick: All these comments are awesome ..
But they should be JavaDoc .. e.g.
if you do this -> /**
and hit return, eclipse auto-creates the rest of the javadoc - you just fill it in

src/main/java/org/myrobotlab/service/LcdDisplay.java
```diff
@@ -0,0 +1,232 @@
...
+   */
+  public void write_cmd(byte cmd) {
+    if (isReady()) {
```
supertick: should be writeCmd(byte cmd)

src/main/java/org/myrobotlab/service/LcdDisplay.java
```diff
@@ -0,0 +1,232 @@
...
+  public void clear() {
+    lcd_write((byte) LCD_CLEARDISPLAY);
+    lcd_write((byte) LCD_RETURNHOME);
+    screenContent.clear();
+    log.info("LCD content cleared");
+  }
+
+  @Override
+  public void preShutdown() {
+    if (isReady()) {
```
supertick: cool .. nice use of preShutdown

src/main/java/org/myrobotlab/service/LcdDisplay.java
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  private void lcd_write(byte cmd, byte mode) {
+    lcd_write_four_bits((byte) (mode | (cmd & 0xF0)));
+    lcd_write_four_bits((byte) (mode | ((cmd << 4) & 0xF0)));
+  }
+
+  // write a command to lcd
+  private void lcd_write(byte cmd) {
+    lcd_write(cmd, (byte) 0);
```
supertick: all above should be standard Java camel case method names lcdWrite lcdStrobe lcdWriteFourBits ... etc..

src/main/java/org/myrobotlab/service/LcdDisplay.java
```diff
@@ -0,0 +1,232 @@
...
+  public LcdDisplay(String reservedKey) {
+    super(reservedKey);
+    setReady(false);
+  }
+
+  public final static Logger log = LoggerFactory.getLogger(LcdDisplay.class);
+  private static final long serialVersionUID = 1L;
+
+  private Pcf8574 i2cControler;
+
```
supertick: this probably should be an org.myrobotlab.service.interfaces.I2CController not a Pcf8574 ?

supertick:
Mozr !  ... :D
Very cool I2C Lcd Display Driver ...
I'm excited to see Spot in action ... wow you been busy

I made a few suggestions.  Additionally, I think the service name is too general.
Its I2C only vs many other types of LcdDisplay
Also I'm curious how "standard" are the I2C Lcd Display commands are ?
Are they to a very specific model and manufacturer ?  If so - we might want to incorporate that into the name.

To let different controllers use your driver - your controller should be a I2CController not a Pcf8574.
Cool stuff Moz4r !  Glad to see you back

supertick:
Where's the webgui parts ? ;)

src/main/java/org/myrobotlab/service/LcdDisplay.java
moz4r: Yeh , so useful ! thanks for added this

src/main/java/org/myrobotlab/service/LcdDisplay.java
moz4r: ohh yeh 👍  tks

src/main/java/org/myrobotlab/service/LcdDisplay.java
moz4r: /**
   *  checked in
   * 
   */

src/main/java/org/myrobotlab/service/LcdDisplay.java
moz4r: Pcf8574 is not a I2CController but IO expander for i2c bus.
I will rename some stuff because I used "i2cControler" as var name..

moz4r:
Hi ! thank you for the review! I will polish
Maybe we can enhance support for more devices, like direct arduino pin attach without Pcf8574 but this is an other (long) story I think. 

I will rename to use for a specific model, maybe this will work for others.
Ohh yeh webGui is missing ! I was lazy hehe, will push it :)

About the robot I think I will need some help soon for the walk ! ( gyro / pid  ... exciting ! )
Some stuff will also be used later for my other big robot hehe

For now I will try to drive & test every hardware inside nixie

MatsMRL:
Great stuff. Perhaps it would be better to name the service based on the driver name, HD44780 since that driver is used to drive many different LCD displays.

moz4r:
Hi Mats !! Yeh you are right. Thanks for all ic2 stuffs ! it help a lot

src/main/java/org/myrobotlab/service/Hd44780I2c.java
```diff
@@ -0,0 +1,242 @@
...
+/**
+ * 
+ * HD44780 - Driver to play with lcd display panel
+ * Tested with HD44780 1602 panel, attached to pcf8574
+ * Driver pasted from Poduzov : https://github.com/Poduzov/PI4J-I2C-LCD
+ * 
+ * @author Moz4r
+ * 
+ */
+public class Hd44780I2c extends Service {
```
kwatters: Can we rename this class to remove the "i2c" from it.   Update the name to "HD44780"  seems a bit cleaner.

src/main/java/org/myrobotlab/service/Hd44780I2c.java
```diff
@@ -0,0 +1,242 @@
...
+
+  public Hd44780I2c(String reservedKey) {
+    super(reservedKey);
+    setReady(false);
+  }
+
+  public final static Logger log = LoggerFactory.getLogger(Hd44780I2c.class);
+  private static final long serialVersionUID = 1L;
+
+  private Pcf8574 pcf8574;
```
kwatters: This device should be able to be attached to any I2CController service.  Can we re-write this to use the Interface, rather than hard coding that it needs to be attached to a Pcf8574 ?

src/main/java/org/myrobotlab/service/Hd44780I2c.java
moz4r: yes, will do

src/main/java/org/myrobotlab/service/Hd44780I2c.java
moz4r: No, I tried, pcf8574 don't use I2CController interface , it is not a I2CController service but a I/O bus extender for i2c bus. 
I'm not sure if we can get it worky with other stuff inside mrl ... maybe later

MatsMRL: The PCF8574 is a i2c bus extender. It will give 8 extra I/O pins. The pcf8574 is built to be able to connect to any I2cController. So it can be used on the Arduino, RasPi or ESP8266_01 already. So that goal is already acheived.

supertick: Who's going to write the interface ? :)

kwatters: I was thinking that the display LcdDisplay is a standard i2c device.  it, thus should be an I2CControl and should be attachable to an I2CController.   We don't have to do it now, but I think this is a direction we should go in, for future i2c devices.

MatsMRL: Hi. 
There is two different devices involved here:

One is the LCD display driver board HD44780. It has no i2c interface. It can potentially also be connected directly to the pins of an Arduino or RasPi.
The service moz4r has built is a represenation of that board. The reason he is using the service Pcf8574 directly, and not using an interface, is because I never put the methods that you can use to communicate with it in a generic port extender interface. I think I should do that, because then moz4r could use that interface.

The Pcf8574 is an i2c device, and implements the I2CControl interface and can talk to any i2c controller using the I2CController interface. So I think we are good there.
It also implements PinArrayControl, and thats how I was expecting it to be used. That works fine and for example the Pcf8574Gui use the same Oscope component that is used in the ArduinoGui. But I will add a generic PortExtenderControl interface that moz4 can use, because that is the missing piece of that service. 

If LcdDisplay (HD44780) indeed was an i2c device, I would agree 100% with what Kwatters write. But in this case the pcf8574 is the i2c device and LcdDisplay (HD44780) is not.
The alternative would be to look at the combination of HD44780 and pcf8574 as a single unit. In that case I would also agree with Kwatters. Perhaps that is a better way to go. Then no new interface would be needed, but all i2c configuration ( I2CControl ) would need to be implemented in that service. 
And I think some generic LcdDisplayControl interface should be defined for the LCD methods.

supertick: Sounds good - would you like to write another ticket(s) for the missing work ?
All you have more experience than myself for these components ... I guess you still have a couple of decisions to go through but possible tickets (as I understand it would be)

- create generalized LcdDisplayControl interface
- create PortExtenderControl interface - integrate it with this service (or not?) and implement I2CControl in this service

supertick:
merging moz4r's new service - additional tickets/interfaces/etc are coming

PR #568, opened 2019-11-08
More cli fixes

supertick:
Huge overhaul Log service and its webgui - seems very worky
image


src/main/java/org/myrobotlab/service/Log.java
```diff
@@ -25,23 +25,18 @@
...
-import java.net.URI;
+import java.io.IOException;
 import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.List;
 
-import org.myrobotlab.framework.MRLListener;
-import org.myrobotlab.framework.Message;
+import javax.annotation.processing.FilerException;
```
kwatters: Can we choose a different exception class there?  javax.annotation isn't super java11 friendly.

src/main/java/org/myrobotlab/service/Log.java
```diff
@@ -283,42 +252,61 @@ public void setRootLogLevel(String level) {
...
+    super.startService();
+    startLogging();
+  }
+
+  @Override
+  public void stop() {
+  }
+
+  public String getLog() throws IOException {
+    return FileIO.toString("myrobotlab.log");
```
kwatters: seems like the log file name should be pulled out as a constant, or alternatively, a member variable so you could override it if you wanted.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -38,6 +38,8 @@
 import java.util.TreeSet;
 import java.util.UUID;
 
+import javax.annotation.processing.FilerException;
```
kwatters: This appears to be unused?

src/main/resources/MANIFEST.MF
```diff
@@ -11,7 +11,7 @@ GitCommitIdDescribe: da9bc99
 GitBuildUserName: 
 GitRemoteOriginUrl: https://github.com/MyRobotLab/myrobotlab.git
 GitTags: 
-GitBranch: develop
+GitBranch: more_cli_fixes
```
kwatters: If this file is generated with the build, I would think it shouldn't be checked in?

src/main/java/org/myrobotlab/service/Log.java
supertick: ZOIKS !  - that was there for  testing purposes only - should have removed it  - will remove it

src/main/java/org/myrobotlab/service/Log.java
supertick: done

src/main/java/org/myrobotlab/service/Log.java
supertick: removed

src/main/java/org/myrobotlab/service/Runtime.java
supertick: removed

src/main/resources/MANIFEST.MF
supertick: its a "basic" dev file - since locally running through the IDE you don't necessarily need nor want the generate file stuff to run .. there should be "something" checked in .. I tweak it - but rarely

PR #575, opened 2019-11-19
Distributed Mrl

supertick:
image

The purpose of this commit is to have multiple mrl instances capable of connecting and providing a messaging system that allow successful intra-communication.  Above you can see 2 mrl instances connected together.  The only requirement is to have one instance running the WebGui service.  This becomes a "hub" to a possible star network topology.  And on this network any mrl instance can send or receive messages from any other mrl instance.

Sending messages locally still works with a "short" name ...  e.g. send('clock','startClock') - but if you want to send a message to a remote system you must fully address it with the '@{instance}'
e.g. send('clock@raspi', 'startClock')


supertick:
image

The purpose of this commit is to have multiple mrl instances capable of connecting and providing a messaging system that allow successful intra-communication.  Above you can see 2 mrl instances connected together.  The only requirement is to have one instance running the WebGui service.  This becomes a "hub" to a possible star network topology.  And on this network any mrl instance can send or receive messages from any other mrl instance.

Sending messages locally still works with a "short" name ...  e.g. send('clock','startClock') - but if you want to send a message to a remote system you must fully address it with the '@{instance}'
e.g. send('clock@raspi', 'startClock')

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -2068,6 +2090,22 @@ public void subscribe(String topicName, String topicMethod) {
...
+  
+  public void subscribeToRuntime(String method) {
+    subscribe(Runtime.getInstance().getName(), method, getName(), CodecUtils.getCallbackTopicName(method));
+  }
+  
+  public void unsubscribeTo(String service, String method) {
+    unsubscribe(service, method, getName(), CodecUtils.getCallbackTopicName(method));
+  }
+  
+  public void unsubscribeToRuntime(String method) {
```
kwatters: just a nit pick.. unsubscribeFromRuntime might be a slightly better method name.

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -2505,4 +2543,8 @@ public String getFullName() {
   public void copyResource(String src, String dest) throws IOException {
     FileIO.copy(getResourceDir() + File.separator + src, dest);
   }
+
+  public void setId(String id) {
```
kwatters: just curious, why should you be able to set the id of a service?  seems like this is something that should be immutable?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -2932,27 +3105,49 @@ public void removeConnection(String uuid) {
...
+   * list connections - current connection names to this mrl runtime
+   * 
+   * @return
+   */
+  public Map<String, List<Map<String, String>>> lc() {
+    return getConnectionHeaders();
   }
 
-  public String getConnectionName(String uuid) {
+  public Map<String, String> getConnectionHeader(String uuid) {
```
kwatters: would it ever be possible to have multiple headers with the same key? or are we forbidding that?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3074,7 +3274,11 @@ public Object sendBlockingRemote(Message msg, Integer timeout) throws IOExceptio
       // --> remote
       // FIXME implement !
       Endpoint endpoint = (Endpoint) conn.get("c-endpoint");
-      endpoint.socket.fire(CodecUtils.toJson(msg));
+      String json = CodecUtils.toJson(msg);
+      if (json.length() > 65536) {
```
kwatters: why do we have a 64k limit here?  is that some protocol limitation?

src/main/java/org/myrobotlab/framework/Service.java
supertick: yeah ... very much agree ...
It was residue from trying to get serialization all worky 
Serialization needs to set the id (from remote systems) - now it does without this

src/main/java/org/myrobotlab/service/Runtime.java
supertick: absolutely not - a connection is identified by a uuid

src/main/java/org/myrobotlab/service/Runtime.java
supertick: oop ... nice spot check ..
Not sure if its a websocket RFC spec or a safety measure employed by the Atmosphere implementation. 

By default using atmosphere, if the message is over 65K it gets dumped.
I think overall its a good idea to try to keep the message payloads small, so its a bit of a reminder.  With configuration I made it have more capacity, but I still want to slice down a couple bloat messages.

I'll rip out this debugging stuff

supertick: I was writing out the large messages and inspecting them (too that out) - I left the warnings in - there could be more refactoring done with the bloated messages

PR #576, opened 2019-12-04
Get servos worky

supertick:


src/main/java/org/myrobotlab/service/Joystick.java
```diff
@@ -293,9 +293,10 @@ public void addInputListener(Service service) {
     service.subscribe(this.getName(), "publishJoystickInput");
   }
 
-  public void addListener(String serviceName, String id) {
```
kwatters: this is a breaking api change.. seems like we should just deligate this method to attach and mark it as depricated.

supertick: originally adding this function is breaking the normal pub/sub framework of mrl.
I consider this a breaking bug fix, because its overriding frameworks addListener(topicName, topicCallback).
I discovered this working on the webgui, the gui was broken because it could create a valid subscription.
This fixes it.

PR #583, opened 2019-12-15
Caching time encoder positions

supertick:
Servos using time encoders how have checkpoints to save their 'current' position
this can substantially reduce crazy coffee moves on startup

### DIFF
--- a/src/main/java/org/myrobotlab/framework/Service.java
+++ b/src/main/java/org/myrobotlab/framework/Service.java
@@ -148,7 +148,7 @@ public abstract class Service implements Runnable, Serializable, ServiceInterfac
   /**
    * for promoting portability and good pathing
    */
-  transient protected String fs = File.separator;
+  transient protected static String fs = File.separator;
 
   /**
    * for promoting portability and good pathing
@@ -782,14 +782,18 @@ public abstract class Service implements Runnable, Serializable, ServiceInterfac
   public String getHomeDir() {
     return System.getProperty("user.home");
   }
-
-  public String getDataDir() {
-    String dataDir = Runtime.getOptions().dataDir + fs + getClass().getSimpleName();
+  
+  static public String getDataDir(String typeName) {
+    String dataDir = Runtime.getOptions().dataDir + fs + typeName;
     File f = new File(dataDir);
     if (!f.exists()) {
       f.mkdirs();
     }
-    return Runtime.getOptions().dataDir + fs + getClass().getSimpleName();
+    return Runtime.getOptions().dataDir + fs + typeName;
+  }
+
+  public String getDataDir() {
+    return getDataDir(getClass().getSimpleName());
   }
 
   public String getDataInstanceDir() {
--- a/src/main/java/org/myrobotlab/sensor/TimeEncoder.java
+++ b/src/main/java/org/myrobotlab/sensor/TimeEncoder.java
@@ -360,7 +360,6 @@ public class TimeEncoder implements Runnable, EncoderControl {
     return estimatedPos;
   }
 
-  @Override
   public void setPos(Double pos) {
     beginPos = targetPos = estimatedPos = pos;
     positions.setPosition(name, estimatedPos);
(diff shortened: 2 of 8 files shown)


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -782,14 +782,18 @@ public String getRootDataDir() {
   public String getHomeDir() {
     return System.getProperty("user.home");
   }
-
-  public String getDataDir() {
-    String dataDir = Runtime.getOptions().dataDir + fs + getClass().getSimpleName();
+  
+  static public String getDataDir(String typeName) {
```
kwatters: Why does this method need a typeName passed into it now?  Implicitly, the service should already know what type it is.

src/main/java/org/myrobotlab/sensor/TimeEncoder.java
```diff
@@ -79,8 +85,93 @@
...
+    @Override
+    public void run() {
+
+      running = true;
+
+      while (running) {
+        try {
+          Thread.sleep(2000);
+          log.debug("saving {} positions of {} servos", filename, positions.size());
+          FileIO.toFile(filename, CodecUtils.toJson(positions).getBytes());
```
kwatters: This seems very heavy handed to me.  Why would we bother with this?  On startup , servos go to their rest position at half speed.  The only time we should save the positions/state of any service should be when we explicitly tell the service to save its state.

src/main/java/org/myrobotlab/service/LegacyServo.java
```diff
@@ -691,7 +691,7 @@ public void setInverted(boolean invert) {
   }
   
   @Override
-  public void unsetSpeed() {
+  public void disableSpeedControl() {
     log.info("disabling speed control");
     velocity = null;
```
kwatters: for consistency, if we're killing "velocity" this variable should be renamed to "speed".

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -238,6 +238,13 @@ public AbstractServo(String n, String id) {
     // create our default TimeEncoder
     if (encoder == null) {
       encoder = new TimeEncoder(this);
+      // if the encoder has a current value - we initialize the
+      // servo with that value
+      Double savedPos = encoder.getPos();
+      if (savedPos != null) {
+        log.info("found previous values for {} setting initial position to {}", getName(), savedPos );
```
kwatters: I think we're trying to do too much with the initial position of the servos.  I like  the deterministic behavior as stated by Gael in the shout box.  Saving and attempting to know the current state of a hobby servo after a power cycle really need to be a problem that is out of scope.  Servos do not provide positional feedback, so , i think we're adding complexity to solve a problem that fundamentally isn't solvable with a standard hobby servo.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -948,6 +955,10 @@ public Double getTargetPos() {
   @Override
   public void setPosition(Double pos) {
     currentPos = targetPos = pos;
+    if (encoder != null) {
+      if (encoder instanceof TimeEncoder )
+      ((TimeEncoder)encoder).setPos(pos);
+    }
```
supertick: sort of ugly - but probably "expected" behavior

kwatters: we can revisit this, but for now, i think we want to be pretty explicit about how certain features are enabled on different types of encoders.

src/main/java/org/myrobotlab/sensor/TimeEncoder.java
kwatters: a future enhancement should be something that skips writing the update if the servos haven't moved since the last time...

src/main/java/org/myrobotlab/sensor/TimeEncoder.java
```diff
@@ -79,8 +85,94 @@
 
   boolean enabled = true;
 
+  static class Positions implements Runnable {
+
+    static Positions instance = null;
+    Map<String, Double> positions = new HashMap<>();
```
kwatters: I'm a bit worried about threading here... perhaps we should be using a ConcurrentHashMap instead?

PR #586, opened 2020-01-20
Servo pub sub

supertick:
pub sub servos, mappers, and google search service


src/main/java/org/myrobotlab/service/Sphinx.java
```diff
@@ -1,700 +0,0 @@
-/**
```
kwatters: Please restore this file.  We shouldn't get rid of this class / service until we have another offline speech recognition alternative that is already borged in.

src/main/java/org/myrobotlab/service/WebkitSpeechRecognition.java
```diff
@@ -17,276 +12,23 @@
...
-  private boolean speaking = false;
-
-  private long lastAutoListenEvent = System.currentTimeMillis();
-  public boolean stripAccents = false;
-  private boolean lockOutAllGrammar = false;
-  private String lockPhrase = "";
-
-  public WebkitSpeechRecognition(String n, String id) {
-    super(n, id);
-    languagesList.put("en-US", "English - United States");
```
kwatters: how will you get the list of supported languages from webkit if this is removed from here?

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
```diff
@@ -1,35 +1,589 @@
...
+  protected boolean isSpeaking = false;
+
+  /**
+   * wait for 1 sec after my speaking has ended
+   */
+  protected long afterSpeakingPauseMs = 1000;
 
   public AbstractSpeechRecognizer(String n, String id) {
     super(n, id);
+    languages.put("en-US", "English - United States");
```
kwatters: This list of langauges is specific to the WebKitSpeechRecognizer class..  it should not be in the base abstract class.

src/main/java/org/myrobotlab/service/WebkitSpeechRecognition.java
supertick: This all will be resolved with LocaleProvider interface.
Fist thing was fixing up the large amount of confusion between language and locale.
Next address the large amount of confusion and problems from java.util.Locale supporting some flavor of iso-630.  So, make a Locale which accepts old and new locale "code" format and delivers new IETF standard (with hyphens).  ie - you can set a LocaleProvider with underscore hyphen without countries etc .. and it will try to figure out what you want, but it will "always" provide IETF standard.

A LocalProvider can supply its own set of supported Locales.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
supertick: this is changed locally - my local has SpeechRecognizer and SpeechSynthesis both extend LocaleProvider interface - and the respective abstracts contain "default" sets of Locales, if a service needs to implement some limited subset, it simply overrides the abstract and returns the set it actually supports in a "getLocales()" call

The nice part is we're now always dealing with the same format and same methods from the interface.

src/main/resources/resource/WebGui/app/service/js/WebkitSpeechRecognitionGui.js
```diff
@@ -1,239 +1,290 @@
...
-                if ((!$scope.recognizing && $scope.service.listening) || !clickedFromWebGui) {
-                    // TODO: start Recognition is actually a toggle.. not start.
-                    $scope.startRecognition();
-                }
-            }
-            ;
-        } else {
-            $log.info("State did not change.");
-        }
+    $scope.languages = {
```
kwatters: These likely need to move down to the base java class.. sorry.  if a browser isn't open.. you won't be able to query for what the langauges are and that isn't super.

src/main/resources/resource/WebGui/app/service/js/WebkitSpeechRecognitionGui.js
```diff
@@ -1,239 +1,290 @@
...
+                $scope.restartCnt += 1
+                recognizer.start()
+            }
+            break
+
+        case 'onerror':
+            $scope.errorText = event.error
+            let errorTs = new Date().getTime()
+            if ((errorTs - $scope.startTimestamp) < 100) {
+                $scope.errorText += ' - high error rate - check other tabs for an active webkit speech recognizer, and close it'
```
kwatters: This error message is basically useless to people.  it is very misleading.  the old error messages should be restored.

src/main/resources/resource/WebGui/app/service/js/WebkitSpeechRecognitionGui.js
```diff
@@ -1,239 +1,290 @@
...
+        if (!$scope.service.isListening) {
+            $scope.setState('start')
+        } else {
+            $scope.setState('stop')
         }
     }
-    ;if ($scope.wkavailable) {
-        // called when there's an error (handles a few cases)
-        $scope.recognition.onerror = function(event) {
-            if (event.error == 'no-speech') {
```
kwatters: please restore this if block to make sure the proper error messages and actions are exposed to end users.

src/main/java/org/myrobotlab/service/Sphinx.java
supertick: it will be restored in another commit - i have no idea what the workyness of it is ...

src/main/java/org/myrobotlab/service/WebkitSpeechRecognition.java
supertick: Not done with this - we both agree it originates in Java-Land, but the Locales need to work in IETF / Javascript land

src/main/resources/resource/WebGui/app/service/js/WebkitSpeechRecognitionGui.js
supertick: ya .. discussed, we agree, its not quite there - the Java Land Chinese Locales need some work

src/main/java/org/myrobotlab/service/interfaces/LocaleProvider.java
```diff
@@ -0,0 +1,36 @@
...
+public interface LocaleProvider {
+  
+  /**
+   * set a locale from a code string.
+   * The code string accepts underscore or hyphens
+   * en-US fr-FR or just "en" or "fr" is the country is not needed
+   * @param code
+   * @return
+   */
+  public String setLocale(String code);
```
kwatters: this should return void.

src/main/resources/resource/WebGui/app/service/js/WebkitSpeechRecognitionGui.js
supertick: This needs work with the error state handling, which is not done yet.
I think having the "last" opened recognizer work is nice - vs, scouring through your tabs looking for the other one.  Its either that or the both get turned off.  The speech recognizer is global in chrome - only one tab can own the resource ... that's just the way it is ...

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -946,4 +951,29 @@ public void attachTextPublisher(TextPublisher service) {
...
+  }
+
+  @Override
+  public Locale getLocale() {
+    return locale;
+  }
+
+  @Override
+  public Map<String, Locale> getLocales() {
+   return Locale.getMap("en-US", "fr-FR", "es-ES", "de-DE", "nl-NL", "ru-RU", "hi-IN","it-IT", "fi-FI","pt-PT");
```
kwatters: This isn't right... only by convention do the inmoov chat bots have locales assocated with them...  as far as program ab is concerned. it actually only supports english or japanese.  (based on the currently implemented tokenziers.

src/main/java/org/myrobotlab/service/WebkitSpeechSynthesis.java
```diff
@@ -0,0 +1,75 @@
...
+ * Free Tier The Amazon Polly free tier includes 5 million characters per month,
+ * for the first 12 months, starting from the first request for speech.
+ * 
+ * Polly Pricing Pay-as-you-go $4.00 per 1 million characters (when outside the
+ * free tier).
+ *
+ * @author GroG
+ *
+ */
+public class WebkitSpeechSynthesis extends AbstractSpeechSynthesis {
```
kwatters: This class is not named correctly.  There is no such thing as WebkitSpeechSynthesis.  Webkit is a speech recognizer.  I think perhaps you mean this to be Polly?  Or perhaps this class was added by accident?  Please remove/revert.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
```diff
@@ -18,16 +19,20 @@
...
+    public boolean isRecording = false;
     public boolean isFinal = false;
     public String text;
   }
 
   /**
    * generalized list of languages and their codes - if useful
    */
-  static protected Map<String, String> languages = new HashMap<>();
+  protected Map<String, Locale> locales = new HashMap<>();
```
kwatters: No Required if this implements the LocaleInterface.  I don't think you need/want this local variable here.  The interface should be enfoced.  if you want the locals. you should use the setters/getters from the interface..  not this local protected member.  this should be removed.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
```diff
@@ -112,82 +117,10 @@
 
   public AbstractSpeechRecognizer(String n, String id) {
```
kwatters: Please add the "abstract" key word here.  just because the class starts with the word "Abstract" doesn't mean it's abstract.  classes defined as abstract should be   "public abstract class"

src/main/resources/resource/WebGui/app/service/js/WebkitSpeechRecognitionGui.js
supertick: Sure, I'd like to report the proper error states to the user, but its more complex than just slapping this code back in.  There were lots of problems, and errors states it was not handling either - for example the test for appropriate webkit support wasn't being done, additionally - the user was not being notified if webkit or speechrecognition was not supported.   We need the ability to show "recording" vs "listening" - it becomes important for privacy and the ability to support a wake word.
Working on these parts currently

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -1029,6 +1059,23 @@ public Boolean setBlocking(Boolean b) {
   public boolean isMute() {
     return mute;
   }
+  
+
+  @Override
+  public String setLocale(String code) {
```
kwatters: These should be marked as abstract and let the subclasses should implement them.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -41,6 +41,17 @@
...
+  /**
+   * generalized list of languages and their codes - if useful
+   */
+  protected Map<String, Locale> locales = new HashMap<>();
+  
+
+  /**
+   * the current locale this service is set to e.g. en-US it-IT fr etc...
+   */
+  protected Locale locale;
```
kwatters: i don't think this should be here.. the subclasses should be implementing the interfaces for LocaleProvider.  the subclass should be able to determine if it needs a locale object locally or not.

src/main/resources/resource/WebGui/app/service/js/WebkitSpeechRecognitionGui.js
kwatters: Back on October this was working great.  At a minimum, it needs to detect 
1. when no microphone is found.
2. when the microphone is denied access.   In either of those scenarios it must not attempt to re-listen again.. because those are non-recoverable errors.  
current behavior ultimately will crash and lockup chromium on raspi how it's implemented, and never let the end user know what the underlying cause was.

supertick: yup - agree

src/main/java/org/myrobotlab/service/interfaces/LocaleProvider.java
supertick: ok, although it is satisfying in REST apis to have the value echoed back...
done.

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: I'll let you change it .. I've been unwinding all sorts of surprises I've never seen.  
It would be nice to have a strategy of how "the user" will be less burden by changing or aligning all Locales.

src/main/java/org/myrobotlab/service/WebkitSpeechSynthesis.java
supertick: No, its not incorrectly named.   It's "discovered".  In reading the documentation regarding webkitSpeechRecognizer, it turns out Chrome support SpeechSynthesis.  They are two independent functions so, they deserve to be separate.   I did some testing initially in WebkitSpeechRecognizer, but didn't want what I had learned to be there so I created the appropriate files.  This is on the same level as Polly, but no need of a key.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
supertick: interface is enforced, this is "set" in the constructor by the interfaces getLocales() which the derived class must implement.  Its a convenience property, because otherwise the webgui would need to subscribe to getLocales explicitly. In this case, all services sent to javascript contain their appropriate locales.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
supertick: No, it should simply set the locale .. unless for some very unlikely reason, it needs to be overridden.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
supertick: I think your getting confused - subclasses should simply create their own getLocales - ie give a list of locales they support.  The internals of setLocale and now that I think of it sending an error if setLocale is outside the set of getLocales.  It would be a shame and a mess to duplicate this code in each service.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
supertick: I guess your looking at the constructor .. the class is decorated with abstract

kwatters: public AbstractSpeechRecognizer(String n, String id) 
should read
public abstract AbstractSpeechRecognizer(String n, String id)

o/w it's not actaully abstract.

supertick: You are incorrect.
https://docs.oracle.com/javase/tutorial/java/IandI/abstract.html

_"An abstract class is a class that is declared abstract—it may or may not include abstract methods. Abstract classes cannot be instantiated, but they can be subclassed."_

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
kwatters: not that we really obey it, i'm just going with the java bean convention..  we can leave this as is.

src/main/java/org/myrobotlab/service/WebkitSpeechRecognition.java
supertick: image
Normalized all chaos to single set of Locales supplied by Java-land WebKitSpeechRecognizer.

src/main/resources/resource/WebGui/app/service/js/WebkitSpeechRecognitionGui.js
supertick: This has been fixed, tested and verified.

src/main/resources/resource/WebGui/app/service/js/WebkitSpeechRecognitionGui.js
supertick: All error information is propagated to UI, nothing is hidden.  Only "high rate" is appended when appropriate.   All errors stop and do not try to restart, with the exception of "no-speech" which is not an error that should be stopped on.  This change should prevent the tight endless loop which you noticed previously.

supertick: Tested, verified.

src/main/java/org/myrobotlab/service/interfaces/LocaleProvider.java
supertick: Changed, although not only in REST, but additionally this would prevent the value if something was interested in subscribing to it.

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: Leaving this for you to take care of.

PR #603, opened 2020-03-01
latest support - for unpinning dependency versions

supertick:


src/main/resources/resource/WebGui/app/service/js/InMoov2Gui.js
```diff
@@ -219,7 +219,7 @@ angular.module('mrlapp.service.InMoov2Gui', []).controller('InMoov2GuiCtrl', ['$
     
     calculatButtonPos()
 
-    $scope.setPanel('settings')
+    $scope.setPanel('InMoov')
```
kwatters: Should this be InMoov2 ?

supertick: Funny you should say that - since the beginning webgui has all its ui under /resource/WebGui/app .. i was just saying to Gael in the shoutbox - it would be really cool if a service bundled "all" of its parts .. including ui
It's never been done in mrl - but it should be

PR #612, opened 2020-03-05
Some missing icons

hairygael:
Not sure to understand why we had to make a new set called InMoov2


supertick:
Because I had to make new services for InMoov2, I was suppose to leave original InMoov alone.
That is the reason why.

PR #615, opened 2020-03-07
Webgui new default ui

supertick:


src/main/java/org/myrobotlab/service/Agent.java
```diff
@@ -1047,10 +1047,10 @@ public Process spawn(CmdOptions inOptions) throws IOException, URISyntaxExceptio
     }
 
     if (options.services.size() == 0) {
-      options.services.add("log");
-      options.services.add("Log");
-      options.services.add("gui");
-      options.services.add("SwingGui");
```
kwatters: does this mean swing gui doesn't start by default anymore?

kwatters:
i'm all for webgui starting by default.. however i am against disabling the swing gui starting by default.

kwatters:
additionally.. if this pull request is actually just to change the default gui.. it should only contain the 1 file.. not a whole bunch of stuff related to webkit speech synth.

supertick:
gah .. the swing gui lives on .... it just won't die !

PR #625, opened 2020-03-10
FIXME note

hairygael:


supertick:
helpful, although we have no methods and process to define type of controller

hairygael:
Moz4r had made a solution with the current ini config, we define which controller is used. This means the "head" can have more controllers than only one.
https://github.com/MyRobotLab/InMoov2/blob/master/resource/InMoov2/config/skeleton_head.config.default#L8

PR #630, opened 2020-03-11
checking pir

hairygael:


src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -1842,6 +1842,31 @@ public InMoov2Torso startTorso(String port) {
     return torso;
   }
 
+  public Pir startPir() {
+    return startPir(null);
```
kwatters: we need to choose a default "port" and "pirPin" value here so that it calls the method startPir(port, pirPin) as defined below.

hairygael: ok thanks i will add that

hairygael: and if I remove?
  public Pir startPir() {
    return startPir(null);
  }

kwatters: you can remove the whole startPir() method if you like.. sure..

PR #633, opened 2020-03-15
back to acking acking

supertick:


src/main/java/org/myrobotlab/arduino/ArduinoMsgGenerator.java
```diff
@@ -25,7 +25,7 @@
 
   public transient final static Logger log = LoggerFactory.getLogger(ArduinoMsgGenerator.class);
 
-  static final Integer MRLCOMM_VERSION = 63;
```
kwatters: There's no changes to the MRLComm code in this pull request.  We shouldn't be increasing the version number here unless there is a change in MrlComm.cpp (or other arduino code)

PR #634, opened 2020-03-15
logging in programab webgui

supertick:


src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -41,1927 +41,1951 @@
...
+	public static LinkedHashMap<String, String> lpVars = new LinkedHashMap<String, String>();
+
+	// FIXME - why
+	static boolean RobotCanMoveRandom = true;
+	private static final long serialVersionUID = 1L;
+
+	static String speechRecognizer = "WebkitSpeechRecognition";
+
+	/**
+	 * This static method returns all the details of the class without it having to
```
kwatters: Why are there white space changes in this class as part of this pull request?

src/main/java/org/myrobotlab/service/interfaces/LogPublisher.java
```diff
@@ -0,0 +1,30 @@
+package org.myrobotlab.service.interfaces;
```
kwatters: This file doesn't seem to be obeying 2 spaces no tabs. are you using the proper code formatter?

PR #635, opened 2020-03-15
very small cosmetic updates

supertick:


src/main/java/org/myrobotlab/audio/AudioProcessor.java
```diff
@@ -227,7 +228,11 @@ public AudioData play(AudioData data) {
     } catch (Exception e) {
       log.error(e.getMessage(), e);
       if (audioFile != null) {
-        audioFile.error("%s - %s", e.getMessage(), data.getFileName());
+    	if (e instanceof LineUnavailableException) {
+    		audioFile.error("line is not available");
```
kwatters: appears to be whitespace problems here also..

PR #642, opened 2020-03-21
adding ServoMixer

hairygael:
the purpose is to add the servo mixer into InMoov webgui


hairygael:
hello Grog, kwatters,
I am still not familiar with pulling requests.
This one doesn't get through and I am wondering what is the issue

PR #643, opened 2020-03-21
fix errors on Neopixel gui

hairygael:


src/main/resources/resource/WebGui/app/service/views/NeoPixelGui.html
```diff
@@ -65,13 +65,13 @@
   <div class="col-lg-12">
     <div class="input-group">
     	<span class="input-group-addon">Pixel #</span>
-    	<input type="text class="form-control" ng-model=eachpixel.address ng-disabled="true">
+    	<input type="text class="form-control ng-model="eachpixel.address" ng-disabled="true">
     	<span class="input-group-addon">Red</span>
```
supertick: this is not correct .. neither was the original ... double quotes need to surround each value .. e.g.

<input type="text" class="form-control" ng-model="eachpixel.address" ng-disabled="true">

within tags property and values  should always be of the form 
<tag   property="value"  ></tag>

hairygael: arfff, some dumb copy paste

PR #655, opened 2020-03-23
error when writting setPosition

hairygael:
error when writting setPosition which cannot be executed


src/main/java/org/myrobotlab/lang/ServoLang.java
```diff
@@ -14,7 +14,7 @@ public String toPython(Servo s) {
     // sb.append(name + ".detach()\n");
     sb.append("# sets initial position of servo before moving\n");
     sb.append("# in theory this is the position of the servo when this file was created\n");
-    sb.append(name + String.format(".setPosition(%.1f)\n", s.getPos()));
+    sb.append(name + String.format(".setPosition(%.1f)\n". s.getPos()));
     sb.append(name + ".setMinMax(" + s.getMin() + "," + s.getMax() + ")\n");
```
supertick: Hi Gael,
This will not fix the problem, this won't compile

supertick:
Ok Gael, I made a new pr that should fix the problem

PR #677, opened 2020-04-08
Webgui work

supertick:


src/main/java/org/myrobotlab/service/InMoov2Arm.java
```diff
@@ -135,10 +135,15 @@ public InMoov2Arm(String n, String id) throws Exception {
     shoulder.setPin(10);
     omoplate.setPin(11);
 
-    bicep.setMinMax(5.0, 90.0);
-    rotate.setMinMax(40.0, 180.0);
+    bicep.setMinMax(0.0, 180.0);
```
kwatters: These shouldn't have changed..  The bicep min/max should be 5-90 ... not 0 to180.  The change in behavior in setMinMax was fixed .. these changes should be reverted.

kwatters: setMinMax is deprecated on last_servo_fixes branch.  It's behavior is to set both the input and the output min/max values for the servos.  setMinMax(5.0,90.0)  is the same as, and should be replaced with the more explicit map(5,90,5,90)   .

there's been an issue with autoDisable for a while.  I'll try to work through getting that fixed on last_servo_fixes branch.. at that point, hopefully all of this stuff is straightened out. 

There were many years of bad variable names and unnecessary logic that have piled up in the Servo class.  It's been a significant effort to unroll it all and clean it up.  I think GroG & I are very close though.

hairygael: Hello Kevin,
The way it should be is :
 bicep.setMap(5.0, 90.0, 0.0, 180.0);
Because we need to set the input values as default, the output values need to be done by the user.
I was trying to fix that in the InMoov2Arm.java but unfortunately setMap was returning a error when compiling.
So because it was not compiling correctly I just set back as it was and added annotations.
When you say setMinMax is fixed and then you say it is deprecated, does it mean that setMinMax represent now the input or the output, or still both together?

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -55,6 +55,11 @@
   public final static Logger log = LoggerFactory.getLogger(AbstractServo.class);
 
   private static final long serialVersionUID = 1L;
+  
+  /**
```
supertick, line 59: How is this planned to be resolved ?

kwatters: Out .. delete this !!!!

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -237,6 +242,7 @@ public AbstractServo(String n, String id) {
       }
     }    
     currentOutputPos = mapper.calcOutput(targetPos);
+    autoDisable = autoDisableDefault;
```
kwatters, line 245: delete this.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -938,4 +944,8 @@ public void writeMicroseconds(int uS) {
     invoke("publishServoWriteMicroseconds", this, uS);
   }
 
+  public static void setAutoDisableDefault(boolean b) {
```
kwatters, line 947: definitely delete this.

src/main/resources/resource/Intro/Servo01_start.py
```diff
@@ -0,0 +1,49 @@
...
+# set speed
+servo01.setSpeed(100.0)
+
+# attach servo
+# print("attaching servo with pins to controller")
+# servo01.attach(arduino.getName(), servoPin01)
+
+# auto disable - this enables (starts pwm) before a movement
+# and disables (stops pwm) after a movement
+servo01.setAutoDisable(True)
```
kwatters, line 47: we should also show an example of  servo01.setIdleTimeout(1000)  to set the amount of time before the auto disable kicks in..

hairygael: Sure, although, this script is not to be seen by the user.
I made this script to be launched via click click during the Intro servo tutorial. Functions to be applied to a servo are mentionned at the end of the tutorial on the Intro webgui page.
setAutoDisable(True) still sets a default IdleTimeout ? correct?

src/main/java/org/myrobotlab/service/InMoov2Head.java
```diff
@@ -67,8 +67,10 @@ public InMoov2Head(String n, String id) {
     eyeX.setPin(22);
     eyeY.setPin(24);
     jaw.setPin(26);
-    //FIXME rollNeck must be connected to right controller on pin 12 not on pin 30
-    rollNeck.setPin(30);
```
kwatters, line 71: Where is rollNeck pin set if it's not being set here?

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
kwatters: remove this.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -55,6 +55,11 @@
   public final static Logger log = LoggerFactory.getLogger(AbstractServo.class);
 
   private static final long serialVersionUID = 1L;
+  
+  /**
+   * setting the default for all new servos
+   */
+  protected static boolean autoDisableDefault = false;
```
kwatters: remove this

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
kwatters: remove this method

src/main/resources/resource/Intro/Servo01_start.py
kwatters: default idleTimeout is 3.0 seconds.  yes. setAutoDisable(True) will use the current value of idleTimeout of the servo.

src/main/java/org/myrobotlab/service/InMoov2Head.java
hairygael: InMoov2Head.java was not written correctly because it was estimated that all servos would be on left controller, but it is not correct. Some head servos like rollNeck and eyeLids are on the right controller, because they share the same number pins. Two Arduino with two NervoBoards.

PR #683, opened 2020-04-18
checking neopixel1

hairygael:


supertick:
Still in process?

hairygael:
Yes that is a currently noworky.
When startNeopixel is called via InMoov2 gui, it just doesn't work.
Same problem with PIR and both UltraSonic sensors

supertick:
Added neopixel peer in InMoov2 with start functions - closing this pr

PR #690, opened 2020-04-29
refactored programab

supertick:


src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -129,13 +104,98 @@ public String toString() {
    */
   public ProgramAB(String n, String id) {
     super(n, id);
-    availableBots = getBots();
+    // botPaths = initBotPaths();
```
kwatters, line 107: I feel like this should be done in the startService.. not in the constructor...

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -343,7 +356,7 @@ public void removePredicate(String userName, String botName, String predicateNam
    */
   public void addToSet(String setName, String setValue) {
     // add to the set for the bot.
-    Bot bot = bots.get(getCurrentBotName().toLowerCase());
+    Bot bot = getBot(getCurrentBotName());
```
kwatters, line 359: You've marked getBot as depricated, but you're using it here.  perhaps it shouldn't be deprecated?

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -656,102 +577,71 @@ public void startSession(String path, String userName, String botName) throws IO
...
-      bot = new Bot(botName, path, locale);
-      // Hijack all the SRAIX requests and implement them as a synchronous call
-      // to
-      // a service to return a string response for programab...
-      bot.setSraixHandler(new MrlSraixHandler(this));
-      // put the bot into the bots map/cache referenced by it's lowercase
-      // botName.
-      bots.put(botName.toLowerCase(), bot);
+
+  @Deprecated /*
```
kwatters, line 592: you've marked this as deprecated, but I don't see a new version of startSession to use instead.  You should only mark something deprecated if there is a replacement way of doing it.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -798,28 +678,69 @@ public void writeAIML() {
...
+  /**
+   * Verifies and adds a new path to the search directories for bots
+   * 
+   * @param path
+   * @return
+   */
+  public String addBotPath(String path) {
+    // verify the path is valid
+    File botPath = new File(path);
+    File verifyAiml = new File(FileIO.gluePaths(path, "aiml"));
```
kwatters, line 704: I thought we were going to drop the requirement for having an aiml subdirectory...

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -934,8 +911,13 @@ static public ServiceType getMetaData() {
     // know they are different than the inmoov bots.
     // each bot should have their own name, it's confusing that the inmoov bots
     // are named en-US and so are the program ab bots.
-    meta.addDependency("program-ab", "program-ab-data", "1.1", "zip");
-    meta.addDependency("program-ab", "program-ab-kw", "0.0.8.4");
+
+    // meta.addDependency("program-ab", "program-ab-data", "1.2", "zip");
+    // meta.addDependency("program-ab", "program-ab-kw", "0.0.8.5");
+
+    meta.addDependency("program-ab", "program-ab-data", null, "zip");
```
kwatters, line 918: you're including program-ab-data here as a dependency, however it was removed from the pom.xml ...

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -934,8 +911,13 @@ static public ServiceType getMetaData() {
...
     // each bot should have their own name, it's confusing that the inmoov bots
     // are named en-US and so are the program ab bots.
-    meta.addDependency("program-ab", "program-ab-data", "1.1", "zip");
-    meta.addDependency("program-ab", "program-ab-kw", "0.0.8.4");
+
+    // meta.addDependency("program-ab", "program-ab-data", "1.2", "zip");
+    // meta.addDependency("program-ab", "program-ab-kw", "0.0.8.5");
+
+    meta.addDependency("program-ab", "program-ab-data", null, "zip");
+    meta.addDependency("program-ab", "program-ab-kw", "0.0.8.5");
```
kwatters, line 919: This should be 0.0.8.6

src/main/resources/resource/framework/pom.xml.template
```diff
@@ -307,9 +307,16 @@
                 <configuration>
                   <filesets>
                     <fileset>
-                      <directory>.myrobotlab</directory>
+                      <directory>data/.myrobotlab</directory>
                       <followSymlinks>false</followSymlinks>
 				    </fileset>
+					<fileset>
```
kwatters, line 313: please use spaces, not tabs.

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: replacement would be startSession(String userName, String botName) which already exists - I'll comment it in ..

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: I'd like one indicator besides failing when lazy-loading creating the bot, to give a user some warning that it won't work.  Since the new Bot(path) in programab-kw has hardcoded sub-dirs, a subdir of aiml seemed to be an "ok" test

pom.xml
```diff
@@ -1091,17 +1091,11 @@
...
 
 <!-- ProgramAB begin -->
-  <dependency>
-    <groupId>program-ab</groupId>
-    <artifactId>program-ab-data</artifactId>
-    <version>1.1</version>
-    <scope>provided</scope>
-    <type>zip</type>
-  </dependency>
+<!-- skipping program-ab program-ab-data program-ab-program-ab-data-null null version/latest -->
```
kwatters, line 1094: Where are the default bots coming from if this isn't here?

src/main/java/org/myrobotlab/service/ProgramAB.java
kwatters: Then this code should be removed and call the other method.  however, doing so, you will loose functionality of specifying the locale.  This is probably ok.  If this is the way we are going forward, this method should be cleaned out, leaving only a warning message and a call to the proper method...  it means that the locale parameter will be ignored.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -656,102 +577,71 @@ public void startSession(String path, String userName, String botName) throws IO
...
+   * @param botName
+   */
+  public void setCurrentSession(String userName, String botName) {
+    setCurrentUserName(userName);
+    setCurrentBotName(botName);
   }
 
   public void addCategory(Category c) {
-    Bot bot = bots.get(getCurrentBotName().toLowerCase());
+    Bot bot = getBot(getCurrentBotName());
```
kwatters, line 644: getBot is marked as deprecated, you shouldn't be using it if you are deprecating it.

src/main/java/org/myrobotlab/service/ProgramAB.java
kwatters: This isn't a warning..  this will prevent you from creating a bot that doesn't have any aiml yet.  I think we should remove this verification.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -995,13 +982,17 @@ public Locale getLocale() {
 
   @Override
   public Map<String, Locale> getLocales() {
-	  // FIXME should be based on bots found ???
-   return Locale.getLocaleMap("en-US", "fr-FR", "es-ES", "de-DE", "nl-NL", "ru-RU", "hi-IN","it-IT", "fi-FI","pt-PT");
+    // FIXME should be based on bots found ???
+    return Locale.getLocaleMap("en-US", "fr-FR", "es-ES", "de-DE", "nl-NL", "ru-RU", "hi-IN", "it-IT", "fi-FI", "pt-PT");
```
kwatters, line 986: This method returns non-sense..  the list of locales that this instance of ProgramAB supports would be the list of locales for any of the bots that are currently available..  not this list.   make it return null.. or something.. this list is just wrong.

src/main/resources/resource/framework/pom.xml.template
kwatters: This isn't / shouldn't be related to the program ab change.. is it ?

src/test/java/org/myrobotlab/service/SolrTest.java
```diff
@@ -1,11 +1,12 @@
 package org.myrobotlab.service;
 
-import static org.bytedeco.opencv.global.opencv_imgcodecs.cvLoadImage;
+import static org.bytedeco.opencv.helper.opencv_imgcodecs.cvLoadImage;
```
kwatters, line 3: Why is this import changing as a result of the PR ?  we didn't switch opencv versions ...

src/test/java/org/myrobotlab/service/SolrTest.java
```diff
@@ -114,7 +115,7 @@ public void testService() throws Exception {
     
     Assert.assertEquals(1, resp.getResults().getNumFound());
     
-    ProgramAB.Response programABResponse = new ProgramAB.Response("joe", "lloyd", "this is a test response.", null, new Date());
+    Response programABResponse = new Response("joe", "lloyd", "this is a test response.", null);
```
kwatters, line 118: What happened to the date object in the constructor here?  This shouldn't change.

src/main/resources/resource/framework/pom.xml.template
supertick: agreed

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: removed deprecated - keeping private

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: added comment /* use startSession(String userName, String botName) */

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: agreed - its nonsense ..
made it return the list of bot locales

src/test/java/org/myrobotlab/service/SolrTest.java
supertick: its an appropriate internal timestamp, as it should be

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: agreed ... moved

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: correct,
its included here with null - which will make ivy query for the latest during runtime..
its not included inside the maven, because I believe eclipse has a bug in dealing with it, while maven seems ok with parens notation (as saintandreas deps have always had)

src/test/java/org/myrobotlab/service/SolrTest.java
supertick: hmm.. setting back

pom.xml
supertick: runtime ivy gets the latest program-ab-data - same as inmoov2
we'll pin it on release

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: So empty dir as a start?, yet Bots actually do require convention - they have hardcoded sub-directories. 

If its only a empty directory and i can dump an aiml file named anything and get some kind of cause and effect, I'd consider taking the validation out.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -995,13 +986,32 @@ public Locale getLocale() {
 
   @Override
   public Map<String, Locale> getLocales() {
```
kwatters, line 988: Seems like this should return a set, not a map..  no need to change now.. it just seems odd..  or if it's a map, i'd expect some way to lookup a which bots have which locale..

src/test/java/org/myrobotlab/service/ProgramABTest.java
```diff
@@ -308,5 +312,50 @@ public void umlautTest() throws IOException {
...
     assertEquals("He's a character from Guitar Hero!", resp.msg);
   }
+  
+  @Test
+  public void newTests() throws IOException {
+    // minimal startup - create the service get a response
+    ProgramAB alice = (ProgramAB)Runtime.start("alice", "ProgramAB");
+    assertTrue(alice.getBots().size() > 0);
+    
+    Response response = alice.getResponse("Hello");
```
kwatters, line 322: Where is the start session call?  I'd expect that you need to tell it which bot you want to have a conversation with and additionally, which user is talking with the bot.  ProgramAB should not load a default bot unless you start a session with it.

src/test/java/org/myrobotlab/service/ProgramABTest.java
```diff
@@ -308,5 +312,50 @@ public void umlautTest() throws IOException {
     // @GroG says - "this is not working"
     assertEquals("He's a character from Guitar Hero!", resp.msg);
   }
+  
+  @Test
+  public void newTests() throws IOException {
```
kwatters, line 317: This test method should be renamed to indicate what it's actually testing.

src/test/java/org/myrobotlab/service/ProgramABTest.java
supertick: start session doesnt even start a bot ..
lazy loading prevents a bot from actually loading unless theres a request for a response

src/test/java/org/myrobotlab/service/ProgramABTest.java
supertick: updated

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: you can - botInfo has a mrl Locale

PR #691, opened 2020-05-02
Better acks2

kwatters:
not ready to merge yet.. just wanted to create the PR to start getting some more feedback.  This build passes locally for me.

### DIFF
--- a/src/main/java/org/myrobotlab/serial/PortStream.java
+++ b/src/main/java/org/myrobotlab/serial/PortStream.java
@@ -55,15 +55,18 @@ public class PortStream extends Port {
     // the "pure" Java (non-JNI/JNA) world...
     return new ArrayList<String>();
   }
-
-  @Override
-  public int read() throws IOException {
-    return in.read();
-  }
   
   public byte[] readBytes() {
     // TODO: implement this.. i don't actually want integers.. i want a byte array!
-    throw new NotImplementedException();
+    try {
+      Integer val = in.read();
+      byte[] data = new byte[1];
+      data[0] = val.byteValue();
+      return data;
+    } catch (IOException e) {
+      log.warn("Interrupted PortStream in readBytes.  Perhaps port was closed?", e);
+    }
+    return null;
   }
   
   // WORTHLESS INPUTSTREAM FUNCTION !! -- because if the size of the buffer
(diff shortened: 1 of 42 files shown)


src/main/java/org/myrobotlab/serial/PortStream.java
```diff
@@ -54,17 +54,30 @@ public OutputStream getOutputStream() {
...
+      int numRead = in.read(data);
+      if (numRead != size) {
+        // Uh oh.. let's just log this warning, we shouldn't see it.
+        log.warn("Port Stream Read possible error.  numRead {} does not equal size {}", numRead, size);
+        // return just the buffer of what we think was actually read.
+        return Arrays.copyOfRange(data, 0, numRead);
+      }
+      // Assume that all data was read properly and return the full buffer
+      return data;
+
```
supertick, line 76: This is much improved !

src/main/java/org/myrobotlab/service/Arduino.java
```diff
@@ -1438,83 +1431,19 @@ public void neoPixelWriteMatrix(NeoPixel neopixel, List<Integer> data) {
...
-      byteCount = 0;
-      Logging.logError(e);
-    }
-    return newByte;
+  public synchronized void onBytes(byte[] bytes) {
+    // log.info("On Bytes called in Arduino. {}", bytes);
+    // These bytes arrived from the serial port data, push them down into the msg parser.
+    // if a full message is detected, the publish(Function) method will be directly called on
+    // this arduino instance.
+    msg.onBytes(bytes);
```
supertick, line 1439: Nice refactor !

src/main/java/org/myrobotlab/service/Arduino.java
```diff
@@ -1651,17 +1576,18 @@ public BoardInfo publishBoardInfo(Integer version/* byte */,
...
-      }
-    }
+    // TODO: consider, can we really just re-sync when we see begin only.. ?  feels better/safer.
+//    if (boardInfo != null) {
+//      DeviceSummary[] ds = boardInfo.getDeviceSummary();
+//      if (deviceList.size() - 1 > ds.length) { /* -1 for self */
+//        log.info("Invoking Sync DeviceList: {} and DeviceSummary: {}", deviceList, ds);
+//        invoke("sync");
+//      }
+//    }
```
supertick, line 1586: does device sync still work ?

kwatters: sync works now any time a MrlCommBegin message is seen.  at that point.. a "send" call is created to sync() ..  the actual re-attaching of the devices needs to be done from a different thread so it doesn't cause a deadlock waiting on an ack in the Msg.java class.   

you can't initiate a sync call in the publishBoardInfo because there might be a slight lag between the current deviceList in arduino vs the device list being published in the MrlComm board info...

PR #693, opened 2020-05-03
oh my god .. its so clean

supertick:


src/main/java/org/myrobotlab/image/Util.java
```diff
@@ -408,6 +425,23 @@ public final static BufferedImage readBufferedImage(String filename) {
     }
   }
 
+  public final static String getImageAsString(String filename) {
+    return getImageAsString(filename, "png");
+  }
+
+  public final static String getImageAsString(String filename, String type) {
```
kwatters, line 432: seems like a better name for this method would be getImageAsBase64 ...  but I guess that's a string.

PR #694, opened 2020-05-04
changed name to base 64 - and set currentBotName in ProgramAB test

supertick:


src/test/java/org/myrobotlab/service/ProgramABTest.java
```diff
@@ -314,6 +314,7 @@ public void umlautTest() throws IOException {
   public void testLocales() {
     // have locales
     ProgramAB alice = (ProgramAB)Runtime.start("alice", "ProgramAB");
+    alice.setCurrentBotName("Alice");
```
kwatters, line 317: This unit test is supposed to use the Lloyd.aiml.  why is this change here?

kwatters:
Sorry, thought that was for the above test for umlats

PR #697, opened 2020-05-08
add simulator for servo

hairygael:


supertick:
work in progress?

hairygael:
There is no need for this anymore, I have done it via mini python scripts
in resource/Intro.
Unless we decide to implement it later into Intro.java
But then we will need to find a way to give the correct path to fetch the
proper assets, because currently it systematically fetches the VinMoov.j3o
in data/JMonkeyEngine.


Gael Langevin
Creator of InMoov

supertick:
closed per Gael's request - hopefully there is a virtual servo somewhere :)

PR #698, opened 2020-05-08
locale refactoring

supertick:


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -156,6 +157,17 @@
   
   protected String serviceVersion = null;
   
+    /**
+   * map of keys to localizations - 
+   * <pre>
+   *  Match Service with current Locale of the Runtime service
+   *  Match Runtime with current Locale of the Runtime service.
+   *  Match Runtime with Default (english) Locale
```
kwatters, line 165: I think there should be a check for "Match Service with Default (english) Locale"  prior to falling back to runtime.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -518,6 +516,34 @@ public void setProcessOOB(boolean processOOB) {
     getSession().processOOB = processOOB;
   }
 
+  /**
```
kwatters, line 519: These following changes seem unrelated to the localization support.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -929,17 +955,17 @@ public SearchPublisher getSearch() {
 
   @Override
   public void setLocale(String code) {
```
kwatters, line 957: Setting locale on the ProgramAB service changes it for the entire platform? this surly can't be correct.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
```diff
@@ -570,18 +567,25 @@ public void unsetWakeWord() {
   }
 
   @Override
-  public void setLocale(String code) {
-    locale = new Locale(code);
+  public String getLanguage() {
```
kwatters, line 570: The intention of this method is to tell you what language the speech recognizer is detecting.  it's is not the same as Runtime locale.   This seems wrong to me.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
```diff
@@ -570,18 +567,25 @@ public void unsetWakeWord() {
...
-    return locale.getLanguage();
+  public Locale getLocale() {
+    return Runtime.getInstance().getLocale();
   }
 
   @Override
-  public Locale getLocale() {
-    return locale;
+  public Map<String, Locale> getLocales() {
+    return Locale.getAvailableLanguages();
```
kwatters, line 581: This is wrong, this should be the set of locales that the speech recognizer actually supports.  This should be abstract and force the subclasses to return the list of supported locales.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
```diff
@@ -570,18 +567,25 @@ public void unsetWakeWord() {
...
+  }
+
+  @Override
+  public void setLocale(String code) {
+    // locale = new Locale(code);
+    Runtime.getInstance().setLocale(code);
```
kwatters, line 587: Setting the locale on this service should not change locale for the platform.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -1068,17 +1065,17 @@ public boolean isMute() {
...
   @Override
   public void setLocale(String code) {
-    locale = new Locale(code);
+    Runtime.getInstance().setLocale(code);
   }
 
   @Override
   public String getLanguage() {
-    return locale.getLanguage();
+    return Runtime.getInstance().getLanguage();
```
kwatters, line 1073: This should return the language associated with the currently chosen voice.  not runtime locale

src/main/java/org/myrobotlab/service/data/Locale.java
```diff
@@ -162,10 +166,16 @@ public String getDisplayCountry() {
...
   public String toString() {
     return getTag();
+  } 
+
+  public static Properties loadLocalizations(String fullPath) {
+    Properties props = new Properties();
+    try {
+      props.load(new InputStreamReader(new FileInputStream(fullPath), Charset.forName("UTF-8")));
+    } catch(Exception e) {
+      /* don't care common use case */
```
kwatters, line 176: There should at a minimum be a log message here with the exception.

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: it's preserving the same interface as before, so consider it a helper - or consider the back end has finally been normalized

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
supertick: no .. its original intention was to support the Locale.  And its interface supported changing and getting Locale information including language

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
supertick: agreed - fixed

src/main/java/org/myrobotlab/framework/Service.java
supertick: done ..

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
supertick: it supports the LocaleProvider interface - a setting getting voice operates on different data

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
supertick: I believe it should - the details of hierarchy of locales over the hierarchy of properties for multiple locales - first we need to make it work as best as possible with a single Locale .. mutliple Locales can be addressed in future functionality

src/main/java/org/myrobotlab/service/data/Locale.java
supertick: I'll update it to an a debug, but its not an error or even a warning

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: one very small addition - I don't want to split up ProgramAB into parts for this commit - originally I was working on ProgramAB, before I got squirreled with localize

src/main/java/org/myrobotlab/framework/Service.java
kwatters: Javadoc wasn't updated...

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -2842,5 +2872,67 @@ public boolean isDev() {
...
+  public String localize(String key, Object ... args) {
+     if (key == null) {
+       log.error("localize(null) not allowed");
+       return null;
+     }     
+     key = key.toUpperCase();     
+     Object prop = localization.get(key);
+     
+     if (prop == null) {
+       prop = defaultLocalization.get(key);
```
kwatters, line 2903: Will defaultLocalization ever be null?  I think you need to check that.. no?

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -929,17 +955,17 @@ public SearchPublisher getSearch() {
...
   @Override
   public void setLocale(String code) {
-    this.locale = new Locale(code);
+    Runtime.getInstance().setLocale(code);
   }
 
   @Override
   public String getLanguage() {
-    return locale.getLanguage();
+    return Runtime.getInstance().getLanguage();
```
kwatters, line 963: This should be the language from the bot that is currently loaded.  I don't think this is right.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
kwatters: This should be the language that the speech recognizer is recognizing in.  I don't think this is expected behavior.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
kwatters: There is no validation that you can set the locale of the speech recognizer that the speech recognizer supports it.  This is very misleading...

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
kwatters: Remove the interface then.  This is wrong.  Ifyou set the langauge of speech synthesis, it should choose a voice that supports the desired langauge.. or thow and exception if the speech synthesis doesn't support that language.  This is very confusing now.

src/main/java/org/myrobotlab/service/data/Locale.java
kwatters: Why woulnd't it be an error message?  if the file system croaks on you.. this would throw an IO exception.A log message should be here.  Also.. what is closing the Input stream readers, aren't you leaking file handles here?

src/main/java/org/myrobotlab/framework/interfaces/ServiceInterface.java
```diff
@@ -134,4 +134,7 @@
 
   public String getFullName();
 
+  public void loadLocalizations();
+  public void loadLocalizations(String tag);
```
kwatters, line 138: "tag" should be called "language" or "locale"  it doesn't make sense to call it "tag"

src/main/java/org/myrobotlab/service/GoogleSearch.java
```diff
@@ -94,7 +92,8 @@ public void clearFilters() {
 
   @Override
   public void setLocale(String code) {
-    locale = new Locale(code);
+    // locale = new Locale(code);
+    Runtime.getInstance().setLocale(code);
```
kwatters, line 96: You should be able to set the language of your search.  This should not be setting the locale of runtime.

src/main/java/org/myrobotlab/service/GoogleSearch.java
```diff
@@ -126,7 +125,7 @@ public SearchResults search(String searchText) throws IOException {
       String encodedSearch = URLEncoder.encode(searchText, "UTF-8");
 
       // not sure if locale is supported tag probably is ....
-      String request = "https://google.com/search?lr=lang_" + locale.getTag() + "&q=" + encodedSearch + "&aqs=chrome..69i57.5547j0j7&sourceid=chrome&ie=UTF-8";
+      String request = "https://google.com/search?lr=lang_" + Runtime.getInstance().getLocaleTag() + "&q=" + encodedSearch + "&aqs=chrome..69i57.5547j0j7&sourceid=chrome&ie=UTF-8";
```
kwatters, line 128: This should be using the locale of the google search service.. not runtime.

src/main/java/org/myrobotlab/service/GoogleSearch.java
```diff
@@ -229,7 +228,7 @@ public SearchResults search(String searchText) throws IOException {
     try {
       // can only grab first 100 results
 
-      String url = "https://www.google.com/search?lr=lang_" + locale.getTag() + "&site=imghp&tbm=isch&source=hp&q=" + searchText + "&gws_rd=cr";
+      String url = "https://www.google.com/search?lr=lang_" + Runtime.getInstance().getLocaleTag() + "&site=imghp&tbm=isch&source=hp&q=" + searchText + "&gws_rd=cr";
```
kwatters, line 231: this should not be from runtime.. this should be the locale of the google search service.   you should be able to search in french even if your current runtime locale is english.

src/main/java/org/myrobotlab/service/GoogleSearch.java
```diff
@@ -385,12 +384,12 @@ public static void main(String[] args) {
 
   @Override
   public String getLanguage() {
-    return locale.getLanguage();
+    return Runtime.getInstance().getLanguage();
```
kwatters, line 387: This means you can't search in any language other than the current runtime language.

src/main/java/org/myrobotlab/service/GoogleSearch.java
```diff
@@ -385,12 +384,12 @@ public static void main(String[] args) {
...
   @Override
   public String getLanguage() {
-    return locale.getLanguage();
+    return Runtime.getInstance().getLanguage();
   }
 
   @Override
   public Locale getLocale() {
-    return locale;
+    return Runtime.getInstance().getLocale();
```
kwatters, line 392: Should be locale of the google search service..  if that's null, then fallback to runtime locale.

src/main/java/org/myrobotlab/service/ProgramAB.java
kwatters: This language should be derived from the locale property setting for the current bot.   not from runtime.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -929,17 +955,17 @@ public SearchPublisher getSearch() {
...
   }
 
   @Override
   public Locale getLocale() {
-    return locale;
+    return Runtime.getInstance().getLocale();
```
kwatters, line 968: should be locale of the currently loaded bot.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -2697,6 +2698,12 @@ public static Double getBatteryLevel() {
   @Override
   public void setLocale(String code) {
     locale = new Locale(code);
+    // attempt to reload all services localizations
```
kwatters, line 2701: This will override the localization for every service..  if you call setLocale at the runtime level,  then you'll have to go back through and reset the locale on speech recognition and on speech synthesis to ensure that those services are still operating in the proper language.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
```diff
@@ -5,6 +5,7 @@
 import java.util.Set;
 import java.util.TreeSet;
 
+import org.apache.tools.ant.taskdefs.LoadProperties;
```
kwatters, line 8: Seems like a strange library to be importing,, is there a better import that is more generic?

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
kwatters: Get language should return the current language that the current voice is for.  This is returning the runtime langauge which is incorrect.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -1069,16 +1050,18 @@ public boolean isMute() {
...
   }
 
   @Override
   public Locale getLocale() {
-    return locale;
+    return Runtime.getInstance().getLocale();
```
kwatters, line 1064: This should be the langauge of the current voice, not runtime.

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -640,11 +629,16 @@ public InMoov2Arm getLeftArm() {
...
+    return Runtime.getInstance().getLanguage();
+  }
 
-	@Override
-	public Locale getLocale() {
-		return locale;
-	}
+  @Override
+  public Locale getLocale() {
+    return Runtime.getInstance().getLocale();
```
kwatters, line 640: This should be the locale of the InMoov2 service.. not Runtime.

src/main/java/org/myrobotlab/programab/BotInfo.java
```diff
@@ -28,32 +29,41 @@
...
     if (bot == null) {
       // lazy loading of bot - created on the first use
-      
       if (properties.containsKey("locale")) {
         bot = new Bot(name, path.getAbsolutePath(), java.util.Locale.forLanguageTag(properties.get("locale")));
       } else {
         bot = new Bot(name, path.getAbsolutePath());
       }
 
+      // merge properties - potentially there are 2 sets
```
kwatters, line 57: Where do these other properties come from and how are they managed?  This seems unrelated to localization.

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -640,11 +629,16 @@ public InMoov2Arm getLeftArm() {
 	public InMoov2Hand getLeftHand() {
 		return leftHand;
 	}
+	
+  @Override
+  public String getLanguage() {
+    return Runtime.getInstance().getLanguage();
```
kwatters, line 635: This should be the language that the InMoov2 is set to.. not Runtime's language.

src/main/java/org/myrobotlab/framework/Service.java
supertick: updated

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
supertick: updated - pushed to Service - each service has its own Locale

src/main/java/org/myrobotlab/framework/Service.java
supertick: no .. it will not. The Service constructor loads localization and default if null will always be initialized to a map - even if en.properties is not found

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: this has been pushed up to Service .. and all services use their local property - if not set then its set from runtime

src/main/java/org/myrobotlab/framework/interfaces/ServiceInterface.java
supertick: its been removed

src/main/java/org/myrobotlab/service/GoogleSearch.java
supertick: been completely removed and pushed up into Service

src/main/java/org/myrobotlab/service/GoogleSearch.java
supertick: nice catch ! .. thanks ..
fixed

src/main/java/org/myrobotlab/service/GoogleSearch.java
supertick: good catch .. these needed to be removed

src/main/java/org/myrobotlab/service/GoogleSearch.java
supertick: fixed - taken care of by parent Service

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: removed .. promoted .. normalized - no longer an issue

src/main/java/org/myrobotlab/service/Runtime.java
supertick: uses Service.setLocale .. no longer an issue

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
supertick: dunno how that got pulled in .. but all this code is "Outdated" - no longer an import

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
supertick: promoted normalized updated - no longer an issue

src/main/java/org/myrobotlab/service/InMoov2.java
supertick: wow .. i can't believe I left this in .. thanks - again .. nice catch.. removed

src/main/java/org/myrobotlab/programab/BotInfo.java
supertick: these are not the properties your looking for ... a programab pr will be coming up .. let it go...

src/main/java/org/myrobotlab/service/InMoov2.java
supertick: it is .. should be outdated .. check now

PR #701, opened 2020-05-11
cleanup indentation errors

hairygael:
some indentations occured when pulling from git to vs code and vice and vice versa.
I finally understood what was the issue, after many pull and pushes...


supertick:
Fixed indentation - eclipse and a formatter.xml is the standard for all java files.
Also merged in develop branch back to webgui_work

PR #703, opened 2020-05-12
Grog/programab webgui

supertick:


src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1064,13 +1064,31 @@ public String getAimlFile(String botName, String name) {
...
     try {
-      FileIO.toString(f);
+      ret = FileIO.toString(f);
     } catch (IOException e) {
       log.error("getAimlFile threw", e);
     }    
     return ret;
   }
 
+  public void saveAimlFile(String botName, String filename, String data) {
```
kwatters, line 1074: This is ok, however, I'd like to see something more like using the addCategory interface.. and updating the writeAiml method in the future.

PR #705, opened 2020-05-12
opencv properties and coco.names

supertick:


src/main/java/org/myrobotlab/opencv/OpenCVFilterYolo.java
```diff
@@ -90,8 +79,8 @@ public OpenCVFilterYolo() {
     this(null);
   }
 
-  private void downloadYoloModel() {
-    log.info("downloadYoloModel - begin");
+  private void checkFiles() {
+    log.info("checkFiles - begin");
     File yoloHome = new File(darknetHome);
     if (!yoloHome.exists()) {
```
kwatters, line 85: this method isn't downloading files anymore.. it shouldn't create the files.  additionally, this method doesn't return and value, so the caller of this method doesn't know if the files were valid.  This method should either be removed, or it should include a return value that has some useful meaning.

src/main/java/org/myrobotlab/opencv/OpenCVFilterYolo.java
```diff
@@ -105,66 +94,23 @@ private void downloadYoloModel() {
...
-    }
-    log.info("downloadAndCache - end");
+    log.info("checkFiles - end");
   }
 
   private void loadYolo() {
     log.info("loadYolo - begin");
 
     // If the model isn't there, we should download it and cache it.
     log.info("Staritng yolo download verification");
```
kwatters, line 112: this log message is wrong, it doesn't download anymore..  same with the following log message.  and there's no validation that "checkFiles" was successful.

src/main/java/org/myrobotlab/service/OpenCV.java
```diff
@@ -455,7 +455,7 @@ static public ServiceType getMetaData() {
     meta.addDependency("com.github.axet", "vget", "1.1.34");
 
     // yolo models
-    meta.addDependency("yolo", "yolov2", "v2", "zip");
+    meta.addDependency("yolo", "yolov2", "v3", "zip");
```
kwatters, line 458: When did we start supporting yolo v3?!  Was that added recently?  I didn't think that was supported in the current version of opencv...

src/main/java/org/myrobotlab/opencv/OpenCVFilterYolo.java
supertick: will remove it

src/main/java/org/myrobotlab/opencv/OpenCVFilterYolo.java
supertick: this is dead to me...

src/main/java/org/myrobotlab/service/OpenCV.java
supertick: changing to "yolo", "yolov2", "0.0.2"

PR #707, opened 2020-05-14
add type to leftArm

hairygael:


supertick:
What was the purpose of "type" ?

supertick:
This has conflicts, type is unused,  other pins not checked for null - we should work on configuration from the web

hairygael:
I can't remember what issue I was trying to solve with this. It's about a year ago.
I only remember that the right arm didn't have the issue I was trying to fix, that is why I had copy pasted(maybe?) things from the right arm to the left. Anyway, it was a draft and the build din't like it.

PR #709, opened 2020-05-16
simplify controller and angle publishing

supertick:


src/main/java/org/myrobotlab/service/IntegratedMovement.java
```diff
@@ -511,7 +511,7 @@ public Thread getEngine(String arm) {
   }
 
   @Override
-  public Map<String, Double> publishJointAngles(HashMap<String, Double> angleMap) {
+  public AngleData publishJointAngles(AngleData angleMap) {
```
kwatters, line 514: If we only publish a single value in the AngleData this method should be publishJointAngle .. not plural. Also, AngleData is not a "map"

src/main/java/org/myrobotlab/service/InverseKinematics3D.java
```diff
@@ -239,9 +240,10 @@ public void publishTelemetry(String name) {
       // - 180 to + 180 ?
       double angle = MathUtils.radToDeg(theta) + l.getOffset();
       angleMap.put(jointName, (double) angle % 360.0F);
+      invoke("publishJointAngles", new AngleData(jointName, (double) angle % 360.0F));
```
kwatters, line 243: Angles plural is incorrect here.  AngleData is singular.

src/main/java/org/myrobotlab/service/InverseKinematics3D.java
```diff
@@ -239,9 +240,10 @@ public void publishTelemetry(String name) {
       // - 180 to + 180 ?
       double angle = MathUtils.radToDeg(theta) + l.getOffset();
       angleMap.put(jointName, (double) angle % 360.0F);
+      invoke("publishJointAngles", new AngleData(jointName, (double) angle % 360.0F));
       log.info("Servo : {}  Angle : {}", jointName, angleMap.get(jointName));
     }
-    invoke("publishJointAngles", angleMap);
+    // invoke("publishJointAngles", angleMap);
```
kwatters, line 246: This was useful to ensure that you have a consistent set of angles from which to do your updates.. this means that for the inmoov arm, you will need to publish 4x the number of messages to get the data.. and it's up to the listener to accumulate all of those together.. seems bad to remove this.

src/main/java/org/myrobotlab/service/InverseKinematics3D.java
```diff
@@ -362,9 +364,8 @@ public static void main(String[] args) throws Exception {
   }
 
   @Override
-  public Map<String, Double> publishJointAngles(HashMap<String, Double> angleMap) {
-    // TODO Auto-generated method stub
-    return angleMap;
+  public AngleData publishJointAngles(AngleData angleData) {
+    return angleData;
```
kwatters, line 368: it's not plural.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -417,6 +424,11 @@ public void attach(String controllerName, Integer pin, Double pos) {
     attach(controllerName, pin, pos, null);
   }
 
+  @Override
+  public AngleData publishJointAngles(AngleData angle) {
```
kwatters, line 428: not plural.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -891,11 +901,11 @@ protected void processMove(Double newPos, boolean blocking, Long timeoutMs) {
...
 
-    for (String controller : this.getControllers()) {
-      // TODO: just have a direct reference to the controllers 
-      // avoid this type cast and runtime lookup!!!!!
-      ((ServoController)Runtime.getService(controller)).onServoMoveTo(this);
-    }
+    // GroG: I think in the long run this direct call vs using invoke/send is less preferrable
+    // thinking on a distributed network level you can't do this when the other thing is in
+    // a different process
+    ((ServoController)Runtime.getService(controller)).onServoMoveTo(this);
```
kwatters, line 907: for a future pull request.. thanks for adding the comment in the code so we don't forget about it.

src/main/java/org/myrobotlab/service/interfaces/IKJointAnglePublisher.java
```diff
@@ -1,12 +1,11 @@
...
-import java.util.HashMap;
-import java.util.Map;
+import org.myrobotlab.service.data.AngleData;
 
 public interface IKJointAnglePublisher {
 
   public String getName();
 
-  public Map<String, Double> publishJointAngles(HashMap<String, Double> angleMap);
+  public AngleData publishJointAngles(AngleData angle);
```
kwatters, line 9: not plural in this refactor.  Also,  I though we were creating a new interface.. not modifying this existing one..

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
supertick: fixed

src/main/java/org/myrobotlab/service/InverseKinematics3D.java
supertick: fixed

src/main/java/org/myrobotlab/service/IntegratedMovement.java
supertick: fixed

src/main/java/org/myrobotlab/service/interfaces/IKJointAnglePublisher.java
supertick: fixed to singular, no there was no mention of creating a new interface or its the name of that interface

PR #716, opened 2020-05-28
Grog/meta

supertick:


src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -125,35 +66,35 @@ public void execScript(String someScriptName) {
     }
   }
 
-  /**
-   * This method will load a python file into the python interpreter.
-   */
+	/**
```
kwatters, line 69: @Grog check your auto-format settings.. i think they've switched to tabs not spaces .. at least for this file.

PR #719, opened 2020-05-31
adding stop to disable

supertick:


supertick:
Added new test to ServoTest to lock in behavior.

PR #732, opened 2020-06-07
Worke

supertick:


src/main/java/org/myrobotlab/service/interfaces/SpeechSynthesisControl.java
```diff
@@ -0,0 +1,28 @@
...
+
+  public String setVoice(String name);
+  
+  public double setVolume(double volume);
+  
+  public boolean setMute(boolean mute);
+  
+  public void replaceWord(String word, String substitute);
+  
+  public String publishSpeak(String text);
```
kwatters, line 18: the publish versions of this probably belong in a different interface.. maybe ControlPublisher?

PR #733, opened 2020-06-08
reset setAutoDisableDefault(true)

hairygael:


supertick:
I see InMoov2.setAutoDisable(bool) ...  this will set the currently started inmoov servos to autoDisable(b) ..
but new servos can be created after this function is called and they will not be auto disabled ..
Is it desirable that "all servos" be autoDisabled ? This can be done inInMoov - because it can examine each new service created, and if its a servo set it to autoDisable true ...

hairygael:
It was a security, you start by default with autoDisable. Beginners have no idea what is autoDisable in the config or the script and they would burn servos before finding out. With InMoov, it's easy to burn them, because servos tend to try to reach or maintain positions that are heavy to reach.

supertick:
Yes, makes complete sense - a good idea - is fixed in - #826  servos created before and after i01 will auto-magically get their setAutoDisable set to true

PR #738, opened 2020-06-12
Intro fixes

supertick:


src/main/java/org/myrobotlab/service/JMonkeyEngine.java
```diff
@@ -1391,6 +1426,16 @@ public void onCvData(CvData data) {
     // FIXME - do something interesting ... :)
   }
 
+  @Override
+  public void onJointAngles(Map<String, Double> angleMap) {
+    for (String name : angleMap.keySet()) {
+      ServiceInterface si = Runtime.getService(name);
+      if (si instanceof Servo) {
```
kwatters, line 1433: should be instance of ServoControl not Servo.. :)

PR #745, opened 2020-06-20
East detector

supertick:


src/main/java/org/myrobotlab/opencv/OpenCVFilterTextDetector.java
```diff
@@ -0,0 +1,364 @@
...
+      RotatedRect box = boxes.get(i); 
+      confidencesFV.position(i);
+      // we don't have text yet, that will be filled in later by the ocr step.
+      DetectedText dt = new DetectedText(box, confidencesFV.get(), null);
+      goodOnes.add(dt);
+    }
+    return goodOnes;
+  }
+
+  // utilty helper function to put an array of floats into a javacpp float pointer
```
supertick: perhaps generalized type conversion methods should be promoted to the org.myrobotlab.opencv as helper methods other filters can use ?

src/main/java/org/myrobotlab/service/TesseractOcr.java
```diff
@@ -35,70 +44,60 @@
...
 
-  public TesseractOcr(String n, String id) {
-    super(n, id);
+  private void initModel(String lang) {
+    api = new TessBaseAPI();
+    String tessData = "resource/TesseractOcr";
+    // load the models
+    if (api.Init(tessData, lang) != 0) {
+      log.warn("Unable to load tesseract model in {} with language {}", tessData, lang);
+      // TODO: what's the lifecycle if a service fails to start?  maybe we should throw?
```
supertick: when the service fails to initialize it would warrant communicating back to the "user"
so the logging statement should be changed to:
warn("Unable to load tesseract model in %s with language %s", tessData, lang);  - this will log and broadcast the warning message to any UI

src/main/java/org/myrobotlab/service/TesseractOcr.java
```diff
@@ -35,70 +44,60 @@
...
-    // A class called PIXConversions will convert directly
-    // but I'm afraid to import the library in that it might clash with the
-    // version
-    // javacv is expecting.. so we are going to do it the "file" way :P
-    File temp = File.createTempFile("tesseract", ".jpg");
+    // The tesseract api has some issues reading the image directly as a byte array.
+    // so for now... until that changes, we'll write a temp file to be ocr'd and 
+    // TODO: where is this.. and i think we need to randomize the name  of it
+    // or synchronize this method.
+    File temp = File.createTempFile("tesseract", ".png");
```
supertick: Perhaps manage your own temp file in the resource or data dir of TesseractOcr ?
I might be wrong, but I think your file is only guaranteed to be removed with the jvm exits, so this becomes a burden of file handles for long running processes

src/main/java/org/myrobotlab/opencv/OpenCVFilterTextDetector.java
kwatters: move them to the current image.Util class.. we can move them around somewhere later if we like.

src/main/java/org/myrobotlab/service/TesseractOcr.java
kwatters: done

src/main/java/org/myrobotlab/service/TesseractOcr.java
kwatters: good idea.  done.

PR #746, opened 2020-06-26
Servicetype to metadata

supertick:


src/main/java/org/myrobotlab/framework/interfaces/ServiceInterface.java
```diff
@@ -9,9 +9,10 @@
 import org.myrobotlab.framework.MRLListener;
 import org.myrobotlab.framework.MethodEntry;
 import org.myrobotlab.framework.Outbox;
+import org.myrobotlab.service.interfaces.ServiceLifeCycleListener;
 
 public interface ServiceInterface
-    extends ServiceQueue, LoggingSink, NameTypeProvider, MessageSubscriber, MessageSender, StateSaver, Invoker, StatePublisher, StatusPublisher, ServiceStatus, Attachable {
+    extends ServiceLifeCycleListener, ServiceQueue, LoggingSink, NameTypeProvider, MessageSubscriber, MessageSender, StateSaver, Invoker, StatePublisher, StatusPublisher, ServiceStatus, Attachable {
```
kwatters, line 15: seems like the service interface should implement the ServiceLifeCycle interface... unless you're saying generically that a service can generically listen to another services lifecycle..  is that what you're intending here?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -105,7 +106,7 @@
  * check for 64 bit OS and 32 bit JVM is is64bit()
  *
  */
-public class Runtime extends Service implements MessageListener, RemoteMessageHandler, Gateway, LocaleProvider {
+public class Runtime extends Service implements MessageListener, ServiceLifeCyclePublisher, RemoteMessageHandler, Gateway, LocaleProvider {
```
kwatters, line 109: why is runtime the only thing that is publishing lifecycle stuff?  i mean. i'd expect services to implement a SerivceLifeCycle  interface (start/stop/created..etc..)    .. i guess services could publish when their lifecycle was called to notify other listeners, however that should happen after the service processes it's lifecycle request.

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
```diff
@@ -26,9 +26,7 @@
 
 package org.myrobotlab.service.abstracts;
```
kwatters, line 27: changes to motor control should be done in a different pull request. please.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -24,6 +24,7 @@
 import org.myrobotlab.service.interfaces.KeyConsumer;
```
kwatters, line 24: Changes to speech synthesis should be done in a different pull request from the service life cycle. please

src/main/java/org/myrobotlab/service/interfaces/JoystickListener.java
```diff
@@ -1,10 +1,9 @@
 package org.myrobotlab.service.interfaces;
```
kwatters, line 1: changes to joystick interface should be done in a different pull request from service lifecycle please.

src/main/java/org/myrobotlab/service/interfaces/ServiceLifeCyclePublisher.java
```diff
@@ -0,0 +1,17 @@
+package org.myrobotlab.service.interfaces;
```
kwatters, line 1: This should probably be renamed to ServiceLifeCycle not publisher. the publisher interface would be "publishStarted"   not  "started"

src/main/java/org/myrobotlab/framework/interfaces/ServiceInterface.java
supertick: correct - the abstract base class Service now implements a set of callbacks,
the "interested" services can simply override these if they want to handle the notification events of
registered, created, started, stopped and released.
If they are not interested, they simply do not override.

src/main/java/org/myrobotlab/service/Runtime.java
supertick: Some events can be handled inside a class - started and stopped
Other events cannot be handled inside any service .. e.g. registered, released, created
The one thing other services are guaranteed is runtime exists - therefore since registered, released, and created are already handled by runtime, it was logical that it be responsible for broadcasting the remainder of the events started and stopped

src/main/java/org/myrobotlab/service/interfaces/ServiceLifeCyclePublisher.java
supertick: Sure .. I'll change it to ServiceLifeCycle.  Its a publishing interface, however registered and released have been around for bajillion years (as string values), so even though it is a publishing interface I'll wait to refactor to publishRegistered etc for another refactor)

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
supertick: These file only get changed updated, because I'm keeping it in sync with develop.  These are merge conflicts that have been resolved, because changes occurred on this file and were pushed into develop from another pr.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
supertick: These file only get changed updated, because I'm keeping it in sync with develop.  These are merge conflicts that have been resolved, because changes occurred on this file and were pushed into develop from another pr.

src/main/java/org/myrobotlab/service/interfaces/JoystickListener.java
supertick: These file only get changed updated, because I'm keeping it in sync with develop.  These are merge conflicts that have been resolved, because changes occurred on this file and were pushed into develop from another pr.

kwatters:
Sorry pull request too large.  If you want to do MetaData / ServiceType change, do that as a separate pull request.

PR #747, opened 2020-06-27
Servicetype to metadata 01

supertick:


kwatters:
Sorry.. pull request too big.  If we want to do "ServiceType" to "MetaData" change,, make a PR for that change only.

PR #749, opened 2020-06-30
smallest service life cycle i can make :(

supertick:


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -3023,4 +2536,81 @@ public String getLocaleTag() {
...
+  @Override
+  public boolean isType(String clazz) {
+    // probably a bad idea - but nice for lazy people
+    if (!clazz.contains(".")) {
+      clazz = String.format("org.myrobotlab.service.%s", clazz);
+    }   
+    return serviceClass.equals(clazz);
+  }
+
+  @Override
```
kwatters, line 2582: These method names don't make sense to me.  When are these called by the framework?  Does a service get an  "onCreated" called in addition to calling create ?  Or is this for peer services to notify a service...  very confusing.

src/main/java/org/myrobotlab/framework/ServiceType.java
```diff
@@ -32,28 +33,57 @@
...
-    return link;
-  }
-
-  public void setLink(String link) {
-    this.link = link;
-  }
-
-  String state = null;
-  Integer workingLevel = null;
+  public static ServiceType fromMetaData(MetaData meta) {
```
kwatters: Whats the difference between this class and MetaData ..  I can't see any justification for having both.

src/main/java/org/myrobotlab/framework/ServiceType.java
```diff
@@ -96,181 +115,232 @@ public ServiceType(String name) {
...
+    }
+    
+    for (ServiceReservation sr : peers.values()) {
+      sb.append(sr).append("\n");
+    }
+    
+    return sb.toString();
+  }
+
+  // GAH ! - more convertions for smaller pr :(
```
kwatters, line 304: When I ask for smaller PRs i mean a smaller number of things being fixed in the PR.  not a smaller number of lines.

src/main/java/org/myrobotlab/framework/ServiceType.java
supertick: Ha !  356 files diff - and refusal to look at it.  Yes there is no difference, that's what I have been saying in my large pull request.  Merging the two means a large amount of file imports will change.  I UN-MERGED the two classes to satisfy the small pr .. you can not have both. Pick one or the other.

src/main/java/org/myrobotlab/framework/ServiceType.java
supertick: I said that making ServiceType and MetaData the same thing would touch a large number of files.  To keep the number diffs between files small (down to 35 vs 365) these converters were necessary.

src/main/java/org/myrobotlab/framework/repo/ServiceData.java
```diff
@@ -286,62 +435,59 @@ public boolean save(String filename) {
...
+  /**
+   * Start at root and build all the meta data - add 
+   * 
+   * @param serviceName
+   * @param serviceType
+   * @return
+   */
+  public static Plan getPlan(String serviceName, String serviceType) {
+    
+    Plan root = new Plan();
```
kwatters, line 453: This is really just a set (treeset)  .. it's a little difficult to justify a whole new class just to get you a toString method..

src/main/java/org/myrobotlab/i2c/I2CBus.java
```diff
@@ -133,4 +137,61 @@ public boolean isLocal() {
...
+  public boolean isType(Class<?> clazz) {
+    return isType(clazz.getCanonicalName());
+  }
+
+  @Override
+  public boolean isType(String clazz) {
+    return getClass().getCanonicalName().equals(clazz);
+  }
+
+  @Override
```
kwatters, line 173: remove these methods. they don't do anything.

src/main/java/org/myrobotlab/sensor/TimeEncoder.java
```diff
@@ -398,4 +401,56 @@ public void setPos(Double pos) {
...
+    return false;
+  }
+
+  @Override
+  public boolean isType(String clazz) {
+    // TODO Auto-generated method stub
+    return false;
+  }
+
+  @Override
```
kwatters, line 433: remove these methods, they don't do anything

src/main/java/org/myrobotlab/framework/Service.java
supertick: Any service has the opportunity to override and capture
onRegistered
onCreated
onStarted
onStopped
and
onReleased

registered and released have been events for a very long time.
The other callbacks were added to support create, start, and stopped.

One part of the complexity, is when a service starts and their are pre-existing services, should the starting service get existing service events?  I debated on this for a little, but the use case typically is fulfilling possibilities of other services for which it can attach.  If that is the most common use case, then its advantageous the new service see existing service events.  It makes a Service implementation of business logic less clean if it has to both wait for future events to handle in addition to query existing services manually.

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
```diff
@@ -1468,14 +1468,18 @@ public void onPointCloud(PointCloud pc) {
...
   // auto Register
-  public void onRegistered(Registration registration) throws Exception {
-    // new service - see if we can virtualize it
-    log.info("{}.onRegistered({})", getName(), registration);
-    if (registration.getName().contentEquals("i01.head.jaw")) {
-      log.info("here");
-    }
-    if (autoAttach) {
-      attach(registration.getFullName());
+  public void onRegistered(Registration registration) {
```
kwatters, line 1471: is this a notification of all services that are registered?  Does this get invoked for every service that gets created?  Does that mean that the number of times that onRegistered gets called its O^2 on the number of services?  ouch.

src/main/java/org/myrobotlab/framework/repo/ServiceData.java
supertick: Its deleted in a future pr.  Was "left" here to reduce number of changes

src/main/java/org/myrobotlab/service/Proxy.java
```diff
@@ -7,6 +7,7 @@
 import org.myrobotlab.framework.Registration;
 import org.myrobotlab.framework.Service;
 import org.myrobotlab.framework.ServiceType;
+import org.myrobotlab.framework.repo.ServiceData;
```
kwatters, line 10: Why do we have service Data when all this info is supposed to be on the MetaData object?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -105,7 +107,7 @@
  * check for 64 bit OS and 32 bit JVM is is64bit()
  *
  */
-public class Runtime extends Service implements MessageListener, RemoteMessageHandler, Gateway, LocaleProvider {
+public class Runtime extends Service implements MessageListener, ServiceLifeCycle, RemoteMessageHandler, Gateway, LocaleProvider {
```
kwatters, line 110: Service should implement ServiceLifeCycle not Runtime to ensure all services properly implement their lifecycle methods.

src/main/java/org/myrobotlab/sensor/TimeEncoder.java
supertick: They both implement Attachable.  Usually, this is an indication that this TimeEncoder and I2CBus should have been services. Not doing it in this pr.

src/main/java/org/myrobotlab/service/Proxy.java
supertick: Tried to merge, got slapped for too many changes.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1970,17 +2000,27 @@ public Runtime(String n, String id) {
...
         }
 
-        repo = (IvyWrapper) Repo.getInstance(options.libraries, "IvyWrapper"); // previously
-                                                                               // was
-                                                                               // not
-                                                                               // (IvyWrapper)
+        repo = (IvyWrapper) Repo.getInstance(options.libraries, "IvyWrapper"); 
+        
+        boolean readyForPrimetime = false;
+        if (options.spawnedFromAgent && readyForPrimetime) {
```
kwatters: dead code

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3458,4 +3498,39 @@ static public void setAllLocales(String code) {
     }
   }
 
+  @Override
+  public String created(String serviceName) {
```
kwatters, line 3501: runtime doesn't do anything when it's created?  what's the point of this method?  if it's framework stuff and it's for publishing .. it should be called publishCreated .. publishStarted.. publishStopped  etc..

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3458,4 +3498,39 @@ static public void setAllLocales(String code) {
...
+  public String started(String serviceName) {
+    return serviceName;
+  }
+
+  @Override
+  public String stopped(String serviceName) {
+    return serviceName;
+  }
+
+  public static void setPeer(String fullKey, String actualName, String serviceType) {
```
kwatters, line 3515: seems like this should be on the base service class, not runtime.

src/main/java/org/myrobotlab/service/interfaces/ServiceLifeCycle.java
```diff
@@ -0,0 +1,17 @@
+package org.myrobotlab.service.interfaces;
+
+import org.myrobotlab.framework.Registration;
+
+public interface ServiceLifeCycle {
+  
+  public Registration registered(Registration registration);
+
+  public String created(String serviceName);
```
kwatters, line 9: If this is actually a publishing sort of interface. it should be publishCreated / publishStarted ... etc..  and these should just be pushed down to the Service class for their implementation.  Generally these shouldn't be overridden in subclasses, and this should be called ServiceLifeCyclePublisher ...

src/main/java/org/myrobotlab/service/meta/abstracts/Meta.java
```diff
@@ -0,0 +1,20 @@
+package org.myrobotlab.service.meta.abstracts;
+
+import java.util.Map;
+import java.util.TreeMap;
+
+import org.myrobotlab.framework.ServiceReservation;
+
+abstract public class Meta {
```
kwatters, line 8: Why this class and the MetaData class?  that seems silly to have both.. and the still kicking around ServiceType class.

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
supertick: Just O(n).

src/main/java/org/myrobotlab/service/meta/abstracts/MetaData.java
```diff
@@ -0,0 +1,418 @@
...
+ * It's dependencies, categories and other meta information.
+ * 
+ * It also describes its list of peers. They define a list of group this 
+ * service expects to interact with.  When a service is started it gets
+ * a copy of MetaData based on its instance name. When a new instance is
+ * created a list of ServiceData.overrides are consulted, to allow the user
+ * to override actual name and type information.
+ * 
+ */
+public class MetaData implements Serializable {
```
kwatters: This class should me marked as abstract and probably renamed to AbstractMetaData.

src/main/java/org/myrobotlab/service/Runtime.java
supertick: Services don't register nor release themselves .. Runtime does.
Runtime is responsible for registering, creating, starting, stopping and releasing all services.  There is no point that the service do this.

src/main/java/org/myrobotlab/service/Runtime.java
supertick: legacy - previously it has always been register(ed) and release(ed).
I didn't want to incur too many changes.
It can be done in another pr.

src/main/java/org/myrobotlab/service/Runtime.java
supertick: No .. the opportunity to override meta information, cannot be done in the created service. It's "before" that lifecycle phase has begun.

src/main/java/org/myrobotlab/service/interfaces/ServiceLifeCycle.java
supertick: It always has - its legacy that it wasn't publishRegistered.
It can be done in a later pr.

src/main/java/org/myrobotlab/service/meta/abstracts/Meta.java
supertick: don't get me started.

src/main/java/org/myrobotlab/service/meta/abstracts/MetaData.java
supertick: done.

src/main/java/org/myrobotlab/i2c/I2CBus.java
supertick: Same as TimeEncoder this should have been a service.
Not doing it this pr.

src/main/java/org/myrobotlab/service/Runtime.java
supertick: good catch .. must have been debugging kruft .. removed.

src/main/java/org/myrobotlab/framework/ServiceType.java
kwatters: I was referring to the motor/speech and joystick changes that leaked into the other pull request.  i don't mind a boiler plate change to a lot of files. that's fine.

supertick: sure .. but I'll save the many boiler plate updates for next pr

src/main/java/org/myrobotlab/service/Runtime.java
kwatters: If it's "before" the lifecycle, it should be defined as part of the lifecycle,  no?  i mean. you're defining a step prior to the started phase of the lifecycle.

src/main/java/org/myrobotlab/framework/Service.java
kwatters: I feel like there should be another phase of the lifecycle between created and started..  or we need to consider that started will always load resources...  it seems like there is always a phase between the service being created.. where you are configuring it and loading resources.. and then finally you are starting the service ..

src/main/java/org/myrobotlab/service/Runtime.java
kwatters: We probably don't need an interface for ServiceLifeCycle if runtime is the only thing that implements it..

src/main/java/org/myrobotlab/service/Runtime.java
kwatters: Ok.. I think this is related to my other comment..  I think there shoudl be a step between created and started.. where you are configuring and initializing resources.

PR #751, opened 2020-07-02
massive grunge work

supertick:


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -414,7 +414,7 @@ public String getDataInstanceDir() {
   // ============== resources begin ======================================
 
   /**
-   * Non-static getResourceDir() will return /resource/{ServiceType}
+   * Non-static getResourceDir() will return /resource/{MetaData}
```
kwatters: This seems wrong.. the resource directory is actually based off the string identifier of the service type.

src/main/java/org/myrobotlab/framework/repo/Repo.java
```diff
@@ -290,7 +290,7 @@ public String getFullTypeName(String serviceType) {
       }
      
 
-      ServiceType st = sd.getServiceType(type);
+      MetaData st = sd.getServiceType(type);
```
kwatters: the method name for getServiceType should probably be renamed?  It doesn't return a ServiceType anymore.

src/main/java/org/myrobotlab/framework/Service.java
supertick: Your correct - overzealous search and replace ... corrected with example

src/main/java/org/myrobotlab/framework/repo/Repo.java
supertick: Agree, updated

src/main/java/org/myrobotlab/arduino/Msg.java
```diff
@@ -2,18 +2,16 @@
 
 
 import java.io.ByteArrayOutputStream;
-import java.io.OutputStream;
+import java.io.FileOutputStream;
```
kwatters: This file is generated.. was the template updated?

supertick: I tried to do organize import based on a package level, but it didn't work.
No the file wasn't regenerated, it doesn't use the import, so "organized" means it goes away.
Next time its generated it will re-create the not used OutputStream reference.

kwatters: ok.. but this file shouldn't be modified unless it's generated...

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -40,20 +41,33 @@
    * substitutions are phonetic substitutions for a specific instance of speech
    * synthesis service
    */
-  Map<String, String> substitutions = new HashMap<String, String>();
-  
+  transient protected Map<String, String> substitutions = new ConcurrentHashMap<String, String>();
```
kwatters, line 44: These changes are unrelated to the metadata/service type change. please revert for this PR.

src/main/java/org/myrobotlab/service/interfaces/SpeechSynthesis.java
```diff
@@ -5,6 +5,7 @@
 import org.myrobotlab.framework.interfaces.NameProvider;
```
kwatters, line 5: these changes seem unrelated to metadata / service type changes.. please revert

src/main/java/org/myrobotlab/service/interfaces/SpeechSynthesisControl.java
```diff
@@ -0,0 +1,18 @@
+package org.myrobotlab.service.interfaces;
```
kwatters, line 1: this has nothing to do with the service type / metadata change. please revert from this pr.

src/main/java/org/myrobotlab/service/meta/DocumentPipelineMeta.java
```diff
@@ -1,41 +1,37 @@
...
-    meta.addDependency("com.thoughtworks.xstream", "xstream", "1.4.9");
-
-    // FIXME - add service page, python script, give example of how to use
-    meta.setAvailable(false);
-    return meta;
-  }
-
-  
-}
+public class DocumentPipelineMeta extends MetaData {
```
kwatters, line 8: These are all tabs not spaces.  Please validate your formatting in your IDE.

src/main/java/org/myrobotlab/service/meta/DruppNeckMeta.java
```diff
@@ -1,37 +1,31 @@
...
+	public final static Logger log = LoggerFactory.getLogger(DruppNeckMeta.class);
+
+	/**
+	 * This class is contains all the meta data details of a service. It's peers,
+	 * dependencies, and all other meta data related to the service.
+	 * 
+	 */
+	public DruppNeckMeta() {
+
+		Platform platform = Platform.getLocalInstance();
```
kwatters: the white space here changed.. now it's tabs not spaces. please fix.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -422,6 +436,9 @@ public void attachSpeechRecognizer(SpeechRecognizer recognizer) {
    */
   @Override
   public void setVolume(double volume) {
+    if (audioFile == null) {
```
kwatters, line 439: this change is not related to metadata.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -500,6 +517,10 @@ AudioData play(AudioData data, boolean block) {
     if (block) {
       data.mode = AudioData.MODE_BLOCKING;
     }
+    if (audioFile == null) {
```
kwatters, line 520: this change is not related to the metadata change.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -539,8 +561,10 @@ public String publishSpeechRequested(String toSpeak) {
     toSpeak = toSpeak.toLowerCase();
 
     // process substitutions
-    for (String substitute : substitutions.keySet()) {
-      toSpeak = toSpeak.replace(substitute, substitutions.get(substitute));
+    if (substitutions != null) {
```
kwatters, line 564: this change is not related to the metadata change.

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -581,16 +605,25 @@ public String publishSpeechRequested(String toSpeak) {
...
+   * replace word for example : worke could get substituted to worky or work-ee
    * or "something" that phonetically works for the current speech synthesis
    * service
    * 
    * @param key
    * @param replace
    */
-  public void addSubstitution(String key, String replace) {
-    substitutions.put(key.toLowerCase(), replace.toLowerCase());
+  public void replaceWord(WordFilter filter) {
```
kwatters, line 625: this change is not related to the metadata change.

src/main/java/org/myrobotlab/service/interfaces/SpeechSynthesisControlPublisher.java
```diff
@@ -0,0 +1,18 @@
+package org.myrobotlab.service.interfaces;
```
kwatters, line 1: This has nothing to do with the metadata / service  type changes.

kwatters (changes requested):
white space changes need to be addressed.  this PR converts all of the Meta classes to tabs not spaces.

src/main/java/org/myrobotlab/service/meta/Deeplearning4jMeta.java
```diff
@@ -1,71 +1,65 @@
 package org.myrobotlab.service.meta;
```
kwatters, line 1: formatting here changed from spaces to tabs. please fix.

src/test/java/org/myrobotlab/service/WorkETest.java
```diff
@@ -8,27 +8,25 @@
 import org.junit.AfterClass;
```
kwatters, line 8: I don't see any asserts on this test. please put the @Ignore back unless it actually tests something.

src/main/java/org/myrobotlab/arduino/Msg.java
supertick: its been reverted to its original generated, yet unused java-lint warning import state

src/main/java/org/myrobotlab/service/meta/DruppNeckMeta.java
supertick: all meta classes have been re-formatted

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
supertick: an npe check ?  really?
If it was all converted to do pub/sub properly then this wouldn't be needed, but I can't do all of that at once.
Ok, I can let speech synthesis classes explode until the next pr

src/main/java/org/myrobotlab/service/meta/Deeplearning4jMeta.java
supertick: all meta classes have been reformatted after the regex replacement

src/test/java/org/myrobotlab/service/WorkETest.java
supertick: Heh, good point ;)
I'll test-driven dev .. soon as I figure out what its supposed to do ...

src/main/java/org/myrobotlab/service/meta/abstracts/AbstractSpeechSynthesisMeta.java
```diff
@@ -0,0 +1,26 @@
...
+   * dependencies, and all other meta data related to the service.
+   * 
+   */
+  public AbstractSpeechSynthesisMeta() {
+
+    addPeer("audioFile", "AudioFile", "audioFile");
+    addCategory("speech");
+
+    addDependency("org.myrobotlab.audio", "voice-effects", "1.0", "zip");
+    addDependency("javazoom", "jlayer", "1.0.1");
```
kwatters: This class is useless.  None of the subclasses call  super() as far as i can tell. . and even if they did.. why would every speech synt force jlayer and mp3spi as a dependency.  Either make this class useful or remove it.

src/main/java/org/myrobotlab/service/meta/abstracts/MetaData.java
```diff
@@ -8,200 +8,137 @@
...
- * to override actual name and type information.
+ * It also describes its list of peers. They define a list of group this service
+ * expects to interact with. When a service is started it gets a copy of
+ * MetaData based on its instance name. When a new instance is created a list of
+ * ServiceData.overrides are consulted, to allow the user to override actual
+ * name and type information.
  * 
  */
-public class AbstractMetaData implements Serializable {
+public class MetaData implements Serializable {
```
kwatters, line 29: MetaData is such an overused term. we should probably rename this to ServiceMetaData ... calling it just "metadata" is very vague to a noob, would think.

src/main/java/org/myrobotlab/service/meta/abstracts/AbstractSpeechSynthesisMeta.java
supertick: The dependencies can go but the peer is critical, so is this class.
Calling super? Search for derived classes and you'll find all the meta speech synthesis that cache their mp3 files.
If you removed it you'd have 6 speech synthesis services without cache.

kwatters: This seems to have been my confusion.  I didn't realize that the constructor of the parent class was called implicitly,  I was always taught that you had to explicitly call super() to call the parent method... I assumed that applied to consturctros also.  (that being said.. the mp3 stuff is really part of audioFile ..so.. cool.)

PR #755, opened 2020-07-04
Python and other JUnit runtime errors fixed with Javadoc errors fixed

supertick:
jenkins died - try 2


src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -561,19 +600,25 @@ public void openScriptFromFile(String filename) throws IOException {
     openScript(filename, data);
   }
 
-  public void onRegistered(Registration r) {
-    ServiceInterface s = r.service;
+  public void onStarted(String serviceName) {
```
kwatters, line 603: this feels really odd to me.. this stuff should be done in startService ...  not in the constructor. and not in a call back like  onStarted... no?

src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -266,6 +265,46 @@ public Python(String n, String id) {
 
     createPythonInterpreter();
     attachPythonConsole();
+
```
kwatters, line 268: Why did this move here?  seems like this should be in startService...  keep the constructors lightweight.

PR #757, opened 2020-07-05
Grog/develop delta

supertick:


supertick:
merged astro changes from webgui_work

PR #758, opened 2020-07-11
Grog/meta data generation fix

supertick:


src/main/java/org/myrobotlab/service/meta/abstracts/AbstractSpeechSynthesisMeta.java
```diff
@@ -12,8 +12,9 @@
    * dependencies, and all other meta data related to the service.
    * 
    */
-  public AbstractSpeechSynthesisMeta() {
+  public AbstractSpeechSynthesisMeta(String name) {
 
+super(name);
```
kwatters: white space. please indent.  :)

src/test/java/org/myrobotlab/framework/ServiceLifeCycleTest.java
```diff
@@ -41,9 +41,20 @@ public static void setUpAfterClass() throws Exception {
...
+    // push the meta data into the planStore - this will persist 
+    // overrides if the meta data definition has them
+    ServiceData.setMetaData("i01", "InMoov");
+    
+    // pull it back out
+    MetaData data = ServiceData.getMetaData("i01", "InMoov");
+    log.warn("static meta data {}", data);
+
+    Plan plan = ServiceData.getPlan("i01", "InMoov");
+    log.warn("static plan {}", plan);
```
kwatters, line 57: Lets add some sort of an assert.. at a minimum that the plan isn't null.

src/main/java/org/myrobotlab/service/meta/abstracts/MetaData.java
```diff
@@ -123,23 +123,16 @@
 
   Integer workingLevel = null;
 
-  public MetaData() {
+  public MetaData(String name) {
```
kwatters, line 126: so, now metadata is specific to an instance of a service, instead of specific to a service type?

supertick: Yes - although MavenWrapper and IvyWrapper are consumers of MetaData and do not care what instance they are - as dependency management has no interest.

This revision of Peers and MetaData is vastly smaller and cleaner than the mess I put in initially (back in Manticore).
A very big bonus is the meta data generated with its peer definition is assigned to Service.serviceType (ya the name is vestigial but if I change it, it will break stuff) 
But this information gives the potential for dynamic buttons to be created to manage the lifecycles of the services' peer.
Less code more capability ... good times

src/main/java/org/myrobotlab/service/meta/abstracts/AbstractSpeechSynthesisMeta.java
supertick: fixed

src/test/java/org/myrobotlab/framework/ServiceLifeCycleTest.java
supertick: added a couple statements and noticed - logging level was too high

src/main/java/org/myrobotlab/framework/repo/ServiceData.java
```diff
@@ -516,4 +516,62 @@ public static void main(String[] args) {
...
+      }
+      planStore.put(peerKey, entry.getValue());
+     
+    }
+    
+    // breadth first recursion
+    for (ServiceReservation peer : peers.values()) {
+      // for all children do the same ..
+      // String peerKey = "something";
+      setMetaData(peer.actualName, peer.type, force);
```
kwatters: if a service has a peer that points back to the service, this will cause an infinite loop and ultimately a stack overflow exception.   I think we need to track all  peer.actualNames that have been processed in a set, and skip if if we've already visited that peer to ensure proper cycle/loop detection.

PR #759, opened 2020-07-11
worke-02

supertick:


src/main/resources/resource/Arduino/MrlComm/NewPing.cpp
```diff
@@ -345,11 +345,15 @@ ISR(TIMER2_COMPA_vect) {
 // ---------------------------------------------------------------------------
 
 unsigned int NewPing::convert_cm(unsigned int echoTime) {
+	// FIXED return raw time - http://myrobotlab.org/content/ultrasonicsensor
```
kwatters: If you touch this file, we need to up the mrlcomm version.

PR #760, opened 2020-07-12
servo sync

supertick:


src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -126,6 +126,11 @@ protected boolean processMove(Double newPos, boolean blocking, Long timeoutMs) {
      * </pre>
      *
      */
+    // FIXME - poor implementation - should addListener(sync)
+    // and use pub/sub :(
+    for(String syncServo: syncedServos) {
+      send(syncServo, "moveTo", newPos);
```
kwatters, line 132: There should be some test to make sure a servo isn't sync'd to itself.. otherwise, you'll get into a nasty loop here.

supertick: fixed

PR #761, opened 2020-07-15
pure webgui opencv update

supertick:


src/main/java/org/myrobotlab/service/OpenCV.java
```diff
@@ -1350,6 +1254,25 @@ synchronized public void pauseCapture() {
 
   }
 
+  /**
+   * conversion from buffered image to base64 encoded jpg
+   * 
+   * @param img
+   * @return
+   */
+  public String toBase64Jpg(BufferedImage img) {
```
kwatters, line 1263: this could be static.

PR #762, opened 2020-07-15
Blur detector

kwatters:
opencv... now with more blur detection!

### DIFF
--- a/src/main/java/org/myrobotlab/opencv/OpenCVData.java
+++ b/src/main/java/org/myrobotlab/opencv/OpenCVData.java
@@ -155,6 +155,8 @@ public class OpenCVData extends CvData {
   // for use with text detection / ocr filters.
   private ArrayList<DetectedText> detectedText;
   
+  private Double blurriness = null;
+  
   public OpenCVData() {
   }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/myrobotlab/opencv/OpenCVData.java
```diff
@@ -155,6 +155,8 @@ public static void main(String[] args) {
   // for use with text detection / ocr filters.
   private ArrayList<DetectedText> detectedText;
   
+  private Double blurriness = null;
+
```
supertick, line 159: This will go out on every opencvdata ?

src/main/java/org/myrobotlab/opencv/OpenCVFilterBlurDetector.java
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.Test;
+import org.myrobotlab.service.OpenCV;
+
+public class OpenCVFilterBlurDetector extends OpenCVFilter {
+
+  private static final long serialVersionUID = 1L;
+
+  // this threshold is only for the display label
+  float threshold = 100.0f;
+
```
supertick, line 31: member vars in general should probably be marked as protected - this will still get them to and from webgui dialogs

src/main/java/org/myrobotlab/opencv/OpenCVData.java
kwatters: I didn't really know where else to put it.

PR #766, opened 2020-07-26
serial setup fix

supertick:


kwatters:
So.. what do you want to do with this PR?  I'd really like to make sure that we don't ignore a driver issue with Linux if that is actually the issue...

kwatters:
You merged this?  Why?  I really don't want this merged...

PR #772, opened 2020-09-15
Agent removal

supertick:


src/main/resources/resource/Arduino/MrlComm/MrlUltrasonicSensor.cpp
```diff
@@ -33,7 +33,7 @@ void MrlUltrasonicSensor::update() {
 	if (!isRanging) {
 		return;
 	}
-	unsigned long distance = newping->ping_cm();
+	unsigned long distance = newping->ping();
```
kwatters, line 36: this should cause the mrlcomm version number to increment.

PR #775, opened 2020-09-30
improved webcam

supertick:


pom.xml
```diff
@@ -974,8 +984,8 @@
   </dependency>
   <dependency>
     <groupId>com.github.sarxos</groupId>
-    <artifactId>webcam-capture</artifactId>
-    <version>0.3.10</version>
+    <artifactId>webcam-capture-driver-v4l4j</artifactId>
+    <version>0.3.13-SNAPSHOT</version>
```
kwatters, line 988: Why are we using a snapshot build here?  is there something in 0.3.13-SNAPSHOT specifically that we need compared to 0.3.12 ?

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
```diff
@@ -389,7 +389,12 @@ synchronized public void install(String location, String[] serviceTypes) {
       Set<ServiceDependency> targetLibraries = getUnfulfilledDependencies(serviceTypes);
 
       if (targetLibraries.size() == 0) {
-        info("%s already installed", (Object[]) serviceTypes);
+        StringBuilder sb = new StringBuilder();
+        for (String type : serviceTypes) {
+          sb.append(type.substring(type.lastIndexOf(".") + 1));
```
kwatters: This logic is fragile as it could yield an out of bounds array problem.

src/main/java/org/myrobotlab/framework/repo/Repo.java
```diff
@@ -141,6 +141,7 @@ protected Repo() {
       remotes.add(new RemoteRepo("bintray", "https://jcenter.bintray.com", "the big kahuna"));
       remotes.add(new RemoteRepo("bintray2", "https://dl.bintray.com", "more big kahuna"));
       remotes.add(new RemoteRepo("myrobotlab", "http://repo.myrobotlab.org/artifactory/myrobotlab", "all other mrl deps"));
+      remotes.add(new RemoteRepo("sarxos", "http://oss.sonatype.org/content/repositories/snapshots", "for sarxos webcam"));
```
kwatters, line 144: we won't need this additional repo if we use 0.3.12 of sarxos

src/main/java/org/myrobotlab/service/Webcam.java
```diff
@@ -1,49 +1,225 @@
...
+    worker = new Thread(new VideoProcessor(), String.format("%s-video-processor", getName()));
+    worker.start();
+  }
+
+  public void capture(String name, String type, Integer fps, Integer width, Integer height, Double quality) {
+    if (name == null) {
+      capture(0, type, fps, width, height, quality);
+    } else {
+      List<com.github.sarxos.webcam.Webcam> cams = com.github.sarxos.webcam.Webcam.getWebcams();
+      for (int i = 0; i < cams.size(); ++i) {
```
kwatters, line 161: Why are we capturing from all webcams on the system here?   We should specify which cameras should be captured from.. not all available cameras on the system..

pom.xml
supertick: its the only version that works on all other platforms "and" raspi

src/main/java/org/myrobotlab/framework/repo/Repo.java
supertick: does "not" work with raspi

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
supertick: fixed

PR #776, opened 2020-09-30
Git service added

supertick:
- Git and Updater service added (previously part of Agent)
- Made RasPi more function - identifies board and queries pins using pi4j
- WebGui now does mDns where --id raspi will be accessible via http://raspi.local:8888 
- Get local services only to export in python LangUtils - remote services do not make sense
- Self contained websocket client class updated - used to test distributed in some use cases
- OpenCVFilterDL4J - fixed non serializable members - marked them transient
- Renamed OpenCVFilterNot to Invert with the idea Not makes more sense to more people
- YoloDetectOjbect - fixed non serializable members - marked them transient
- PinData toString prints pin not address
- small polish, updated some service mains, corrected typos etc


src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
```diff
@@ -389,7 +389,12 @@ synchronized public void install(String location, String[] serviceTypes) {
       Set<ServiceDependency> targetLibraries = getUnfulfilledDependencies(serviceTypes);
 
       if (targetLibraries.size() == 0) {
-        info("%s already installed", (Object[]) serviceTypes);
+        StringBuilder sb = new StringBuilder();
+        for (String type : serviceTypes) {
```
kwatters: There might be a merge problem here..  I thought this was fixed in the webcam PR.

supertick: fixed - merged develop

supertick: added semi-colon

PR #779, opened 2020-10-15
Various upgrades

kwatters:
this upgrades versions of javacv , tesseract , deeplearning4j , solr and junit.


pom.xml
```diff
@@ -65,1442 +65,1442 @@
...
+  </dependency>
+  <dependency>
+    <groupId>org.jmonkeyengine</groupId>
+    <artifactId>jme3-desktop</artifactId>
+    <version>3.3.2-stable</version>
+    <scope>provided</scope>
+  </dependency>
+  <dependency>
+    <groupId>org.jmonkeyengine</groupId>
+    <artifactId>jme3-lwjgl3</artifactId>
```
supertick: Your putting lwjgl3 back in ?  It will break oculus .. again...

pom.xml
```diff
@@ -65,1442 +65,1442 @@
...
+  <dependency>
+    <groupId>net.sf.opencsv</groupId>
+    <artifactId>opencsv</artifactId>
+    <version>2.3</version>
+    <scope>provided</scope>
+  </dependency>
+<!-- CsvConnector end -->
+
+<!-- Deeplearning4j begin -->
+  <dependency>
```
supertick, line 244: Do you know why the whitespace difference, was this generated

supertick (changes requested):
we should sync on deps for lwjgl

pom.xml
kwatters: merge problem.. will fix. good catch.

pom.xml
kwatters: Was the file generated 100% before?  This file is the result of running MavenWrapper.xml  no changes.

PR #782, opened 2020-11-01
Changed Authorize to Describe

supertick:
* Initial "authenticate" changed to "describe" 
    - authentication is handled on a connection basis 
    - authorization is on a per message basis
* Refactor cmd options - CmdOptions now abile to serialize back to a list of arguments

* CmdOptionsTest added

* HelloRequest renamed/refactored to DescribeQuery object

* AuthResponse renamed/refactored to DescribeResults object
* Removed unused Heartbeat object
* New Connection requirements id and uuid
* Client re-attach bug fixes
* RouteTable object created
* RouteTable refactor/bug-fixes


src/main/java/org/myrobotlab/framework/Platform.java
```diff
@@ -393,11 +393,11 @@ public boolean isX86() {
         // zf.close(); explodes on closing :(
       } else {
         // IDE - version ...
-        in = new FileInputStream("target/classes/META-INF/MANIFEST.MF");// Platform.class.getResource("target/classes/META-INF/MANIFEST.MF").openStream();
+        // in = new FileInputStream("target/classes/META-INF/MANIFEST.MF");// Platform.class.getResource("target/classes/META-INF/MANIFEST.MF").openStream();
+        in = new FileInputStream("target/classes/git.properties");// Platform.class.getResource("target/classes/META-INF/MANIFEST.MF").openStream();
```
kwatters, line 397: Where does this git.properties come from?

src/main/java/org/myrobotlab/client/Client.java
```diff
@@ -43,330 +46,487 @@
...
+    if ("describe".equals(msg.method)) {
+      serverHelloRequest = gson.fromJson(msg.data[1].toString(), DescribeQuery.class);
+      promptId = serverHelloRequest.id;
+      System.out.println("attaching to id " + promptId);
     }
 
-    synchronized public void stop() {
-      isRunning = false;
+    if (msg.data == null || msg.data[0] == null) {
+      System.out.println("null");
```
kwatters, line 249: we should probably clean  up stdout .. i suspect it's only here for debugging?

src/main/java/org/myrobotlab/net/Pinger.java
```diff
@@ -0,0 +1,65 @@
...
+  public void run() {
+    try {
+      if (InetAddress.getByName(ip).isReachable(timeout)) {
+        Host h = null;
+        if (hosts.containsKey(ip)) {
+          h = hosts.get(ip);
+        } else {
+          h = new Host();
+          h.ip = ip;
+          h.name = "new-host";
```
kwatters, line 36: we probably shouldn't hard code the host name to "new-host" here ?  if we assign it later when we know it, we could set it to a default value of unknown or something?

src/main/java/org/myrobotlab/framework/Platform.java
supertick: generated by the maven build

src/main/java/org/myrobotlab/client/Client.java
supertick: Client is not used by anything - its standalone, part of it was the interest of a single file client with no dependencies ...
So, no slf4j/log4j etc...
But, I've kindof waffled a bit, because it currently uses picocli :P

src/main/java/org/myrobotlab/net/Pinger.java
supertick: the data from the Pinger comes from ICMP - you don't get host name .. "new-host" is currently the default

PR #791, opened 2021-02-01
fixed python callbacks and added unit test - clock had race condition

supertick:


src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -51,17 +55,19 @@
    * 
    */
   public class InputQueueThread extends Thread {
-    private Python python;
+    transient protected Python python;
+    protected boolean isRunning = false;
```
kwatters: isRunning should be marked as volatile if you want it to be interrupted properly from a different thread.  Also, it probably should be renamed to "running" ...  "is" as a prefix, is usually reserved for method names like  isFoo() .. or isRunning() ..

PR #795, opened 2021-03-24
Grog mqtt broker

supertick:
MqttBroker

1. Created fully function embedded MqttBroker
2. WebGui for Mqtt and MqttBroker
3. Fixed getting message id
4. When SLF4J has multiple back ends - set log level fails - Fixed
5. Created Pojo MqttMsg
6. Created SslUtil class for tls - needed for mqtt connectivity to Amazon IoT Core
7. Complete refactor of Mqtt service
8. Made serviceData in Runtime transient - so that Runtime would be under the 128K msg size limit in Amazon
9. Changed Message.dataEncoding to field name to Message.encoding
10. Lowered Locales logging level to cut down on chatter
11. Added MqttBrokerTest and PythonTest to junit

More info can be found here - http://myrobotlab.org/content/mqtt


pom.xml
```diff
@@ -416,7 +416,19 @@
 
 <!-- InMoov2 begin -->
 <!-- skipping fr.inmoov inmoov2 fr.inmoov-inmoov2-null null version/latest -->
-<!-- InMoov2 end -->
+<!-- InMoov2 end -->
+
+<dependency>
```
kwatters: please re-generate the pom.xml to make sure it's still fully in sync with the metadata files.

.gitignore
```diff
@@ -78,3 +78,7 @@
 /libbulletjme.so
 /liblwjgl64.so
 /copyToRaspi.sh
+/lastRestart.py
```
kwatters, line 81: are these ignores required?  We should avoid ignores on things in the actual src/ folder to help keep our working copies cleaner..

pom.xml
supertick: doh .. I did regen it ... I forgot to commit it :(
fixed....

.gitignore
supertick: I agree .. the .gitignore deserves cleanup..
lastRestart.py probably deserves to be there - now that restarting mrl is a valid possibility, this is the file which attempt to preserve state between restarts ..
.factorypath is and vscode IDE file ... so probably shoudl be in there too ..
the other two 
certs should be removed ...
and the other one I see should probably be under /data so I'll fix

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -532,4 +549,63 @@ public static void main(String[] args) {
     }
   }
 
+  /**
+   * Creates a properly double encoded Json msg string.
```
kwatters, line 553: feels icky to have to double encode/decode anything.

src/main/java/org/myrobotlab/framework/Inbox.java
```diff
@@ -166,11 +166,16 @@ public Message getMsg() throws InterruptedException {
             // correct but this works for
             // null data now
             {
-              returnContainer = null;
+              returnContainer[0] = null;
             } else {
               returnContainer[0] = msg.data[0]; // transferring
               // return data !
             }
+            /*
```
kwatters: please delete this commented out code if it's not used.

src/main/java/org/myrobotlab/mqtt/MqttMsg.java
```diff
@@ -0,0 +1,147 @@
+package org.myrobotlab.mqtt;
+
+public class MqttMsg {
```
kwatters, line 5: for pojo's it's a good idea to have a hashCode  and equals methods.

src/main/java/org/myrobotlab/net/SslUtil.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.bouncycastle.openssl.PEMDecryptorProvider;
+import org.bouncycastle.openssl.PEMEncryptedKeyPair;
+import org.bouncycastle.openssl.PEMKeyPair;
+import org.bouncycastle.openssl.PEMParser;
+import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter;
+import org.bouncycastle.openssl.jcajce.JcePEMDecryptorProviderBuilder;
+
+public class SslUtil {
```
kwatters, line 28: please add a javadoc description to this class.

kwatters: one thing to keep in mind is that bouncycastle libraries have export restrictions which means that this code might not legally be able to be used in china or russia.  (and other countries.)  I'm not sure if that's changed in more recent times.

src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -474,19 +475,17 @@ public boolean exec(String code, boolean blocking) {
...
-
     } catch (PyException pe) {
       // something specific with a python error
       error(pe.toString());
       invoke("publishStdError", pe.toString());
     } catch (Exception e) {
       error(e);
+    } finally {
+      if (blocking) {
+        invoke("finishedExecutingScript");
```
kwatters, line 487: nice.. maybe we can actually have proper gesture started / gesture stopped support in inmoov now.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -2512,6 +2512,10 @@ public DescribeResults describe(String uuidX, DescribeQuery hello) {
    */
   public void onDescribe(DescribeResults results) {
     List<Registration> reservations = results.getReservations();
+    if (getId().equals("c1"))
```
kwatters, line 2515: What does "c1" represent here? Is this debug code that should be removed?

src/main/java/org/myrobotlab/service/meta/DocumentPipelineMeta.java
```diff
@@ -27,7 +27,7 @@ public DocumentPipelineMeta(String name) {
     addDependency("org.sweble.wikitext", "swc-engine", "3.1.7");
     addDependency("org.sweble.wom3", "sweble-wom3-core", "3.1.7");
 
-    addDependency("com.thoughtworks.xstream", "xstream", "1.4.13-java7");
+    addDependency("com.thoughtworks.xstream", "xstream", "1.4.15");
```
kwatters, line 30: thanks dependabot...

src/main/java/org/myrobotlab/codec/CodecUtils.java
supertick: to support "distributed" it necessary at the moment ...
but I didn't want to omni bus with pulling the header into a separate sub object

src/main/java/org/myrobotlab/service/Mqtt.java
```diff
@@ -14,231 +18,390 @@
...
 
-  boolean autoSubscribe = true;
-  String userName = null;
-  char[] password = null;
-  boolean autoReconnect = true;
+  protected String username = null;
 
   public Mqtt(String n, String id) {
     super(n, id);
     try {
```
kwatters, line 183: could this stuff be done in a startService instead?  I'd like to move away from initialization code in the constructor.. and moving that code into startService..   this keeps the instantiation of objects lightweight.

src/main/java/org/myrobotlab/service/Mqtt.java
```diff
@@ -14,231 +18,390 @@
...
-            return;
-          }
-        }
+      if (sslCertFilePath != null && sslKeyFilePath != null) {
+        SSLSocketFactory socketFactory = SslUtil.getSocketFactory(sslCaFilePath, sslCertFilePath, sslKeyFilePath, sslPassword);
+        conOpt.setSocketFactory(socketFactory);
+      }
+
+      // default is 10 - which drops messages on a fast local network :(
+      conOpt.setMaxInflight(1000);
```
kwatters, line 273: maybe make this configurable?  if it's important.

src/main/java/org/myrobotlab/framework/Inbox.java
supertick: fixed

src/main/java/org/myrobotlab/service/Mqtt.java
```diff
@@ -14,231 +18,390 @@
...
+        if (!autoReconnect) {
+          broadcastState();
+          return;
+        }
       }
+
+      broadcastState();
+
+    } catch (Exception e) {
+      Service.sleep(1500);
```
kwatters: why sleep here at all? the exception is already blown.

src/main/java/org/myrobotlab/service/Mqtt.java
```diff
@@ -14,231 +18,390 @@
...
+    connecting = false;
+  }
+
+  public void connect(String url) {
+    this.url = url;
+    connect();
+  }
+
+  public void connect(String url, String clientId, String username, String password) {
+    this.url = url;
```
kwatters: This method signature seems to be max complexity.  I'd prefer to see connect(url)  connect() , instead call this method.. with the proper parameters instead of these methods calling connect with no args.

src/main/java/org/myrobotlab/service/MqttBroker.java
```diff
@@ -0,0 +1,628 @@
...
+ *
+ */
+public class MqttBroker extends Service implements InterceptHandler, Gateway {
+
+  public final static Logger log = LoggerFactory.getLogger(MqttBroker.class);
+
+  private static final long serialVersionUID = 1L;
+
+  private static String bytesToHex(byte[] hash) {
+    StringBuilder hexString = new StringBuilder(2 * hash.length);
```
kwatters, line 69: don't re-implement this here. use bytesToHex on the StringUtils class in MRL..

src/main/java/org/myrobotlab/service/MqttBroker.java
```diff
@@ -0,0 +1,628 @@
...
+      String hex = Integer.toHexString(0xff & hash[i]);
+      if (hex.length() == 1) {
+        hexString.append('0');
+      }
+      hexString.append(hex);
+    }
+    return hexString.toString();
+  }
+
+  private static final MqttMsg toMqttMsg(InterceptMessage msg) {
```
kwatters: why would you want it to be both private and static?  only used by the current class in a static context?

src/main/java/org/myrobotlab/service/MqttBroker.java
```diff
@@ -0,0 +1,628 @@
...
+      retMsg.setClientID(m.getClientID());
+      retMsg.setUsername(m.getUsername());
+      retMsg.setTopicName(m.getTopicFilter());
+    }
+    return retMsg;
+  }
+
+  protected String address = "0.0.0.0";
+
+  protected Boolean allow_anonymous = false;
```
kwatters: why use objects when primitive would be better and less ambigious.  boolean is true or false.. Boolean is true,false or null.  Seems like opting for primitives here will avoid a bunch of null pointer exeptions

src/main/java/org/myrobotlab/service/MqttBroker.java
```diff
@@ -0,0 +1,628 @@
...
+
+  /**
+   * determines if the broker will attempt to process any messages published on
+   * the api topics .. typically api/service & api/messages
+   */
+  boolean processApiMessages = true;
+
+  protected boolean listening = false;
+
+  transient final Server mqttBroker = new Server();
```
kwatters, line 140: does this need to be created here.. or perhaps it can be created in the startService method?  seems like that's the right place. but there may be other reasons to do it statically up when the class is loaded.

src/main/java/org/myrobotlab/service/MqttBroker.java
```diff
@@ -0,0 +1,628 @@
...
+    return Runtime.getInstance().isLocal(msg);
+  }
+
+  public void listen() {
+    listen(null, null, null, null, null, null);
+  }
+
+  public void listen(int mqttPort) {
+    this.mqttPort = mqttPort;
+    listen();
```
kwatters: you should call the max complexity method.  not listen() ...    that's the least complexity method!
listen(null,  inMqttPort, null, null, null, null)

src/main/java/org/myrobotlab/service/MqttBroker.java
```diff
@@ -0,0 +1,628 @@
...
+
+  public MqttMsg publishUnsubscribe(InterceptUnsubscribeMessage msg) {
+    return toMqttMsg(msg);
+  }
+
+  private void saveIdentities() {
+    try {
+      if (username != null && username.length() > 0) {
+        FileOutputStream fos = new FileOutputStream(passwordFilePath);
+        MessageDigest digest = MessageDigest.getInstance("SHA-256");
```
kwatters: we should pull the algorithm into a constant at the top of the class.. or make it configurable with a default as provided here.

src/main/java/org/myrobotlab/mqtt/MqttMsg.java
supertick: fixed

src/main/java/org/myrobotlab/net/SslUtil.java
supertick: Ahahahah .. better tell Amazon to shut down their Bejing IoT Core ...
https://docs.amazonaws.cn/en_us/aws/latest/userguide/iot.html#feature-regions

Fixed

src/main/java/org/myrobotlab/codec/CodecUtils.java
supertick: won't fix - this pr

src/main/java/org/myrobotlab/service/Python.java
supertick: mebbe .. or at least a step closer - still needs unique mappings

src/main/java/org/myrobotlab/service/Runtime.java
supertick: yeah - should be removed - but not ready yet ...
I want **_describe_** to work similarly to kubernetes describe - will take some more effort

src/main/java/org/myrobotlab/service/Mqtt.java
supertick: hmmm .. dunno, these are primitives btw .... not really very "heavy weight"

src/main/java/org/myrobotlab/service/Mqtt.java
supertick: ya

src/main/java/org/myrobotlab/service/Mqtt.java
supertick: heh ... I don't know how many times I've seen reconnect logic without a sleep, and the consequence usually is a fast spin that pegs the cpu when the host is unreachable, rather than retrying at a meaningful interval (or progressive backoff) the client pegs the cpu the whole process or whole computer become bork'd...

But this is an artifact from my first implementation, I later discovered the paho client itself implemented reconnect logic - so ya ... mebbe can be removed ... I'll look

src/main/java/org/myrobotlab/service/Mqtt.java
supertick: for some reason I seem to do it backwards .... will fix

src/main/java/org/myrobotlab/service/MqttBroker.java
supertick: will fix

src/main/java/org/myrobotlab/service/MqttBroker.java
supertick: was hoping mqtt client and mqtt broker could share the logic of converting a mqtt msg into a pojo ...
they can't - it will be private

src/main/java/org/myrobotlab/service/MqttBroker.java
supertick: moquette embedded mqtt server modeled its goofy config form mosquittos goofy config which had this boolean vs having username/pwd non empty or multiple credentials - which should have been the source of the true/false ..

not used - removing ....

src/main/java/org/myrobotlab/service/MqttBroker.java
supertick: backwards ...

will fix

src/main/java/org/myrobotlab/service/MqttBroker.java
supertick: that's icing .. but sure ... - the challenge was to get to work with a cloud provider

will fix

src/main/java/org/myrobotlab/service/MqttBroker.java
supertick: your probably right ...

will analyze and fix

supertick: it can't be final and get initialized in startService
The moquette broker is one of the few servers that can actually stop listening then start again successfully - so was going to utilize it - but I believe also the more defensive and safer way is to shut it down completely then new up a new one ...

Will investigate.

src/main/java/org/myrobotlab/service/Mqtt.java
supertick: I don't think its worth exposing at this point ...

src/main/java/org/myrobotlab/service/Mqtt.java
supertick: I removed it .. I'm going to rely on the paho reconnect and see how that goes

src/main/java/org/myrobotlab/service/Mqtt.java
supertick: fixed

src/main/java/org/myrobotlab/service/MqttBroker.java
supertick: its a different implementation - it was originally lifted from the moquette broker which lifted it from the mosquitto broker ...
I currently have devices depending on it..
won't fix ..
at least not this pr

src/main/java/org/myrobotlab/service/MqttBroker.java
supertick: I'm not going to change it ...
I might in the future make a new instance when listening and delete it when not listening ... but there should be some experimentation done around it's "internalPublish" which may not need to be listening to use ... dunno, don't want to mess with it until I do.

src/main/java/org/myrobotlab/service/MqttBroker.java
supertick: fixed

src/main/java/org/myrobotlab/service/MqttBroker.java
supertick: fixed as constant - not configurable

src/main/java/org/myrobotlab/service/Mqtt.java
```diff
@@ -249,56 +421,178 @@ public String getUrl() {
...
+          /*
+           * zzz String rxTopic = String.format("mrl/gw/%s/rx<-%s", getId(),
+           * remoteId); subscribe(rxTopic); info("subscribed to topic %s",
+           * rxTopic);
+           */
+          // something is listening - i need to let them know I'm alive -
+          // broadcast to onConnect
+          broadcastConnect();
+          // wait for them to get setup with appropriate subscription
+          sleep(1500);
```
kwatters, line 524: Is this sleep required?

PR #801, opened 2021-04-01
Grog servomixer UI fixes

supertick:


src/main/java/org/myrobotlab/service/ServoMixer.java
```diff
@@ -41,6 +42,11 @@ public ServoMixer(String n, String id) {
     for (String sc : all) {
       allServos.add(Runtime.getFullName(sc));
     }
+    
+    File poseDirectory = new File(posesDirectory);
```
kwatters: shouldn't this be done in startService, not in the constructor?

supertick: I don't think so ... 
Some services have a concept of "default" and we should try to provide this as best as possible ..
I found the service exploding previously, because a call to get the list of files in the "default" directory was null
I don't believe "everything" should be initialized in start - it seems like rational defaults should still exist in the constructor.  Anything related to "producing data" I could see should be in "start"

kwatters: This means that you have the side effect of creating a directory on the filesystem, and you have no control over changing that default location.  Initialization stuff should happen in a startService method.  The creating of the directory shouldn't happen in the constructor of this object.

supertick:
closed - made new pr - https://github.com/MyRobotLab/myrobotlab/pull/805

PR #813, opened 2021-04-13
make language pack public

kwatters:
so the inmoov scripts can get to it...


supertick (approved):
Yuck ... should be protected as all InMoov2 members will be

PR #816, opened 2021-04-25
ServoMixer Updates

supertick:
ServoMixer now can create poses and sequences.
* removed jme menu (clogging space) and made window resizable (Yay)
* added Pose, PoseSequence and Sequence Pojos
* gave InMoov2 the ability to start servos without controllers
* stop Servo from broadcasting itself on every single incremental move (blech)
* big refactor ServoMixer - enhance with sequence building and sequence files and other niceties 
* changed communication uniqueness on js clients to webgui
* creation of a new service gets inspected possibly refresh controllers 
* lots of webgui work

### DIFF
--- a/src/main/java/org/myrobotlab/kinematics/PoseSequence.java
+++ b/src/main/java/org/myrobotlab/kinematics/PoseSequence.java
@@ -1,0 +1,27 @@
+package org.myrobotlab.kinematics;
+
+public class PoseSequence {
+  
+    public int id;
+
+    /**
+     * name of pose
+     */
+    public String name;
+    
+    /**
+     * number of ms to wait before starting this pose
+     */
+    public Long waitTimeMs;
+    
+    public PoseSequence() {      
+    }
+    
+    public String toString() {
+      if (waitTimeMs == null) {
+        return name;
+      } else {
+        return String.format("%s %d ms");
+      }
+    }
+}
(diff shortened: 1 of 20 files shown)


src/main/java/org/myrobotlab/kinematics/PoseSequence.java
```diff
@@ -0,0 +1,27 @@
+package org.myrobotlab.kinematics;
+
+public class PoseSequence {
```
kwatters, line 3: Rename this class to something easier to understand  like a pose step.. or just a step  (as a base class)..  ..   "sequence" makes me think this is a list of steps. but it's not, it's just one of them.

src/main/java/org/myrobotlab/kinematics/Sequence.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.List;
+
+import org.myrobotlab.codec.CodecUtils;
+import org.myrobotlab.io.FileIO;
+import org.myrobotlab.logging.LoggerFactory;
+import org.myrobotlab.service.InMoov2;
+import org.slf4j.Logger;
+
+/** represent a set of servo positions at a given point in time */
+public class Sequence {
```
kwatters, line 14: this sequence, really should be renamed to gesture, and it should be represented by a list of steps..
steps can be pose steps or speaking steps..  (perhaps others)

PR #817, opened 2021-04-26
added source to range - without breaking pre-existing topic

supertick:


supertick:
greg
added source to range - without breaking pre-existing topic

src/main/java/org/myrobotlab/service/UltrasonicSensor.java
```diff
@@ -74,7 +75,8 @@ public UltrasonicSensor(String n, String id) {
   // change units, sample rate, etc
   // FIXME - NOT SERVICE .. possibly name or interface but not service
   public void addRangeListener(Service service) {
-    addListener("publishRange", service.getName(), "onRange");
+    addListener(""
```
kwatters: What is "ge"  ?

src/main/java/org/myrobotlab/service/data/AbstractData.java
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * An attempt to try to normalize some of the common fields of these pojos
+ * 
+ * @author grog
+ *
+ */
+public abstract class AbstractData {
+
+  public String source;
```
kwatters, line 11: It's a bit overkill to create an abstract base class for this...  don't you think?

src/main/java/org/myrobotlab/service/UltrasonicSensor.java
supertick: what the hell ? ... pre-coffee mess ... thanks for the catch

src/main/java/org/myrobotlab/service/data/AbstractData.java
supertick: I don't think so - as soon as anyone gets more than one of the same type of sensor - regardless of the type, they will want to know the origin of the sensor data, and source/name can sort it out

PR #823, opened 2021-05-04
neopixel solution

supertick:


src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -45,7 +45,7 @@
   static boolean RobotCanMoveRandom = true;
   private static final long serialVersionUID = 1L;
 
-  public Arduino neopixelArduino = null;
+  public Arduino neopixelArduinox = null;
```
kwatters: If this isn't used anymore. we should remove neopixelArduinox

supertick:
this was put in another pr

PR #828, opened 2021-05-07
auto disable cleanup - opencv native canvas - cleanup of jcenter and bintray

supertick:
- autodisable in inmoov2
- native opencv canvasframe viewer by default
- jcenter and bintray removed from runtime repo dep management
- quicker shutdown - no need to save megs of json that is not used

### DIFF
--- a/src/main/java/org/myrobotlab/service/OpenCV.java
+++ b/src/main/java/org/myrobotlab/service/OpenCV.java
@@ -52,7 +52,6 @@ import java.util.concurrent.TimeUnit;
 
 import javax.imageio.ImageIO;
 import javax.imageio.stream.MemoryCacheImageOutputStream;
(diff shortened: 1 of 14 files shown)


supertick:
- autodisable in inmoov2
- native opencv2 viewer by default
- jcenter and bintray removed from runtime dep management

src/main/java/org/myrobotlab/service/OpenCV.java
```diff
@@ -52,8 +52,10 @@
 
 import javax.imageio.ImageIO;
 import javax.imageio.stream.MemoryCacheImageOutputStream;
+import javax.swing.WindowConstants;
```
kwatters: Why is there a swing import here?

PR #839, opened 2021-05-16
right and left ultrasonic

hairygael:


supertick:
Why add stuff to InMoov ? vs InMoov2 ?

hairygael:
Good question.
Because users a still mostly using InMoov with Nixie. 
And I noticed some users were wondering on the forum, why there is two Ultrasonic in the InMoov config files while it doesn't work with InMoov.java
Anyway I had already set both ultrasonic in InMoov2 last year. Not sure it works though.

PR #844, opened 2021-05-24
adding correct attach and supporting updates

supertick:
Implemented attachTextPublish correctly
at the base it should be a pub/sub subscription with a string value, that is all that is needed.
A "service reference" should not be required.  This gives the benefit during runtime where one type of speech synthesis services named 'mouth' can be removed from ProgramAB and a new one brought up without any adverse affects.

It also means ProgramAB or other services which support attaching to Text Publishers can be pre-configured to publish to a named destination without NPEs

### DIFF
--- a/src/main/java/org/myrobotlab/i2c/I2CBus.java
+++ b/src/main/java/org/myrobotlab/i2c/I2CBus.java
@@ -195,4 +195,9 @@ public class I2CBus implements Attachable, I2CBusControl {
 
   }
 
+  @Override
+  public Set<String> getAttached(String publishingPoint) {
+    return getAttached();
+  }
+
 }
--- a/src/main/java/org/myrobotlab/sensor/TimeEncoder.java
+++ b/src/main/java/org/myrobotlab/sensor/TimeEncoder.java
@@ -455,4 +455,10 @@ public class TimeEncoder implements Runnable, EncoderControl {
     // TODO Auto-generated method stub
 
   }
+
+  @Override
+  public Set<String> getAttached(String publishingPoint) {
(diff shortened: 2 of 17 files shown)


src/main/java/org/myrobotlab/i2c/I2CBus.java
```diff
@@ -195,4 +195,9 @@ public void removeListener(String localTopic, String otherService) {
 
   }
 
+  @Override
```
kwatters, line 198: Why override this here?  shouldn't we be using the impl from the base Service class?

src/main/java/org/myrobotlab/sensor/TimeEncoder.java
```diff
@@ -455,4 +455,10 @@ public void removeListener(String localTopic, String otherService) {
     // TODO Auto-generated method stub
 
   }
+
```
kwatters, line 458: Same, Why override this here?  shouldn't we be using the impl from the base Service class?

src/main/java/org/myrobotlab/i2c/I2CBus.java
supertick: heh .. you missed that this does not extend from Service .. ie its not a real service, nor is TimeEncoder ...
both should be promoted to real service class (TimeEncoder should handle multiple requests for time encoding, similar to Pid) .. but not related to this pr

src/main/java/org/myrobotlab/sensor/TimeEncoder.java
supertick: if it was a Service

PR #845, opened 2021-05-25
Grog py generation

supertick:


src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1421,6 +1427,40 @@ public Integer publishShutdown(Integer seconds) {
     return seconds;
   }
 
+  public List<String> publishConfigList() {
+    configList = new ArrayList<>();
+    File configDir = new File("data/config");
```
kwatters: This path should not be hard coded here.  it should reference a variable, be pulled out as a member and javadoc added.   we should be able to configure where the configs live.

supertick: This was intentional - I'd say a very good many bugs have to do with pathing errors.  Especially during runtime.
So this is a "convention" ... the convention is ... config is managed in this one directory ..
you can add stuff to it, you can change it there, you can delete stuff - but i don't want floating variable pointing to all different places at the moment

kwatters: then pull it out as a static constant and put a javadoc on it.

PR #847, opened 2021-05-31
small python fix

supertick:
if python services starts after other services, it should run the same initialization script on all pre-existing services.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,11 +112,13 @@
           <url>https://oss.sonatype.org/content/groups/public</url>
         </repository>
         <!-- for j3d core, utils and vector  -->
+        <!--
         <repository>
           <id>tudelft</id>
           <name>tudelft</name>
           <url>http://simulation.tudelft.nl/maven</url>
         </repository>
+        -->
         <!-- swinggui mxgraph  -->
         <repository>
           <id>alfresco</id>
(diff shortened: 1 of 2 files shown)


src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -732,11 +732,14 @@ public void setLocalScriptDir(String path) {
...
-  /*
-   * no longer needed
-   * 
-   * @Override public void startService() { super.startService(); }
-   */
+  @Override
+  public void startService() {
+    super.startService();
+    Map<String, ServiceInterface> services = Runtime.getLocalServices();
+    for (ServiceInterface s : services.values()) {
```
kwatters, line 739: this should probably be synchronized to make sure there's no race conditions.  a service could be started while in this loop and it would be missed.

pom.xml
```diff
@@ -112,11 +112,13 @@
           <url>https://oss.sonatype.org/content/groups/public</url>
         </repository>
         <!-- for j3d core, utils and vector  -->
+        <!--
```
kwatters, line 115: if we don't need it.. just remove it.. any reason to keep it around & commented out?

supertick: Notes say for j3d core utils and vector ...  vector lib (as well as tools lib) has always been a pain, Oracle doesn't make it available in any public repo and neither have become fully open source ... the tuft repo had it ...
The error from tufts was a PIX TLS chain is broken, we still need the libs.  I commented out - if any of the nodes purge their .m2 we'll quickly find all the missing ones we probably need to put into mrl's repo

supertick: In short - dunno what happened with tufts - but we certainly need vector.lib and others - so i commented it out for a reference.

PR #848, opened 2021-05-31
update pi4j and xtream removed tuft repo

supertick:


pom.xml
```diff
@@ -1186,13 +1181,13 @@
   <dependency>
     <groupId>com.pi4j</groupId>
     <artifactId>pi4j-core</artifactId>
-    <version>1.2</version>
+    <version>1.4</version>
```
kwatters, line 1184: I don't see the RasPiMeta update...  this file must not have been generated...

pom.xml
```diff
@@ -337,7 +332,7 @@
   <dependency>
     <groupId>com.thoughtworks.xstream</groupId>
     <artifactId>xstream</artifactId>
-    <version>1.4.16</version>
+    <version>1.4.17</version>
```
kwatters, line 335: DocumentPipelineMeta needs to be updated also.

PR #856, opened 2021-06-09
servo.setMaxSpeed(x) restrictive value for servo speed

supertick:


src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -212,8 +212,9 @@ public void enableAutoDisable(boolean value) {
   }
 
   @Deprecated
-  public void setMaxVelocity(Double velocity) {
+  public void setMaxVelocity(Double speed) {
     log.warn("SetMaxVelocity does nothing and is deprecated. please update your python scripts, and use fullSpeed() instead");
```
kwatters: this log message is no longer correct.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -821,24 +823,36 @@ public void setRest(double rest) {
...
+    if (degreesPerSecond == null) {
+      log.info("disabling speed control");
+    }    
     broadcast("publishServoSetSpeed", this);
     broadcastState();
   }
+  
+  @Override
+  public void setMaxSpeed(Double degreesPerSecond) {
+	 if (degreesPerSecond != null && (speed == null || speed > degreesPerSecond)){
```
kwatters, line 849: This method also sets the current speed? That seems wrong to me... If I say setMaxSpeed(100)  I expect the servo will never move faster than that.  I wouldn't expect that the servo would start moving at the max speed unless I explicitly called setSpeed on it..

supertick: Whether its done here (which I think is cleaner) ..
or if its set in processMove - if you setMaxSpeed(100) and you have "unbounded" speed currently, what is your expectation?
"Unbounded" is faster than 100
so speed should be 100 in that case ...
Another use case .. what if the speed is currently 200 .. and you setMaxSpeed(100) -  Would be your expectation be that its left at 200 ?

PR #857, opened 2021-06-09
worky !

supertick:
solid fix std in/out with stream gobblers ... additional bonus bug of kill stray orphans

### DIFF
--- a/src/main/java/org/myrobotlab/io/StreamGobbler.java
+++ b/src/main/java/org/myrobotlab/io/StreamGobbler.java
@@ -4,28 +4,29 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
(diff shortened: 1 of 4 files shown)


src/main/java/org/myrobotlab/io/StreamGobbler.java
```diff
@@ -4,28 +4,29 @@
...
-  boolean relay = true;
+  protected transient InputStream processOut;
+  protected transient OutputStream processIn;
+  protected String name;
 
-  public StreamGobbler(String name, InputStream is) {
+  public StreamGobbler(String name, InputStream processOut, OutputStream processIn) {
     super(name);
-    this.in = is;
+    this.processOut = processOut;
```
kwatters, line 16: What about stderr?  This will block if any of our 3rd parties write to stderr.  We don't want to bring back the mysterious dl4j locking up mrl again.

PR #863, opened 2021-06-12
openblas bump

supertick:


src/main/java/org/myrobotlab/service/meta/Deeplearning4jMeta.java
```diff
@@ -26,7 +26,7 @@ public Deeplearning4jMeta(String name) {
 
     // Force javacpp 1.5.3 to resolve conflict between dl4j and javacv
     addDependency("org.bytedeco", "javacpp", "1.5.5");
-    addDependency("org.bytedeco", "openblas", "0.3.10-" + "1.5.4");
+    addDependency("org.bytedeco", "openblas", "0.3.13-" + "1.5.5");
```
kwatters, line 29: we probably no longer need this explicitly here.. but it doesn't hurt for the time being.

PR #865, opened 2021-06-13
attempt to create InMoov2Eyelids

hairygael:
Hello Captains,
This is probably not going to be of your taste, but it's an attempt.
It's mainly a copy paste with name changes, I wouldn't know how to create the service properly anyway, so feel free to fix what you think doesn't fit for Nixie.


supertick:
This can now be handled with config - closing this pr

PR #876, opened 2021-06-20
Config

supertick:


src/main/java/org/myrobotlab/service/config/ServoConfig.java
```diff
@@ -0,0 +1,22 @@
+package org.myrobotlab.service.config;
+
+public class ServoConfig extends Config {
+
+  public Boolean autoDisable;
+  public String controller;
+  public Double currentOutputPos;
```
kwatters: we can specify the current input postion. .we should not put output current position in the config.

src/main/java/org/myrobotlab/service/config/ServoConfig.java
```diff
@@ -0,0 +1,22 @@
...
+  public Boolean isBlocking;
+  
+  public String pin;
+  public Double rest;
+  public Double speed;
+
+  public Double sweepMax;
+  public Double sweepMin;
+  
+  public Double targetPos;
```
kwatters: questionable if we care about targetPos

src/main/java/org/myrobotlab/service/config/ServoConfig.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class ServoConfig extends Config {
+
+  public Boolean autoDisable;
+  public String controller;
+  public Double currentOutputPos;
+  public Boolean enabled;
+  public Boolean idleDisabled;  
+  public Integer idleTimeout;
+  public Boolean isBlocking;
```
kwatters: isn't this a detail of a movement?  we shouldn't have  isBlocking in the config.

supertick: agree

src/main/java/org/myrobotlab/service/config/ServoConfig.java
supertick: It works out to be the value of "last input" ...
not sure - I'm going to take it out, if a "real encoder" existed it should be taking the value when starting there

PR #878, opened 2021-06-24
Grog config yaml

supertick:
New fixes:
Saving Locale fix - now all will be null except runtime, unless explicitly set
OpenCVData List vs ArrayList cleanup
OpenCVFilterFaceDetect needs constructor to be saved in yaml
OpenCVFilterFaceDetectDNN did not populate bounding boxes in opencvdata
Adafruit16CServoDriver - big cleanup on attach
Small Log service fix - explicit cast exception - fixed
OpenCV fix stopping service should dispose of CanvasFrame
RasPi attach fixed
Adafruit16CServoDriverConfig now yaml configurable bus and address

Runtime now runs through all defined service life cycles in order
create
load
start
attach
tricky part was using the config to do the attach, but after all services
were started - attaching "before" starting usually gives undesired results

load() now returns a ServiceConfig reference if successful
and can throw IOException

removed device list from ArduinoConfig - processed by attach list

Attach through config - attaching config is based on exported names from the service function getAttached() when saved. So detailed route information and callbacks are not saved or processed.  This is probably better than saving/loading raw routes.

also removed all the auto runtime subscriptions for auto service lifecycle events from other services ...
e.g.
onCreated(name)
onRegistered(name)
onStarted(name)
onReleased(name)


src/main/java/org/myrobotlab/service/config/ServoConfig.java
```diff
@@ -0,0 +1,26 @@
+package org.myrobotlab.service.config;
+
+public class ServoConfig extends Config {
```
kwatters: why not just extend ServiceConfig ?

src/main/java/org/myrobotlab/service/config/Config.java
```diff
@@ -0,0 +1,5 @@
+package org.myrobotlab.service.config;
```
kwatters: this class should be deleted

src/main/java/org/myrobotlab/service/config/ArduinoConfig.java
```diff
@@ -0,0 +1,10 @@
+package org.myrobotlab.service.config;
+
+import java.util.Set;
+
+public class ArduinoConfig extends Config {
```
kwatters: extend ServiceConfig instead

src/main/java/org/myrobotlab/service/config/ServiceConfig.java
```diff
@@ -0,0 +1,18 @@
...
+  public String type;   
+  public Config config;
+  
+  public ServiceConfig() {    
+  }
+  
+  public ServiceConfig(String name, String type, Config config) {
+    this.name = name;
+    this.type = type;
+    this.config = config;
```
kwatters: remove "config" as a member here.. subclasses should implement service specific configs.

src/main/java/org/myrobotlab/service/config/ServoConfig.java
supertick: maybe ... depends on which structure we like

src/main/java/org/myrobotlab/service/config/ArduinoConfig.java
supertick: mebbe

src/main/java/org/myrobotlab/service/config/ServoConfig.java
kwatters: the subclass can/should have full freedom to implement the config that fits that services needs.

supertick: eh?   it does ...

PR #879, opened 2021-07-03
Grog localspeech 2

supertick:
- massive refactor for LocalSpeech
- LocalSpeech now works with templates
- packaged tts.exe with resources
- remove MimicSpeech


src/main/java/org/myrobotlab/service/meta/LocalSpeechMeta.java
```diff
@@ -20,7 +20,8 @@ public LocalSpeechMeta(String name) {
     addDescription("Local OS text to speech ( tts.exe / say etc ... )");
     setAvailable(true);
     addCategory("speech");
-    addDependency("com.microsoft", "tts", "1.1", "zip");
+    // addDependency("com.microsoft", "tts", "1.1", "zip");
```
kwatters, line 23: how do the binaries get installed?

src/main/java/org/myrobotlab/service/LocalSpeech.java
```diff
@@ -49,78 +46,197 @@
...
+  protected String ttsPath = getResourceDir() + fs + "tts" + fs + "tts.exe";
+  protected String mimicPath = getResourceDir() + fs + "mimic" + fs + "mimic.exe";
   protected String ttsCommand = null;
+  protected String filterChars = "\"\'";
+  protected boolean removeExt = false;
+  protected boolean ttsHack = false;
 
   public LocalSpeech(String n, String id) {
     super(n, id);
+    // setup the default tts per os
```
kwatters, line 62: please avoid code in the constructor.. this is pretty light weight, but i feel it's a bad practice.  stuff like this should be in the startService method instead, not in the constructor

src/main/java/org/myrobotlab/service/meta/LocalSpeechMeta.java
supertick: tts.exe is a tiny executable to access ms SAPI system - which is where the implementation is..
I bundled tts.exe in the resource of LocalSpeech

supertick: mimic sounds like crap on windows - is there any real reason to continue with it on ms os ?

PR #884, opened 2021-07-12
Grog neopixel2

supertick:
## Created new NeoPixel2
- uses Adafruit neopixel library - 1 .h and 1.cpp
- has a more interactive webgui interface
- indexing begins with 0 vs 1
- 3 new mrlcomm functions
- brightness control
- onboard animation in addition to functional animations and Java land animations
- clean attach/detach - should be followed by other services - supports specific and general attach/detach
-

## Cli 
Fixed Clit codec - found issue when trying the following command
/neopixel/setSpeed/20
Type error - so cli didn't support basic primitive types - now it does


src/main/java/org/myrobotlab/service/interfaces/NeoPixel2Control.java
```diff
@@ -0,0 +1,75 @@
...
+ * 
+ * Enjoy !
+ * 
+ * */
+
+package org.myrobotlab.service.interfaces;
+
+import org.myrobotlab.framework.interfaces.NameProvider;
+
+public interface NeoPixel2Control extends NameProvider {
```
kwatters: Why a new interface?

kwatters:
if we are updating neopixel.. let's replace the existing one ..  if neopixel2 is a superset of functionality, let's just replace the old one instead of keeping the tech debt around.

src/main/java/org/myrobotlab/service/interfaces/NeoPixel2Control.java
supertick: because its different than NeoPixelControl..

it has clear() which NeoPixelControl didn't support
Additionally it has fully explicit typed values for 
public void attachNeoPixel2Controller(NeoPixel2Controller controller);
public void detachNeoPixel2Controller(NeoPixel2Controller controller);

and it does not have any "helper" attach methods which have additional parameters.

kwatters:
This fails upload on an arduino Uno complaining that the sketch is too big.  we will need to trim some stuff down if we want to continue supporting the Uno... 



```
Sketch uses 34,230 bytes (106%) of program storage space. Maximum is 32,256 bytes.
Global variables use 1,058 bytes (51%) of dynamic memory, leaving 990 bytes for local variables. Maximum is 2,048 bytes.
processing.app.debug.RunnerException: Sketch too big; see http://www.arduino.cc/en/Guide/Troubleshooting#size for tips on reducing it.
	at cc.arduino.Compiler.size(Compiler.java:315)
	at cc.arduino.Compiler.build(Compiler.java:156)
	at processing.app.Sketch.build(Sketch.java:1111)
	at processing.app.Sketch.exportApplet(Sketch.java:1146)
	at processing.app.Sketch.exportApplet(Sketch.java:1132)
	at processing.app.Editor$DefaultExportHandler.run(Editor.java:2381)
	at java.lang.Thread.run(Thread.java:745)
Sketch too big; see http://www.arduino.cc/en/Guide/Troubleshooting#size for tips on reducing it.
```

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -26,501 +26,829 @@
...
-    animationSettings.put(NEOPIXEL_ANIMATION_RAINBOW_CYCLE, new boolean[] { false, true });
-    animationSettings.put(NEOPIXEL_ANIMATION_FLASH_RANDOM, new boolean[] { true, true });
-    animationSettings.put(NEOPIXEL_ANIMATION_IRONMAN, new boolean[] { true, true });
-    subscribeToRuntime("registered");
-  }
-
-  public void onRegistered(Registration s) {
-    refreshControllers();
-    broadcastState();
+    animationRunner = new AnimationRunner();
```
kwatters: We shouldn't be creating this thread when the service is created.  This should go into the startService lifecycle , not here.   I'm not sure we have a need for the animation runner.  animations run on the microcontroller in mrlcomm.  if users want to create an animation, they would likely write their own python code to script that up.

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -26,501 +26,829 @@
...
+  @Override
+  public void attach(Attachable service) throws Exception {
 
-  // @Override
-  public NeoPixelController getController() {
-    return controller;
-  }
+    if (SpeechSynthesis.class.isAssignableFrom(service.getClass())) {
+      attachSpeechSynthesis((SpeechSynthesis) service);
+      return;
```
kwatters, line 242: What if the service you are attaching to is both a speech synthesis and a neopixel controller?

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -26,501 +26,829 @@
...
 
-  public boolean isAttached() {
+  @Override
+  public void attachNeoPixelController(NeoPixelController neoCntrlr) {
+
     if (controller != null) {
-      if (((Arduino) controller).getDeviceId((Attachable) this) != null) {
-        isAttached = true;
-        return true;
+      if (controller.equals(neoCntrlr.getName())) {
```
kwatters, line 256: this will blow a null pointer exception if neoCntrlr is null.

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -26,501 +26,829 @@
...
+    subscribe(mouth.getName(), "publishEndSpeaking");
+    // equalizer();
   }
 
-  public void setPixel(String address, String red, String green, String blue) {
-    setPixel(address, red, green, blue, "0");
+  @Override
+  public void clear() {
+    // stop java animations
+    animationRunner.stop();
```
kwatters, line 292: this can throw an interrupted exception.. yes?    what if the animation that is running is just on the uc..   we dont' care about the java animation.. the java animation thread shouldn't be running in the first place.

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -26,501 +26,829 @@
...
   }
 
-  public void refresh() {
+  // @Override
+  public String onStartSpeaking(String utterance) {
+    startAnimation();
+    return utterance;
+  }
+
+  public void playAnimation(String animation) {
```
kwatters, line 539: These are all implemented on the UC.. why do we have them here?  This seems like it's more like debug/test code, rather than code that should be running here.

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -26,501 +26,829 @@
...
   @Override
-  public void setAnimation(int animation, int red, int green, int blue, int speed) {
-    // protect against 0 and negative speed
-    if (speed < 1)
-      speed = 1;
-    controller.neoPixelSetAnimation(this, animation, red, green, blue, speed);
-    this.animation = animationIntToString(animation);
-    broadcastState();
+  public void setAnimation(String animation, int red, int green, int blue, int wait_ms) {
+    // TODO convert to int
```
kwatters: This needs to be implemented.

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -26,501 +26,829 @@
...
-        return "Flash Random";
-      case NEOPIXEL_ANIMATION_IRONMAN:
-        return "Ironman";
-      default:
-        log.error("Unknow Animation type {}", animation);
-        return "No Animation";
+  @Override
+  public void setAnimationSetting(String animation) {
+    // TODO Auto-generated method stub
+
```
kwatters: What's the intention of this method?  can we remove it?

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -26,501 +26,829 @@
...
+  }
+
+  public void setBlue(int blue) {
```
kwatters, line 648: Why doe we have something to set blue?  this method should probably be removed.

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -26,501 +26,829 @@
...
+    this.blue = blue;
+  }
+
+  public void setBrightness(int value) {
+    NeoPixelController np2 = (NeoPixelController) Runtime.getService(controller);
```
kwatters, line 653: why does this method need to get a new handle to the neopixel controller.. it should be attached, and therefore have a reference to it already.

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -26,501 +26,829 @@
...
     }
 
+    // update pixel in sequence in matrix
+    Pixel pixel = new Pixel(address, red, green, blue, white);
+
+    // update memory
+    ps.pixels.set(address, pixel);
+
+    // write immediately
+    NeoPixelController np2 = (NeoPixelController) Runtime.getService(controller);
```
kwatters, line 727: This is silly, why do we need to get a handle to the controller each time?

src/main/java/org/myrobotlab/service/VirtualArduino.java
```diff
@@ -244,7 +244,12 @@ public void releaseService() {
     if (runner != null) {
       runner.stop();
     }
+    if (uart != null) {
+      uart.releaseService();
+    }
     releasePeers();
+    sleep(300);
```
kwatters: why is there a sleep here?

src/main/java/org/myrobotlab/service/interfaces/NeoPixelControl.java
```diff
@@ -30,69 +30,45 @@
...
    * @param green
-   *          - value 0-255
    * @param blue
-   *          - value 0-255
    */
   public void setPixel(int address, int red, int green, int blue);
 
   /**
-   * Send a matrix of pixel to the neopixel hardware
+   * Send a matrix of pixels to the neopixel hardware
```
kwatters, line 57: this method should take an array of pixels if that's what it does.

src/main/java/org/myrobotlab/service/interfaces/NeoPixelControl.java
```diff
@@ -30,69 +30,45 @@
...
+  
+  public void setPin(int pin);
 
   public Integer getPin();
 
   public int getNumPixel();
+  
+  public void clear();
+  
+  public void setAnimation(int animation, int red, int green, int blue, int wait_ms);
```
kwatters, line 69: wait_ms is an odd name.. i'm not sure end users will understand what that means.  reality is , this is the amount of time between updates to the animation..  having it defined as some sort of "speed", probably makes more sense.. most commonly, I think people would think in terms of hertz.. which would be the inverse of wait_ms.

src/main/resources/resource/Arduino/MrlComm/MrlNeopixel.h
```diff
@@ -1,45 +1,11 @@
...
-        (((P) >= 62 && (P) <= 69) ? sendBitK(V) : sendBitL(V)))))))))))
-
-#else
-  #define digitalPinToSendBit(P,V) \
-    (((P) >= 0 && (P) <= 7) ? sendBitD(V) : (((P) >= 8 && (P) <= 13) ? sendBitB(V) : sendBitC(V)))
-#endif
-#else
-  #define digitalPinToSendBit(P,V) return
-#endif
+// TODO: Why two enums to tell the animation to stop?  maybe best to just
```
kwatters, line 7: we should remove animation 0 and 1.. they aren't animations..they are control messages.. and if you want to stop an animation you can say "clear()"   if we want an actual stopAnimiation() call, we should add that to just stop the current animation.

src/main/resources/resource/Arduino/generate/arduinoMsgs.schema
```diff
@@ -71,9 +71,14 @@
 < publishI2cData/deviceId/[] data
 
 # NeoPixel
-> neoPixelAttach/deviceId/pin/b32 numPixels/depth
-> neoPixelSetAnimation/deviceId/animation/red/green/blue/b16 speed
+> neoPixelAttach/deviceId/pin/b16 numPixels/depth
```
kwatters, line 74: why did we change to a 16bit here vs 32 from before?

src/main/resources/resource/Arduino/generate/arduinoMsgs.schema
```diff
@@ -71,9 +71,14 @@
...
+> neoPixelSetAnimation/deviceId/animation/red/green/blue/white/b32 wait_ms
```
kwatters, line 75: why did speed change from 16 to 32 bit here (now wait_ms) ?

src/test/java/org/myrobotlab/service/HttpClientTest.java
```diff
@@ -35,36 +30,17 @@ public void testService() throws Exception {
 
       // curl --location --request POST "https://postman-echo.com/post" --data
       // "foo1=bar1&foo2=bar2"
-      http.postForm("https://postman-echo.com/post", "foo1", "bar1", "foo2", "bar2", "foo3", "bar with spaces");
-      log.info(data);
+      String test = http.postForm("https://postman-echo.com/post", "foo1", "bar1", "foo2", "bar2", "foo3", "bar with spaces");
```
kwatters, line 33: bad test, we shouldn't be calling out to an external website from our unit tests.

src/main/resources/resource/Arduino/generate/arduinoMsgs.schema
supertick: because it "is" wait_ms per animation frame
which is far less confusing than the previously defined speed which was based on "who knows what" relativistic attribute (not fps) - therefore it will be in the range of ms to wait for a frame which potentially might be seconds or larger

src/main/resources/resource/Arduino/generate/arduinoMsgs.schema
supertick: because no one will have 65,535 pixels, max do to processing speed is about 2K (very rare)

kwatters: adafruit library says "uint16_t n"   .. so it makes sense to make it 16 here. just noticed the difference.

src/main/resources/resource/Arduino/MrlComm/MrlNeopixel.h
supertick: implemented animation 1 as it previously existed as control message - stop animation ... clear still stops and clears ... 0 is meaningless as it was before

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: it has to be in the constructor - its "final" - its not a thread - its a runnable that manages its own thread ..
I prefer to design like this - its clean - you create it once - it manages its own thread, less likely to create "multiple" and other thread messes

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: you can explicitly attach what you want how you want - you just need to be explicit

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: fixed

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: thread now is only started when equalizer animation starts

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: this is to play any animation - uC or Java-land both with be available be played by name here

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: it has been

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: regrettably "set"Animation vs play animation was used in the past - could deprecated it - it now calls playAnimation

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: These members 
red green blue ... represent a single color for which actions and control messages can be used with.
This is how it worked before - where an animation required a r,g,b value although in some cases it wasn't appropriate.
I separated part of this concept so that a "currently selected color" could be used ... and it is used when you select a color from the UI

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: meh .. direct references are bad ...
true pub/sub doesn't need direct references

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: one step closer to true pub/sub

src/main/java/org/myrobotlab/service/VirtualArduino.java
supertick: I wish I remembered :P
I'll take it out.

src/main/java/org/myrobotlab/service/interfaces/NeoPixelControl.java
supertick: its to set a single pixel - that's why its named setPixel

PR #894, opened 2021-08-03
botname config fix

supertick:


src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -106,6 +106,43 @@
    */
   public ProgramAB(String n, String id) {
     super(n, id);
+
```
kwatters, line 109: I don't like this happening in the constructor..  stuff like this should happen after the object is created.. not during the object creation...  this will not allow an override of getResourceDir() ... so you will be forced to scan a directory every time this object is created.

supertick: As it's currently works I can see that it failed because bot  could not be selected because it was not loaded.
Yes, it should be refactored so that a name can be supplied to config before all possible directories are scanned.
Considering all of what was in startService started in the constructor, I only pulled back the part which breaks the lifecycle
of create->load->start as necessary scans were needed by load but happening in start

kwatters: Ok, so we should refactor it... asking it to load a bot should lazy call the scan for bots or something like that.. we shouldn't do it in the constructor

kwatters: if you want to leave this in the constructor.. then put a TODO to move that code into an init lifecycle call and out of the constructor.

supertick: added TODO

PR #902, opened 2021-08-08
inmoov2 bot fix

supertick:


hairygael:
Hello Grog,
What about the copy of the resource/InMoov2 default configs to data/config ? I see you closed the PR but I don't see something to replace that though

supertick:
updated to handle config too

PR #909, opened 2021-08-09
Grog mpu6050 2

supertick:
working mpu 6050


src/main/java/org/myrobotlab/service/config/Mpu6050Config.java
```diff
@@ -0,0 +1,14 @@
+package org.myrobotlab.service.config;
+
+public class Mpu6050Config extends ServiceConfig {
```
kwatters, line 3: I think this config needs the i2c address as well.. but we can do that in a different PR..

supertick: fixed

PR #910, opened 2021-08-09
motor updates

supertick:
Motor updates / clean up
Added Sabertooth motor controller config
Controllers are name references now 
Concrete linear mapper for serialization


src/main/java/org/myrobotlab/service/interfaces/AxisListener.java
```diff
@@ -0,0 +1,30 @@
+package org.myrobotlab.service.interfaces;
+
+import org.myrobotlab.framework.interfaces.NameProvider;
+import org.myrobotlab.service.data.AxisData;
+
+public interface AxisListener extends NameProvider {
```
kwatters: this should have an attachAxisPublisher and AxisPublisher should be able to attachAxisListener

src/main/java/org/myrobotlab/service/interfaces/AxisPublisher.java
```diff
@@ -0,0 +1,9 @@
+package org.myrobotlab.service.interfaces;
+
+import org.myrobotlab.service.data.AxisData;
+
+public interface AxisPublisher {
```
kwatters: this should have an attachAxisListener in this interface.

src/main/java/org/myrobotlab/service/interfaces/AxisListener.java
supertick: fixed

src/main/java/org/myrobotlab/service/interfaces/AxisPublisher.java
supertick: fixed

PR #921, opened 2021-08-22
Grog subscribe interface change

supertick:


supertick:
pr #930 supersedes this

PR #922, opened 2021-08-22
Grog more UI polish

supertick:
small fix to arduino for neopixel
other small or ui polish too


supertick:
pr in conflict and no longer relevant - closing

PR #923, opened 2021-08-22
removal of writing aiml  ..  Yay !

supertick:


kwatters:
Not so fast.. learned aiml still needs to be written out

PR #926, opened 2021-08-26
removed mrls ProgramAB writing out aiml files

supertick:


src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -690,8 +731,28 @@ public void setCurrentSession(String userName, String botName) {
...
   public void addCategory(Category c) {
-    Bot bot = getBot(getCurrentBotName());
-    bot.brain.addCategory(c);
+    try {
+      Bot bot = getBot(getCurrentBotName());
+      bot.brain.addCategory(c);
+
+      String newCategory = Category.categoryToAIML(c) + "\n";
+
+      File newCatFile = new File(bot.aiml_path + fs + String.format("new_%s_categories.aiml", getName()));
```
kwatters: Isn't this missing the wrapper aiml tag?   what about the close </aiml> tag at the end?  you can't just append here...  it will produce invalid xml.

supertick: yeah ...  took it out anyway - using callbacks now

supertick: updated pom - to program-kw-ab 0.0.8.7 - as far as I know ... this is complete

PR #930, opened 2021-08-30
Interface Management - a way for developers to easily register interfaces they are interested to 'attach' to

hairygael:


src/main/java/org/myrobotlab/service/Adafruit16CServoDriver.java
```diff
@@ -1041,5 +1043,28 @@ public String getBus() {
...
+  @Override
+  public void startService() {
+    super.startService();
+    registerForInterfaceChange(I2CController.class);
+  }
+  
+  @Override
+  public void onInterfaceRegistered(String serviceName, String interfaceName) {
+    if (I2CController.class.toString().equals(interfaceName)) {
+      controllers.add(serviceName);
```
kwatters: I'm not sure I like the idea that we're tracking a set of all possible motor controllers here.  It seems like the real intention of this code is only to provide a helper to the webgui.  Wouldn't it be easier if runtime maintained the interface type to service name list ?  Upon initialization, this service needs to initialize the controllers list. o/w it could get out of sync with the truth that is owned by runtime.

src/main/java/org/myrobotlab/service/Joystick.java
```diff
@@ -247,8 +252,54 @@ public void poll() {
...
+    if (AnalogListener.class.toString().equals(interfaceName)) {
+      analogListenerServices.add(serviceName);
+      broadcastState();
+    }
+  }
+
+  @Override
+  public void onInterfaceReleased(String serviceName, String interfaceName) {
+    if (AnalogListener.class.toString().equals(interfaceName)) {
+      analogListenerServices.remove(serviceName);
```
kwatters: I feel like runtime should be owning the lists of all possible analog listener services.  i worry that when services are maintaining their own lists here, it could get out of sync from the actual list of services from runtime.

src/main/java/org/myrobotlab/service/MouthControl.java
```diff
@@ -23,27 +29,72 @@
...
+  @Deprecated /* future releases members will be protected use appropriate methods instead */
   public int mouthOpenedPos = 4;
+  @Deprecated /* future releases members will be protected use appropriate methods instead */
   public int delaytime = 75;
+  @Deprecated /* future releases members will be protected use appropriate methods instead */
   public int delaytimestop = 150;
+  @Deprecated /* future releases members will be protected use appropriate methods instead */
   public int delaytimeletter = 45;
+  
+  protected final Set<String> servoServices = new HashSet<>();
```
kwatters: where ever we need the list of servo services, we should just ask runtime for it... same with speechServices below.

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
```diff
@@ -135,10 +136,29 @@ public AbstractMotor(String n, String id) {
...
-    refreshControllers();
+    registerForInterfaceChange(MotorController.class);
+    // Runtime.getInstance().subscribeToLifeCycleEvents(getName());
+    // refreshControllers();
+  }
+  
+  @Override
+  public void onInterfaceRegistered(String serviceName, String interfaceName) {
+    if (MotorController.class.toString().equals(interfaceName)) {
+      controllers.add(serviceName);
```
kwatters: tracking the list of controllers here probably shouldn't be done, instead some logic like  "if a controller is registered, and it's the one that I am supposed to be attached to,  then the motor should attach to that controller.

kwatters:
the webgui could subscribe to runtime so it can update the drop downs in the gui for which available controllers exist.  tracking separate lists in each of the services feels like it could get out of sync over time.  (especially in a distributed environment.)

as a nice to have, if a motor sees a controller get registered, and that controller is the one the motor is supposed to be attached to, it should attach at that point.  

I like the services being able to be notified when services/interfaces are registered, but I don't like that individual services would me maintaining a global set of controllers.

supertick:
I'll look into it.  Its cleaner if the service itself propagates the event vs things needing to query or cross-subscribe to runtime, but I mostly agree about having runtime own the data.
What this might mean is runtime publishes the list of interested interfaces to each of the services when those sets change..
e.g.   `Set<String> publishAvailableInterfaces(  Set<String> interfazes) ` becomes a Service method all Services inherit.
This is "pretty" convenient for the ui, however it will need to identify "which" set of interfaces has changed - which currently is taken care of by the service.

kwatters:
I don't think that each service should publish this.  let's take a simple example.. let's say we have 50 servos.. hooked up to 10 arduinos.  when arduino number 11 is connected, it means that 50 servo services will be publishing a list of 11 controllers.  That seems excessive and wasteful.. does it not?  This should be published from runtime directly to the webgui.  otherwise, we'll end up with a lot of useless chatter that ultimately will just cause performance problems as the robots get more complex.

supertick:
Your example is how it works currently...
But the setup for it is done 10 different ways :( 

This was an attempt to standardize the 10 different ways of setup...
I'd be interested in optimizing communication  - kill 2 birds 1 stone, but there are a few points that need resolution.

One is I believe you should be able to query a service and ask for the interfaces it consumes  or registers for.  This is specific to the service type, this is not a property of runtime.

What services can you attach to?  is a question that's specific to a service.   I agree that the service should not need to cache, and ultimately the list of names comes from runtime.  

AbstractServo currently queries runtime on an a onStarted life cycle event of a new service and publishes this info ... which in turn the UI consumes ... this has been around for a very long time  .. (even pre life cycle events)

kwatters:
Yup.. services could implement a method that returns the interfaces that the service is interested in being notified about.
runtime would notify the service when a service that matches those criteria are registered/deregistered.  
helper methods on runtime to get a list of services for an interface would be handy..
I'm hoping we can avoid things being done 10 different ways...  so that'll be some cleanup when we come across it.

src/main/java/org/myrobotlab/service/MouthControl.java
```diff
@@ -23,27 +29,72 @@
...
+  protected final Set<String> speechServices = new HashSet<>();
+
   transient private ServoControl jaw;
   transient private SpeechSynthesis mouth;
+  protected String jawName;
+  protected String mouthName;
 
   public MouthControl(String n, String id) {
     super(n, id);
+    registerForInterfaceChange(ServoControl.class);
```
kwatters, line 48: I feel like all services should register for interfaces they are interested in.  so, the service should have a method that retuns back a list of interface classes ...  when runtime creates the service... it would call that method and register the callbacks automatically.

src/main/java/org/myrobotlab/service/MouthControl.java
```diff
@@ -114,25 +165,98 @@ public synchronized void onEndSpeaking(String utterance) {
...
-    // if (closed < opened) {
-    // jaw.map(closed, opened, closed, opened);
-    // } else {
-    // jaw.map(opened, closed, opened, closed);
-    // }
+  
+  @Override
+  public void onInterfaceRegistered(String serviceName, String interfaceName) {
+    if (ServoControl.class.toString().equals(interfaceName)) {
+      servoServices.add(serviceName);
```
kwatters: we shouldn't be tracking the list of servo services in this service.  that list should come from runtime..

src/main/java/org/myrobotlab/service/MouthControl.java
```diff
@@ -114,25 +165,98 @@ public synchronized void onEndSpeaking(String utterance) {
...
+      broadcastState();
+    }
+    if (SpeechSynthesis.class.toString().equals(interfaceName)) {
+      speechServices.add(serviceName);
```
kwatters: we shouldn't be tracking the list of speech services locally in this service.  that list should be queried from runtime if needed.

src/main/java/org/myrobotlab/service/interfaces/InterfaceChangeListener.java
```diff
@@ -0,0 +1,9 @@
+package org.myrobotlab.service.interfaces;
+
+public interface InterfaceChangeListener {
+
+  public void onInterfaceRegistered(String serviceName, String interfaceName);
```
kwatters: this interface should also have a method , where the change listener returns the list of interface types it's interested in listening for.

src/main/java/org/myrobotlab/service/Joystick.java
supertick: runtime now owns the data structure - "requestedAttachMatrix"
its a definition of {Type} -> { interested itnerface } -> Set of services names that can provide it
this method is no longer needed

src/main/java/org/myrobotlab/service/Adafruit16CServoDriver.java
supertick: runtime owns the data now - and the webgui now processes the single compressed message of changes to the runtime data structure requestedAttachMatrix

Yeah it is a helper function ... for "any" gui .. but I realize its a pretty important concept.
With Java reflection (or any language that support self inspection of a class) - a system can determine what a service "provides" ... but none of them can provide - what a service is "looking for"
So creating these methods and data structure - gives opportunity for services to publish what they are interested in and examine what services can provide it.  The concept goes beyond just UI support.

src/main/java/org/myrobotlab/service/MouthControl.java
supertick: The list changes - it can't just be based on a service querying runtime ... which should be possible.
But if the registry changes - an event needs to be communicated for "interested" services

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
supertick: This is no longer applicable ... I'll update soon

src/main/java/org/myrobotlab/service/MouthControl.java
supertick: yeah - this is no longer applicable ... I'll update soon

src/main/java/org/myrobotlab/service/MouthControl.java
supertick: I agree - removed
Processing is done with normalized runtime data

src/main/java/org/myrobotlab/service/abstracts/AbstractMotor.java
supertick: removed - now using normalized runtime data

src/main/java/org/myrobotlab/service/MouthControl.java
supertick: it currently automatically creates callbacks and invokes them when appropriate events are triggered

src/main/java/org/myrobotlab/service/MouthControl.java
supertick: method is gone - all registering for interface now use normalized runtime data and appropriate callbacks

src/main/java/org/myrobotlab/service/MouthControl.java
supertick: this duplication of data has been removed

src/main/java/org/myrobotlab/service/interfaces/InterfaceChangeListener.java
supertick: this interface has been removed, because it is no longer applicable

kwatters:
We've gotten pretty deep into the lifecycle management of services.  I think it's a step in the right direction.  I'd like to make sure we're going a good direction, so perhaps we can spell out why services need this notification in the first place.  I suspect many service don't care about any specific interface / attach matrix.  (likely that will change over time to the point where all services get registration notifications for other services being started.)

Let's get on the same  page with the use case for this registration.  Let's go back to our good old friend, the Servo service.  

(hobby)Servos are interested in things like servo controllers, because a servo can attach to a servo controller... but here's the question

I figure there are 2 consumers of this new interface being registered.   

1. a servo itself.  If the controller that it is supposed to be attached to.. it should attach at that point.  (think auto-attach..)
2. the webgui to display a dropdown of possible services that a servo could attach to...

Am I thinking about this functionality correctly?  Is there a different use case?

supertick:
It supports exposing the interfaces for which services want to consume, and that I think will support auto-attaching, easier ui development, and the general ability to query what services need or want.

kwatters:
Let's focus on the servo use case for this.  Is there a 3rd use case, or do I have it covered in those 2 bullet points?

I can think of one more use case here.
3. if a servo is attached to a controller, it'd be nice to know that the controller was unregistered, so it can detach/clean up if needed.

src/main/java/org/myrobotlab/service/Adafruit16CServoDriver.java
```diff
@@ -1041,5 +1041,11 @@ public String getBus() {
   public String getAddress() {
     return deviceAddress;
   }
-
+  
+  @Override
+  public void startService() {
+    super.startService();
+    registerForInterfaceChange(I2CController.class);
```
kwatters, line 1048: the framework should call this, subclasses shouldn't have to worry about this detail here.

src/main/java/org/myrobotlab/service/AdafruitMotorHat4Pi.java
```diff
@@ -147,6 +147,7 @@ public AdafruitMotorHat4Pi(String n, String id) {
     super(n, id);
     refreshControllers();
     subscribeToRuntime("registered");
+    registerForInterfaceChange(I2CController.class);
```
kwatters, line 150: the framework should call this, subclasses shouldn't have to worry about this detail here.

src/main/java/org/myrobotlab/service/Joystick.java
```diff
@@ -247,7 +252,8 @@ public void poll() {
   }
 
   public Joystick(String n, String id) {
-    super(n, id);
+    super(n, id); 
+    registerForInterfaceChange(AnalogListener.class);
```
kwatters, line 256: runtime should call this for joystick after the service is created, implicity.

src/main/java/org/myrobotlab/service/MouthControl.java
kwatters: if runtime is handling this , we should remove the registrations here

supertick: this function puts into runtime the fact that this type of service is interested in ServoControl interfaces
its a contract this service is telling runtime about, and runtime in turn notifies the service when a new service with the appropriate requested interface becomes available

src/main/java/org/myrobotlab/service/AdafruitMotorHat4Pi.java
supertick: Sorry I don't understand what your saying ...
registerForInterfaceChange is a detail saying for a specific service "Hey, I'm interested in these interfaces"
What subclass AdafruitMotorhat4Pi is a subclass of service

supertick:
the functionality of this pr was done in several later ones

PR #934, opened 2021-09-06
wrist fix

supertick:


supertick:
Having a whole bunch of direct references I feel is a design flaw with InMoov/InMoov2 
If it used pub/sub to attach limbs etc it would be more robust in that taking a service offline the need to do NPE checking all over the place is removed

PR #940, opened 2021-09-30
removed inmoov

supertick:


src/main/java/org/myrobotlab/service/InMoov.java
```diff
@@ -1,2483 +0,0 @@
...
-
-  public void moveTorsoBlocking(double topStom, double midStom, double lowStom) {
-    if (torso != null) {
-      torso.moveToBlocking(topStom, midStom, lowStom);
-    } else {
-      log.error("moveTorsoBlocking - I have a null torso");
-    }
-  }
-
-  public void onClassification(TreeMap<String, List<Classification>> classifications) {
```
kwatters, line 860: InMoov was copied to InMoov2 before any of this stuff was added to InMoov1...  I think we're throwing away some desired functionality.

PR #944, opened 2021-10-02
Parse available TTS voices under MacOS

argoopen:


argoopen:
Oops - this needs a tiny bit of further work - whilst the available voices are displayed, the selected voice is not yet passed to the 'say' command.

PR #945, opened 2021-10-02
Config fixes and updates

supertick:
* fixes the ability to save and preserve servo values in InMoov2
* fixes null locales
* fixes many npe's
* handles virtual attribute appropriately
* removes name from config - it's now "normalized" with the file name
* cleans out invalid notify entries
* doesnt npe on null locale when creating a chatbot
* removes machine generated attaches from config
* has explicit attaching 
* new AbstractSpeechRecognizer config


src/main/java/org/myrobotlab/service/config/ServiceConfig.java
```diff
@@ -1,19 +1,17 @@
 package org.myrobotlab.service.config;
 
-import java.util.ArrayList;
-
 public class ServiceConfig {
 
-  public String name;
+  // public String name; not normalized
```
kwatters: Why are we removing name ?  all services have a name and a type...

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
```diff
@@ -572,4 +560,29 @@ public void unsetWakeWord() {
     setWakeWord(null);
   }
 
+  @Override
+  public ServiceConfig getConfig() {
+    WebkitSpeechRecognitionConfig config = (WebkitSpeechRecognitionConfig) initConfig(new WebkitSpeechRecognitionConfig());
```
kwatters: Should be AbstractSpeechRecognizerConfig not webkit..

src/main/java/org/myrobotlab/service/config/ServiceConfig.java
supertick: The name is and will always be in runtime.yml
runtime has a list of names of services to create
that name is tied to a filename to configure it
the name inside the file is not normalized and deceiving

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
supertick: oop yep - fixed

PR #948, opened 2021-10-07
Create Hd44780.py

Cyber-One:
Example code for the Hd44780 service.
This is all new documentation :-)

### DIFF
--- a/src/main/resources/resource/Hd44780/Hd44780.py
+++ b/src/main/resources/resource/Hd44780/Hd44780.py
@@ -1,0 +1,51 @@
+#################################################################
+# Example Code                                                  #
+#################################################################
+ 
+#################################################################
+# First start your I2C Bus Master Device                        #
+#################################################################
+# If your using the Arduino Nano, comment out this line and
+# uncomment the ArDuino Nano lines
+raspi = Runtime.createAndStart("raspi","RasPi")
+#################################################################
+# Start the Arduino Nano connected using /dev/ttyUSB0           #
+#################################################################
+#arduinoNano = Runtime.start("arduinoNano","Arduino")
+#arduinoNano.setBoardNano()
+#arduinoNano.connect("/dev/ttyUSB0")
+ 
+#################################################################
+# Next start the PCF8574 service                                #
+#################################################################
+pcf = Runtime.createAndStart("Pcf","Pcf8574")
+# Then attach it to the I2C Bus Master
+# When attaching, we specify the Bus Master Device,
+# the I2C Bus Number
+# and the I2C address
+pcf.attach(raspi,"1","0x27")
+ 
+#pcf.attach(arduinoNano,"0","0x27")
+ 
+#################################################################
+# Next start the Hd44780 service                                #
+#################################################################
+Hd44780 = Runtime.createAndStart("Hd44780","Hd44780")
+ 
+# Once the service has been started, we need to attach it to
+# the PCF service
+Hd44780.attach(pcf)
+ 
+# this will initalise the display.
+Hd44780.init()
+ 
+# when we want to clear the screen call this
+Hd44780.clear()
+ 
+# You can turn the backlight on or off.
+# True will turn it on, False will turn it off.
+Hd44780.stBackLight(True)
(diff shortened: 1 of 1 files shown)


src/main/resources/resource/Hd44780/Hd44780.py
```diff
@@ -0,0 +1,51 @@
...
+# the I2C Bus Number
+# and the I2C address
+pcf.attach(raspi,"1","0x27")
+ 
+#pcf.attach(arduinoNano,"0","0x27")
+ 
+#################################################################
+# Next start the Hd44780 service                                #
+#################################################################
+Hd44780 = Runtime.createAndStart("Hd44780","Hd44780")
```
kwatters: by convention, variable names should start with a lowercase letter.  This example would be slightly more consistent with others if it said:   (first arg is the name of the service, second arg is the type of service.)
hd44780 = Runtime.createAndStart("hd44780","Hd44780")

src/main/resources/resource/Hd44780/Hd44780.py
```diff
@@ -0,0 +1,51 @@
...
+# Start the Arduino Nano connected using /dev/ttyUSB0           #
+#################################################################
+#arduinoNano = Runtime.start("arduinoNano","Arduino")
+#arduinoNano.setBoardNano()
+#arduinoNano.connect("/dev/ttyUSB0")
+ 
+#################################################################
+# Next start the PCF8574 service                                #
+#################################################################
+pcf = Runtime.createAndStart("Pcf","Pcf8574")
```
kwatters: here also.. it's nice if the service name is equal to the variable name in python.. so changing this to 

pcf = Runtime.createAndStart("pcf","Pcf8574")

is a preferable convention

src/main/resources/resource/Hd44780/Hd44780.py
```diff
@@ -0,0 +1,51 @@
...
+# Example Code                                                  #
+#################################################################
+ 
+#################################################################
+# First start your I2C Bus Master Device                        #
+#################################################################
+# If your using the Arduino Nano, comment out this line and
+# uncomment the ArDuino Nano lines
+raspi = Runtime.createAndStart("raspi","RasPi")
+#################################################################
```
supertick: avoid Runtime.createAndStart
just use Runtime.start if you wanted it started - i creates under the hood
Runtime.create is for if you want to have access to the service before starting it

src/main/resources/resource/Hd44780/Hd44780.py
```diff
@@ -0,0 +1,51 @@
...
+# Next start the PCF8574 service                                #
+#################################################################
+pcf = Runtime.createAndStart("Pcf","Pcf8574")
+# Then attach it to the I2C Bus Master
+# When attaching, we specify the Bus Master Device,
+# the I2C Bus Number
+# and the I2C address
+pcf.attach(raspi,"1","0x27")
+
```
supertick: Avoid attach with parameters ...  we want to get rid of them ...
Configure the pcf first with bus and addres "then" attach
e.g. 
pcf.setBus("1") 
pcf.setAddress("0x27")
pcf.attach(raspi)
The reason being this form is directly supportable via configuration and service life cycle - the extra parameter attach is not

src/main/resources/resource/Hd44780/Hd44780.py
Cyber-One: This was a direct copy and paste from the Pcf8574 service page!
http://myrobotlab.org/service/Pcf8574

src/main/resources/resource/Hd44780/Hd44780.py
Cyber-One: This to was a copy paste from http://myrobotlab.org/service/Pcf8574 service page example.
Note there is also an error in the example.
When used with an Arduino as the example Pcf8574 is, the I2C bus is 0, the Raspberry Pi uses 1.

kwatters:
I think we merged this example in a different PR..  I'm going to close this one.  Thanks Cyber-One!

PR #953, opened 2021-10-13
Discordbot

kwatters:
initial impl for the diffbot service.

### DIFF
--- a/src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
+++ b/src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
@@ -28,6 +28,7 @@ public class MrlDiscordBotListener extends ListenerAdapter {
   private final DiscordBot bot;
   public boolean talkToBots = false;
 
+  
   public MrlDiscordBotListener(DiscordBot bot) {
     this.bot = bot;
   }
@@ -37,16 +38,21 @@ public class MrlDiscordBotListener extends ListenerAdapter {
     super.onMessageReceived(event);
     // Create an utterance object from the message.
     Utterance utterance = new Utterance();
+    // Author of the message.
     utterance.username = event.getAuthor().getName();
     utterance.isBot = event.getAuthor().isBot();
     // TODO: maybe we want the raw content? maybe the displayed content?
+    // TODO: replace the bot name if it occurs at the beginning of the utterance.
     utterance.text = event.getMessage().getContentDisplay();
     // TODO: list of other mentions in the utterance.
     // get the response channel and add it to the utterance 
     MessageChannel channel = event.getChannel();
     utterance.channel = channel.getId();
     utterance.channelType = channel.getType().toString();
+    // copy the name of the bot in the channel that is being used to communicate.
+    utterance.channelBotName = bot.discordBotName;
     // publish the utterance!
+    // TODO: don't publish the message if it came from the bots own self!
     bot.invoke("publishUtterance", utterance);
   }
 
--- a/src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
+++ b/src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
@@ -48,66 +48,67 @@ public class MrlDiscordBotListener extends ListenerAdapter {
     utterance.text = event.getMessage().getContentDisplay();
     // TODO: list of other mentions in the utterance.
     // publish this utterance.  we should see it arrive at programab's onUtterance method.
-    bot.invoke("publishUtterance", utterance);
-
+    
     // TODO: what do  I need to get a handle back to the channel
     MessageChannel channel = event.getChannel();
     //    channel.getId();
     //    channel.getType();
     utterance.channel = channel.getId();
-    // TODO: cleanup this message
-    System.err.println("channel... " + channel);
+    utterance.channelType = channel.getType();
+    
+    System.err.println("Invoking publishUtterance(" + utterance + ")");
+    bot.invoke("publishUtterance", utterance);
 
-    if (event.getAuthor().isBot() && talkToBots) {
-      log.info("Not responding to bots.");
-      return;
-    }
-    String userName = event.getAuthor().getName();
-    log.info("Message Received from " + userName + " : " + event.getMessage().getContentDisplay());
-    // Don't talk to myself!
-    if (userName.contentEquals(botName)) {
-      // System.out.println("not me..");
-      return;
-    }
-    boolean shouldIRespond = false;
-    // always respond to direct messages.
-    if (ChannelType.PRIVATE.equals(event.getChannelType())) {
-      shouldIRespond = true;
-    } else {
-      if (!event.getAuthor().isBot()) {
-        // TODO: don't talk to bots.. it won't go well..
-        List<User> mentioned = event.getMessage().getMentionedUsers();
-        for (User u : mentioned) {
-          if (u.getName().equals(botName)) { 
-            shouldIRespond = true;
-            break;
-          }
-        }
-      } 
-
-    }
-
-    // TODO: is there a better way to test for this?
-    if (shouldIRespond) {
-      log.info("I should respond!");
-      // let's respond to the user to their utterance.
-      String utteranceDisp = event.getMessage().getContentDisplay();
-      // let's strip the @+botname from the beginning of the utterance i guess.
-      utteranceDisp = utteranceDisp.replace("@" + botName, "");
-      Response resp = respond(userName, utteranceDisp);
-      // Ok.. now what? respond to the user ...
-      if (!StringUtils.isEmpty(resp.msg) ) {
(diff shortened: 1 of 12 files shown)


src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public void connect(String botName) throws LoginException {
+    // TOOD: create a bot and connect with our token
+    JDABuilder jda = JDABuilder.createDefault(token);
+    MrlDiscordBotListener discordListener = new MrlDiscordBotListener(brain, botName);
+    jda.addEventListeners(discordListener);
+    bot = jda.build();
+    // TODO: what now?
+  }
+
```
supertick, line 68: This should follow the attach pattern
ProgramAB has an example ...
As you said - there currently isn't an interface for a brain - but attach can use the actual class until we get there ..

src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -0,0 +1,101 @@
...
+  String token = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx";
+
+  public DiscordBot(String reservedKey, String inId) {
+    super(reservedKey, inId);
+  }
+  
+  @Override
+  public ServiceConfig load(ServiceConfig c) {
+    DiscordBotConfig config = (DiscordBotConfig)c;
```
supertick, line 42: cool config with token ... :thinking:  config should be enhanced to understand sensitive data - then mebbe the security service can go away, at least its in the user directory and not likely to get pushed to github

src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -0,0 +1,101 @@
...
+    return config;
+  }
+
+  @Override
+  public ServiceConfig getConfig() {
+    // TODO: this is also an ugly pattern. you can't really call super get config here!
+    // ServiceConfig c =  super.getConfig();
+    // TODO: is this unsafe?
+    // TODO: what sets the type of this config?
+    /// TODO: this isn't good OO programming to have to do it this way.
```
supertick, line 53: // Noted ... future prs

src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -0,0 +1,101 @@
...
+    config.token = token;
+    return config;
+  }
+
+  public void connect(String botName) throws LoginException {
+    // TOOD: create a bot and connect with our token
+    JDABuilder jda = JDABuilder.createDefault(token);
+    MrlDiscordBotListener discordListener = new MrlDiscordBotListener(brain, botName);
+    jda.addEventListeners(discordListener);
+    bot = jda.build();
```
supertick, line 64: bot is not used

src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -0,0 +1,101 @@
...
+
+import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.api.JDABuilder;
+
+/**
+ * Discord Bot to connect to a Discord server / channel 
+ * Discord Bots need to be created on the discord server and granted privlidges to join and chat 
+ * with channels.  
+ * The bot user also needs a token that is used to authenticate and identify the bot. 
+ */
```
supertick, line 25: Shouldn't this be a TextListener and TextPublisher ?  or we need new interfaces ?

src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
```diff
@@ -0,0 +1,89 @@
...
+import net.dv8tion.jda.api.entities.User;
+import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
+import net.dv8tion.jda.api.hooks.ListenerAdapter;
+
+/**
+ * Discord bot listener adapter implementation.
+ * This class contains the logic handle a callback when messages
+ * are received by the bot.
+ * 
+ */
```
supertick, line 24: It looks like ListenerAdapter implements the EventListener interface 
If given a choice Discordbot should use the interface instead of creating a separate class - but considering the number of methods (a friggin boatload of callbacks) I can understand the desire to extend from the abstract ListenerAdapter..
But, It looks like **all** are just empty implementations except for one or two...

src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
```diff
@@ -0,0 +1,89 @@
...
+  transient public final static Logger log = LoggerFactory.getLogger(MrlDiscordBotListener.class);
+  private final String botName;
+  private final ProgramAB brain;
+  public boolean talkToBots = false;
+
+  public MrlDiscordBotListener(ProgramAB brain, String botName) {
+    this.brain = brain;
+    this.botName = botName;
+  }
+
```
supertick, line 35: This should publish text right ?

src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
```diff
@@ -0,0 +1,89 @@
...
+            shouldIRespond = true;
+            break;
+          }
+        }
+      } 
+
+    }
+
+    // TODO: is there a better way to test for this?
+    if (shouldIRespond) {
```
supertick: use config

src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
```diff
@@ -0,0 +1,89 @@
...
+      }
+    }
+  }
+
+  private Response respond(String userName, String utterance) {
+    log.info("Get Response " + userName + " for: " + utterance);
+    Response resp  = brain.getResponse(userName, utterance);
+    return resp;    
+  }
+
```
supertick, line 58: Shouldn't this be configured and behave more or less like the HtmlFilter ?
ie. be a TextPublisher and a TextListener and just be attached so its wired as desired ?

src/main/java/org/myrobotlab/framework/repo/Repo.java
```diff
@@ -168,6 +168,9 @@ protected Repo() {
       // org.restlet.jee .. not sure where
       remotes.add(new RemoteRepo("maven-restlet", "https://maven.restlet.talend.com", "Public online Restlet repository"));
 
+      // This is the repo for the Java Discord API for the Discord Bot service lives.
+      remotes.add(new RemoteRepo("dv8tion", "https://m2.dv8tion.net/releases", "Discord Bot - m2-dv8tion"));
+
```
supertick, line 173: bummer

supertick (changes requested):

src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
kwatters: this is controlling what messages should be responded to... for direct communications, the bot will always respond ...  for a message received in the discord channel, it only responds if the bot is mentioned.... and then again only responds if the message did not come from another bot ...  we could probably externalize a little of that behavior to config

supertick (approved):

src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
kwatters: no easy way to wire it in otherwise.

src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
kwatters: it does now :)

src/main/java/org/myrobotlab/discord/MrlDiscordBotListener.java
kwatters: UtterancePublisher and UtteranceListener are born.

src/main/java/org/myrobotlab/service/DiscordBot.java
kwatters: new interfaces!

src/main/java/org/myrobotlab/service/DiscordBot.java
kwatters: agreed.. "secret store" or other integration is a nice to have..

src/main/java/org/myrobotlab/service/DiscordBot.java
kwatters: i've added a shutdown hook.

src/main/java/org/myrobotlab/service/DiscordBot.java
kwatters: hopefully the utterance interfaces will suffice for now.

PR #956, opened 2021-10-18
small neopixel and ui updates

supertick:
mostly neopixel and ui updates ... critical createPeer bugfix too


supertick:
mostly neopixel and ui updates ... critical createPeer bugfix too

PR #958, opened 2021-10-22
A Little DiscordBot Polish

supertick:


src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -129,6 +192,23 @@ public void onUtterance(Utterance utterance) throws Exception {
     }
   }
 
+  public void sendUtterance(String text) {
+    sendUtterance(text, "general");
+  }
+
+  
+  public void sendUtterance(String text, String channelName) {
+    List<TextChannel> channels = bot.getTextChannelsByName(channelName, true);
```
kwatters, line 216: how does this work for private vs publish channel messages?

src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -18,65 +23,121 @@
...
   public ServiceConfig load(ServiceConfig c) {
-    DiscordBotConfig config = (DiscordBotConfig)c;
-    this.token = config.token;
+    DiscordBotConfig config = (DiscordBotConfig) c;
+
+    setToken(config.token);
+
+    if (config.utteranceListeners != null) {
+      for (String local : config.utteranceListeners) {
+        addListener("publishUtterance", local);
```
kwatters: This should use .attachUtteranceListener(string Name)  ... don't listen to a specific method.. use the interface

src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -18,65 +23,121 @@
...
+    // TODO: this is also an ugly pattern. you can't really call super get
+    // config here!
+    // ServiceConfig c = super.getConfig();
     // TODO: is this unsafe?
     // TODO: what sets the type of this config?
     /// TODO: this isn't good OO programming to have to do it this way.
     DiscordBotConfig config = (DiscordBotConfig) initConfig(new DiscordBotConfig());
     config.token = token;
+
+    Set<String> listeners = getAttached("publishUtterance");
```
kwatters, line 107: it's be nice to have a way to know what type of listener , rather than asking about a specific method.

src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -129,6 +192,23 @@ public void onUtterance(Utterance utterance) throws Exception {
     }
   }
 
+  public void sendUtterance(String text) {
```
kwatters, line 202: Add javadoc that this sends a message to the channel called "general"

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1108,7 +1110,12 @@ public ServiceConfig getConfig() {
     Set<String> listeners = getAttached("publishText"); 
     config.textListeners = listeners.toArray(new String[listeners.size()]);
     
+    listeners = getAttached("publishUtterance");
```
kwatters, line 1113: Similar, it's be nice to know about which UtteranceListeners were attached, rather than asking about this specific callback/method.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1135,6 +1142,12 @@ public ServiceConfig load(ServiceConfig c) {
       }
     }
 
+    if (config.utteranceListeners != null) {
+      for (String local : config.utteranceListeners) {
+        addListener("publishUtterance", local);
```
kwatters: Use  the interface,... don't attach to a specific method here..    this.attachUtteranceListener(String name)

src/main/resources/resource/WebGui/app/service/js/DiscordBotGui.js
```diff
@@ -0,0 +1,47 @@
...
+            $scope.onEpoch = data
+            $scope.$apply()
+            break
+        default:
+            console.error("ERROR - unhandled method " + $scope.name + " " + inMsg.method)
+            break
+        }
+    }
+
+    msg.subscribe('publishUtterance')
```
kwatters, line 44: We should ask the java interface for the publishMethods and attach to all of those in a loop.... instead of subscribing to a specific method here. (if possible)

src/main/java/org/myrobotlab/service/DiscordBot.java
supertick: yeah - maybe nicer...
its "nice" not to have some horrible custom local Set<String> variables ... which this solves

src/main/java/org/myrobotlab/service/DiscordBot.java
supertick: dunno ... matches by name .. i assume it could be either ...

src/main/java/org/myrobotlab/service/DiscordBot.java
supertick: agree

src/main/resources/resource/WebGui/app/service/js/DiscordBotGui.js
supertick: another pr - as that should be a base js framework method

src/main/java/org/myrobotlab/service/DiscordBot.java
kwatters: maybe we could ask runtime... we can handle this in a future pr.

src/main/java/org/myrobotlab/service/DiscordBot.java
kwatters: I suspect this might only work for public channels that the bot is in..  not sure. add some javadoc to spell out its behavior a bit.

PR #961, opened 2021-11-02
JUnit Testing Fixes

supertick:
* Service.java - protected ServiceConfig config member added and setName(name) removed isPeerStarted(key) added
* Runtime - release in reverse order, releaseAll with blocking/nonblocking and save runtime options
    reverse order most likely to be clean
    non-blocking - prevent deadlock issue
* Virtual MrlComm.java - fixed npe issues
* Arduino detach - if unconnected quick detach, do not wait for acks on msgs that will never get to mrlcomm
* Arduino - better checks for NPEs
* Arduino - re-entrant neoPixelAttach (previously added 2 when called on the same neopixel)
* AbstractServo - removed stopService - implemented disabling, detaching details in releaseService
* config removed from DeviceMapping (in Arduino)
* Jenkins - skipAfterFailureCount option added to verify
* ArduinoMsgGenerator - VirtualMsgs and MrlComm.java not handling msgs correctly due to invoke switch not set appropriately
* MAX_MSG_SIZE actually checked on Arduino --> MrlComm and dumped with error vs previously sending and the garbage it created
* More NPE checking in Arduino
* MrlNeoPixel virtual device cleaned up
* DiscordBot setBotName getBotName
* Servo - if autoDisable
* WebGui - removed mDNS chatty logging default

### DIFF
--- a/src/main/java/org/myrobotlab/service/InMoov2Head.java
+++ b/src/main/java/org/myrobotlab/service/InMoov2Head.java
@@ -3,8 +3,6 @@ package org.myrobotlab.service;
 import java.io.File;
 import java.io.IOException;
 import java.util.Locale;
-import java.util.Timer;
-import java.util.TimerTask;
 import java.util.concurrent.ThreadLocalRandom;
 
 import org.myrobotlab.framework.Service;
@@ -26,17 +24,15 @@ public class InMoov2Head extends Service {
 
   public final static Logger log = LoggerFactory.getLogger(InMoov2Head.class);
 
+  // peers
   transient public ServoControl jaw;
   transient public ServoControl eyeX;
   transient public ServoControl eyeY;
   transient public ServoControl rothead;
   transient public ServoControl neck;
   transient public ServoControl rollNeck;
-
   transient public ServoControl eyelidLeft;
   transient public ServoControl eyelidRight;
-
-  transient Timer blinkEyesTimer = new Timer();
 
   public InMoov2Head(String n, String id) {
     super(n, id);
@@ -52,29 +48,8 @@ public class InMoov2Head extends Service {
       eyelidRight.setSpeed(tmpVelo);
     moveToBlocking(180, 180);
     moveToBlocking(0, 0);
-
   }
 
-  @Override
-  public void startService() {
-    super.startService();
-    startPeers();
-  }
-
-  class blinkEyesTimertask extends TimerTask {
-    @Override
-    public void run() {
-      int delay = ThreadLocalRandom.current().nextInt(10, 40 + 1);
-      blinkEyesTimer.schedule(new blinkEyesTimertask(), delay * 1000);
-
-      blink();
-      // random double blink
-      if (ThreadLocalRandom.current().nextInt(0, 1 + 1) == 1) {
-        sleep(ThreadLocalRandom.current().nextInt(1000, 2000 + 1));
-        blink();
-      }
-    }
-  }
 
   public void enable() {
     if (eyeX != null)
@@ -150,9 +125,9 @@ public class InMoov2Head extends Service {
     if (rollNeck != null)
       rollNeck.disable();
     if (eyelidLeft != null)
-      eyelidLeft.enable();
+      eyelidLeft.disable();
     if (eyelidRight != null)
-      eyelidRight.enable();
+      eyelidRight.disable();
   }
 
   public long getLastActivityTime() {
@@ -316,13 +291,8 @@ public class InMoov2Head extends Service {
   }
 
   public void releaseService() {
-    try {
-      disable();
-      releasePeers();
-      super.releaseService();
-    } catch (Exception e) {
-      error(e);
-    }
+    disable();
+    super.releaseService();
   }
 
   public void rest() {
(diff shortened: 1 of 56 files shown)


src/main/java/org/myrobotlab/service/Log.java
```diff
@@ -254,6 +254,10 @@ public void setRootLogLevel(String level) {
     getLogLevel();
     broadcastState();
   }
+  
+  @Override
+  public void setName(String name) {
```
kwatters, line 259: why is this here?

src/main/java/org/myrobotlab/service/InMoov2Head.java
```diff
@@ -502,15 +472,12 @@ public void test() {
...
-    }
-    if (param) {
-      blinkEyesTimer = new Timer();
-      new blinkEyesTimertask().run();
-    }
+  /**
+   * FIXME - implement
+   * @param b
+   */
+  public void autoBlink(boolean b) {
```
kwatters, line 479: no functionality anymore..

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1424,17 +1450,17 @@ public static void shutdown() {
...
+
+      // release
       releaseAll();
     } catch (Exception e) {
       log.error("something threw - continuing to shutdown", e);
     }
 
+    // calling System.exit(0) before some specialized threads
+    // are completed will actually end up in a deadlock
+    Service.sleep(1000);
```
kwatters, line 1463: This seems bad.. why would system.exit(0) cause a deadlock?

src/main/resources/resource/WebGui/app/views/tabsViewCtrl.js
```diff
@@ -33,12 +33,12 @@ angular.module('mrlapp.mrl').controller('tabsViewCtrl', ['$location','$scope', '
         $scope.panels = panels
 
         if (!$scope.view_tab && panels.length > 0 && $scope.panels[$scope.panels.length - 1].name.startsWith('intro')) {
```
kwatters, line 35: No body in this if statement.. dead code.

src/test/java/org/myrobotlab/service/ArduinoTest.java
```diff
@@ -483,15 +453,13 @@ public final void testServo() throws Exception {
     // notify and process releasing itself from attached
     // services
     servo.releaseService();
-
+    Service.sleep(200); // async call removal of servo
```
kwatters, line 456: feels like a race condition here...

src/test/java/org/myrobotlab/service/NeoPixelTest.java
```diff
@@ -5,30 +5,32 @@
 
 import static org.junit.Assert.assertTrue;
 
-import org.junit.Before;
+import org.junit.BeforeClass;
```
kwatters, line 8: why BeforeClass instead of just Before?

src/test/java/org/myrobotlab/service/ServoTest.java
```diff
@@ -5,6 +5,7 @@
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
+import org.junit.AfterClass;
```
kwatters, line 8: Why after class instead of just After?

src/main/java/org/myrobotlab/service/Log.java
supertick: required when implementing a Logger interface - ask jvm and jlog makers

src/main/java/org/myrobotlab/service/InMoov2Head.java
supertick: yeah gutted from within, for good reason - but should be re-implemented with sane implementation and config !

src/main/java/org/myrobotlab/service/Runtime.java
supertick: not sure .. but there's a lot of deps & code involved ... mebbe its only eclipse related and their remote tooling ... dunno

src/main/resources/resource/WebGui/app/views/tabsViewCtrl.js
supertick: heh ... you never go here :)
There's lots of bigger fish to fry in these javascript parts

src/test/java/org/myrobotlab/service/ArduinoTest.java
supertick: Arduino is a bit tough ... its attached in two realms java-land and mrlcom land
and they have to be sync'd
and there's connections and serial and ... and...
yeah still too much complexity ... but the "Java land is always right, mrlcomm needs to follow" is certainly helps
(but there are still details .... that are not all resolved)

src/test/java/org/myrobotlab/service/NeoPixelTest.java
supertick: Before is before every test
BeforeClass is before all tests once

src/test/java/org/myrobotlab/service/ServoTest.java
supertick: same as above

PR #963, opened 2021-11-03
Clean Tasks add InvokeFuture

supertick:


src/main/resources/resource/WebGui/app/service/js/ServoGui.js
```diff
@@ -175,11 +170,8 @@ angular.module('mrlapp.service.ServoGui', []).controller('ServoGuiCtrl', ['$time
             onStart: function() {},
             /* - changing only on mouse up event - look in ServoGui.html - cannot do this !!! - sliding to the end an letting go doesnt do what you expect */
             onChange: function() {
-                if ($scope.lockInputOutput) {
-                    $scope.inputSlider.minValue = $scope.outputSlider.minValue
-                    $scope.inputSlider.maxValue = $scope.outputSlider.maxValue
-                }
-                msg.send('map', $scope.inputSlider.minValue, $scope.inputSlider.maxValue, $scope.outputSlider.minValue, $scope.outputSlider.maxValue)
+                // msg.send('map', $scope.inputSlider.minValue, $scope.inputSlider.maxValue, $scope.outputSlider.minValue, $scope.outputSlider.maxValue)
```
kwatters: unrelated to this PR.  is this what we really want to do?  I think the GUI should allow users to configure the min/max in and output...     the conversation from the discord channel was really just to point out that this servo gui is hard to understand.  the color schemes and controls are just not intuitive to use.

PR #966, opened 2021-11-06
Future improvement of config

supertick:


src/main/java/org/myrobotlab/service/config/ServiceConfig.java
```diff
@@ -1,15 +1,27 @@
...
 
-  // The type of the service defined in this config
+  /**
+   * type of service defined for this config
+   */
   public String type;
 
+  public ServiceConfig() {
+    String configTypeName = this.getClass().getSimpleName();
+    String serviceType = configTypeName.substring(0, configTypeName.length() - "Config".length());
```
kwatters, line 17: This means no services can live outside the org.myrobotlab.service package.. yes?    why not make type and name final?  force them to be in the constructor?

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -68,6 +67,8 @@
  * message wrapper
  */
 public class WebGui extends Service implements AuthorizationProvider, Gateway, Handler {
+  
+  protected WebGuiConfig config = new WebGuiConfig();
```
kwatters, line 71: I think this should be private.

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -313,11 +315,11 @@ public boolean allowExport(String serviceName) {
   }
 
   public void autoStartBrowser(boolean autoStartBrowser) {
-    this.autoStartBrowser = autoStartBrowser;
+    config.autoStartBrowser = autoStartBrowser;
```
kwatters, line 318: This means, handling and updating of the config is no longer limited to the same place in the classes...  configs can be updated really anywhere in the class... this is going to lead to places where the config was forgotten about.

src/main/java/org/myrobotlab/service/config/ServiceConfig.java
kwatters: Ignore that previous comment here.... 
I still think these should be made final and passed in.  I don't like having code that does stuff in a pojo.. especially error prone operations such as a substring.

PR #967, opened 2021-11-07
Removed initConfig

supertick:


src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
```diff
@@ -559,7 +559,7 @@ public void unsetWakeWord() {
 
   @Override
   public ServiceConfig getConfig() {
-    AbstractSpeechRecognizerConfig config = (AbstractSpeechRecognizerConfig) initConfig(new AbstractSpeechRecognizerConfig());
+    AbstractSpeechRecognizerConfig config = new AbstractSpeechRecognizerConfig();
```
kwatters: abstract classes probably shouldn't have configs.  they can't be instantiated.

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1387,14 +1387,8 @@ public ServiceConfig load(ServiceConfig config) {
...
-    ServiceConfig sc = new ServiceConfig();
-    initConfig(sc);
-    return sc;
-  }
-
-  protected ServiceConfig initConfig(ServiceConfig config) {
-    config.type = getSimpleName();
-    return config;
+    // FIXME !!! - this should be null for services that do not have it !
+    return new ServiceConfig();
```
kwatters, line 1392: this method..  should be abstract..   (or maybe in the short term add a log message saying no config loading exists for this service type here...  as a reminder that we're missing a config class.)

supertick: long term - with that polished pr i have waiting ... this would only be implemented in Service, and not needed typically in the derived class - so I don't want to make it abstract here...
I don't consider it an error currently for a service not to have config - so I'll log.info it

src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechRecognizer.java
supertick: I'm commenting it out for the moment ...
but I suspect what should happen is this should remain, and the derived speech recognizer should call this super.getConfig()

PR #978, opened 2021-11-21
setting audioPlayer as true

hairygael:


src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -1439,8 +1439,14 @@ public ProgramAB startChatBot() {
         log.error("saving predicates threw", e);
       }
       // start session based on last recognized person
-      if (!chatBot.getPredicate("default", "lastUsername").isEmpty() && !chatBot.getPredicate("default", "lastUsername").equals("unknown")) {
-        chatBot.startSession(chatBot.getPredicate("lastUsername"));
+      //if (!chatBot.getPredicate("default", "lastUsername").isEmpty() && !chatBot.getPredicate("default", "lastUsername").equals("unknown")) {
+        //chatBot.startSession(chatBot.getPredicate("lastUsername"));
+      //}
+      if (!chatBot.getPredicate("Friend", "firstinit").isEmpty() && !chatBot.getPredicate("Friend", "firstinit").equals("unknown") && !chatBot.getPredicate("Friend", "firstinit").equals("started")) {
```
kwatters, line 1445: This hard coded stuff shouldn't be here...  let's work on a generic way to support what you want rather than forcing specific chatbot responses to categories of FIRST_INIT and WAKE_UP.

kwatters: (sorry,  I thought this was in the ProgramAB service when I first reviewed.)  The whole idea of "firstinit" is to set the bots name ?  are there other things that we should do?  When you start the chatbot up for the first time, it's going to default to Friend as the currentUserName... but that really doesn't have anything to do with the InMoovs name being set to something...

hairygael: when you launch the bot the first time it sets as Friend, that's default way you guys decided it should behave. Then it checks if a firstinit was done. This works for all InMoov chatbots. The firstinit allows the user to give a name (botname/ user predicates) for the currentUserName and then ask the name of the user.
This way at next launch, it remembers it's botname and the name of the user.

PR #983, opened 2021-12-07
Servo Performance

supertick:


src/main/java/org/myrobotlab/service/interfaces/ServoSpeed.java
```diff
@@ -0,0 +1,17 @@
+package org.myrobotlab.service.interfaces;
```
kwatters: why is this in the interfaces package?  it's not an interface.

PR #986, opened 2021-12-13
changed the slider for servo position

hairygael:


supertick (commented):
Hi Gael,
excited to see your contributions ! ...
please convert your camelCase class names to use dashes - if your interested these are the reasons - https://stackoverflow.com/questions/7560813/why-are-dashes-preferred-for-css-selectors-html-attributes

hairygael:
Hello Greg,
Not sure where  you need me to put dashes, the camelCase I used is the same name found in the mrl.css. (servoMoveToSlider)[][(https://github.com/MyRobotLab/myrobotlab/blob/develop/src/main/resources/resource/WebGui/app/lib/mrl.css#L3118)](url)

PR #987, opened 2021-12-14
framework for new attach ui

supertick:


kwatters:
this seems like a lot of work ...  i'm curious, why doesn't service just have an abstract method that returns a list of strings.  the strings would represent the interfaces that the service is possible to attach to.  Then, one helper method on runtime for  getServicesForType(string)   where we pass it the type of interface...

supertick:
so  - the way you suggest, would still require "polling" from the UI
The first thing to "know" of a new type of service (change in type relation), or a new instance of a service (change in name list) is Runtime.  If Runtime gets the event, it should be responsible for distribution of that information.
If it sends only the event - it means each UI component must reach back and query ... 
Its more efficient if the Runtime just pushes the data along with the event, vs wait for (potentially all) services to request the data

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -128,8 +129,38 @@
...
+  protected final Map<String, Set<String>> interfaceToType = new HashMap<>();
+
+  protected final Map<String, Set<String>> typeToInterface = new HashMap<>();
+
+  /**
+   * FILTERED_INTERFACES are the set of low level interfaces which we are
+   * interested in filtering out if we want to maintain a data structure which
+   * has "interfaces of interest"
+   */
+  protected final static Set<String> FILTERED_INTERFACES = new HashSet<>(Arrays.asList("org.myrobotlab.framework.interfaces.Broadcaster",
```
kwatters, line 151: not thrilled about having to maintain this list...  perhaps we can figure a way to clean this up so we don't need to list these interfaces here.

PR #991, opened 2021-12-16
Worky MouthControl WebGui

supertick:


src/main/java/org/myrobotlab/service/MouthControl.java
```diff
@@ -17,33 +20,76 @@
  * It's peers are the jaw servo, speech service and an arduino.
  *
  */
-public class MouthControl extends Service implements SpeechListener {
+public class MouthControl extends Service {
```
kwatters: Why not implement SpeechListener anymore?

src/main/java/org/myrobotlab/service/MouthControl.java
```diff
@@ -24,26 +28,69 @@
...
+    registerForInterfaceChange(ServoControl.class);
+    registerForInterfaceChange(SpeechSynthesis.class);
   }
-
+  
   public void attach(Attachable attachable) {
     if (attachable instanceof SpeechSynthesis) {
       mouth = (SpeechSynthesis) attachable;
-      mouth.attachSpeechListener(this);
+      subscribe(mouth.getName(), "publishStartSpeaking");
```
kwatters: why not just use the attachSpeechListener(this) call?  we're duplicating code and implementation this way..

PR #993, opened 2021-12-17
publishServoMoveTo with sane params and return

supertick:


src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -601,7 +602,8 @@ public Double moveTo(Double newPos) {
       return newPos;
     }
 
-    boolean validMoveRequest = processMove(newPos, false, null);
+    broadcast("publishServoMoveTo", newPos);
```
kwatters: without the name of the servo publishing this, this is a pretty useless broadcast, no?

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -713,32 +715,32 @@ public String publishServoDisable(ServoControl sc) {
...
+    log.info("publishServoEnable(servo)");
     return publishServoEnable((ServoControl) sc);
   }
 
   @Override
-  public ServoControl publishServoMoveTo(ServoControl sc) {
-    log.warn("publishServoMoveTo(servo)");
-    return sc;
+  public ServoMove publishServoMoveTo(Double pos) {
+    log.info("publishServoMoveTo({})");
```
kwatters: missing the pos if you are going to log this message here.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -713,32 +715,32 @@ public String publishServoDisable(ServoControl sc) {
...
+    return new ServoMove(getName(), pos);
   }
 
   @Override
   public ServoSpeed publishServoSetSpeed(ServoControl sc) {
-    log.warn("publishServoSetSpeed(servo)");
+    log.info("publishServoSetSpeed(servo)");
```
kwatters: should probably log  servo.getName()   and subsitute it via {} in the log message.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -713,32 +715,32 @@ public String publishServoDisable(ServoControl sc) {
...
     return new ServoSpeed(sc.getName(), sc.getSpeed());
   }
 
   @Override
   public ServoControl publishServoStop(ServoControl sc) {
-    log.warn("publishServoStop(servo)");
+    log.info("publishServoStop(servo)");
```
kwatters: should probably log  servo.getName()   and subsitute it via {} in the log message.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -988,7 +990,7 @@ public void writeMicroseconds(int uS) {
    */
   @Override
   public ServoEvent publishServoStarted(String name, Double position) {
-    log.warn("publishServoStarted(name)");
+    log.info("publishServoStarted(name)");
```
kwatters: incorrect syntax for logging the name of the servo here.

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -1055,7 +1057,7 @@ public void startService() {
 
   @Override
   public String publishServoEnable(String name) {
-    log.warn("publishServoEnable(name)");
+    log.info("publishServoEnable(name)");
```
kwatters: log the actual name here.. not just the string "name"

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
supertick: Still useful to the UI which inherently knows which servo ... but I totally agree - it should return a tuple that has name and position ... (future pr)

src/main/java/org/myrobotlab/service/interfaces/ServoControlPublisher.java
```diff
@@ -9,7 +10,7 @@
  */
 public interface ServoControlPublisher {
 
-  public ServoControl publishServoMoveTo(ServoControl sc);
+  public ServoMove publishServoMoveTo(Double pos);
```
kwatters: This should take a ServoMove object, not a double, right?

src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -128,6 +128,9 @@ protected boolean processMove(Double newPos, boolean blocking, Long timeoutMs) {
       log.info("{} is currently blocking - ignoring request to moveTo({})", getName(), newPos);
       return false;
     }
+    
+    broadcast("publishServoMoveTo", getTargetOutput());
```
kwatters: This should be passing a servoMove object,... and why is this the output ? shouldn't it be publishing the input ?

supertick: current method signature in AbstractServo
```java
  @Override
  public ServoMove publishServoMoveTo(Double pos) {
    log.debug("{}.publishServoMoveTo({})", getName(), pos);
    return new ServoMove(getName(), pos);
  }
```
could make it symmetrical, but not all interfaces adhere to that ... :shrug: - don't have much opinion

To answer the more significant question I think,
is this input or output - This method was made to be consumed by the ServoController ...
And it always has been output.  ie.  the ServoController.onServoMoveTo requires output - but potentially i can find this itself by using the service name to get the service to get the Servo.getTargetOutput()

... its a good question :shrug:

src/main/java/org/myrobotlab/service/interfaces/ServoControlPublisher.java
supertick: yeah ... it could

PR #994, opened 2021-12-17
worky ultrasonicsensor and ui

supertick:


src/main/java/org/myrobotlab/service/UltrasonicSensor.java
```diff
@@ -113,11 +121,51 @@ public void attach(UltrasonicSensorController controller, Integer trigPin, Integ
...
+    broadcastState();
   }
 
-  private boolean isAttached(UltrasonicSensorController controller) {
-    return this.controller == controller;
+  // TODO - this could be Java 8 default interface implementation
+  @Override
+  public void detach(String controllerName) {
+    if (controller == null || !controllerName.equals(controller.getName())) {
+      return;
```
kwatters, line 141: this should probably set isAttached=false here...  consider the scenario where you attach the ultrasonic sensor.. then someone via python, sets the controller =null..   you would never be able to detach properly..

supertick: agree - fixed

PR #1002, opened 2021-12-20
Hd fixes - display consistently worky now

supertick:


src/main/java/org/myrobotlab/service/Hd44780.java
```diff
@@ -220,6 +231,7 @@ private void lcdWrite(byte cmd) {
   private void lcdWrite(byte cmd, byte mode) {
```
kwatters: maybe this should be synchronized ? is it possible that multiple threads could call this and screw up the data...

PR #1003, opened 2021-12-23
Arduino PinListener and PinArrayListener refactor

supertick:
* started with Pir - needs to attach appropriately to Arduino
* removed mrlcomm upload kruft from Arduino service
* changed attach(PinListener) to appropriate explicty attach{type}({type})
* Pins are Strings ! Addresses are Integers ! 
* RasPi consistent pin & address indexes
* pinIndex is <String, PinDefintion>, addressIndex is <Integer, PinDefintion>
* inhanced TestCatcher to provide PinArrayListener and PinListener for testing
* AbstractMicrocontroller pinIndex is <String, PinDefintion>, addressIndex is <Integer, PinDefintion> be consistent
* AbstractMotor - should not implement PinListener .. this is dependent on type of motor
* PinDefintion value is integer not double


src/main/java/org/myrobotlab/service/interfaces/PinArrayPublisher.java
```diff
@@ -10,7 +10,7 @@
  */
 public interface PinArrayPublisher extends NameProvider {
 
-  public void attach(PinArrayListener listener);
+  public void attachPinArrayListener(PinArrayListener listener);
```
kwatters, line 13: should include detachPinArrayListener for completness.

src/main/java/org/myrobotlab/service/Ads1115.java
```diff
@@ -438,12 +438,12 @@ public void attach(I2CController controller, String deviceBus, String deviceAddr
   }
 
   @Override
-  public void attach(PinArrayListener listener) {
+  public void attachPinArrayListener(PinArrayListener listener) {
```
kwatters, line 441: this method should be moved to the interface as a default method and removed from all services.

src/main/java/org/myrobotlab/service/data/PinData.java
```diff
@@ -15,23 +15,19 @@
   /**
    * The value of the pin - can support analog or digital pins
    */
-  public Double value;
+  public Integer value;
```
kwatters, line 18: why not double?

src/test/java/org/myrobotlab/service/ServoTest.java
```diff
@@ -195,40 +195,44 @@ public void releaseService() {
...
   public void testAutoDisable() throws Exception {
-    // Start the test servo.
-    servo01.detach();
-    servo01.setPin(pin);
-    servo01.setPosition(0.0);
+    // Start the test servo
+    arduino01 = (Arduino) Runtime.start(arduino01.getName(), "Arduino");
+    arduino01.connect(port01);
+    servo.detach();
+    sleep(300);
```
kwatters, line 202: this isn't good to have this sleep here.. servo.detach() should probably be blocking/synchronous....  otherwise we'll end up with race conditions elsewhere.

kwatters: perhaps this detach is only an issue in virtual land..

PR #1005, opened 2022-01-02
Grog tracking 1

supertick:


src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -610,6 +612,25 @@ public Double moveTo(Double newPos) {
...
+   * if the moveTo was successfully started - otherwise returns null
+   * @param increment
+   * @return
+   */
+  public Double moveIncr(Double increment) {
+    if (increment == null) {
+      log.info("will not move to null position - not moving");
+      return null;
+    }
+    double newPos = getCurrentInputPos() + increment;
```
kwatters, line 628: during a speed controlled movement, with a low speed set, this will have a strange behavior.  Does getCurrentInputPos take into account the estimated position of the servo as it's moved at this instance in time?

supertick: onEncoderData sets currentOutputPos - and the getCurrentInputPos() reverse map calculates the "current" input from the current output ...

Not pretty but - all that was existing before ...

TLDR - yes

src/main/resources/resource/WebGui/app/service/views/AudioFileGui copy.html
```diff
@@ -0,0 +1,51 @@
+<div class="col-md-6">
```
kwatters: should this file be deleted?  why is there a copy of the AudioFileGui ?

src/main/resources/resource/WebGui/app/service/views/ServoGui copy.html
```diff
@@ -0,0 +1,196 @@
+<!-- ServoGui.html -->
```
kwatters: i suspect this file shouldn't be checked in?

src/main/resources/resource/WebGui/app/service/js/OpenCVGui.js
```diff
@@ -38,133 +38,14 @@ angular.module('mrlapp.service.OpenCVGui', []).controller('OpenCVGuiCtrl', ['$sc
...
-                    showTicks: true,
-                    value: 2,
-                    step: 1,
-                    onChange: function(id) {
-                        $scope.setFilterState()
-                    }
-                }
-            },
-
+            algorithm: 'mean'
```
kwatters, line 41: what happened to these settings?   did this stuff move somewhere else?

supertick: this is replacement of rzslider custom library with standard html5 input/range ..
the defaults rightfully come from the java filters themselves

src/main/java/org/myrobotlab/service/Tracking.java
```diff
@@ -1,592 +1,395 @@
...
-    pid.setSetpoint("y", 0.5); // set center
+    registerForInterfaceChange(ServoControl.class);
+    registerForInterfaceChange(PidControl.class);
+    registerForInterfaceChange(ComputerVision.class);
 
-  }
+    // defaults - pre getConfig or load ?
+    trackingFilters.add("FaceDetectDNN");
+
+    // publishes state/status info from this service
```
kwatters, line 83: why is this in the constructor?  it should be in the start service lifecycle instead.

src/main/java/org/myrobotlab/service/Tracking.java
```diff
@@ -1,592 +1,395 @@
...
+   */
+  // public void onOpenCVData(OpenCVData data) {
+  //
+  // }
 
-  public ServoControl getY() {
-    return servoControls.get("y").servoControl;
-  }
+  public void onClassification(Map<String, List<Classification>> data) {
+    stats.inputCnt++;
```
kwatters, line 213: not all opencv filters publish this method signature.. if you start the dl4j filter, it causes this method to blow up.

src/main/java/org/myrobotlab/service/Tracking.java
```diff
@@ -1,592 +1,395 @@
...
+    }
+    if (tilt == null) {
+      tilt = String.format("%s.tilt", getName());
+      Runtime.start(tilt, "Servo");
+      attachTilt(tilt);
+    }
+    if (pid == null) {
+      pid = String.format("%s.pid", getName());
+      Runtime.start(pid, "Pid");
+      attach(pid);
```
kwatters, line 351: this should configure default pid settings here.  the servo names are known in this scenario.

PR #1009, opened 2022-01-09
Python callback fix

supertick:


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1700,6 +1705,15 @@ public void sendToPeer(String peerName, String method, Object... data) {
   }
 
   public void send(String name, String method, Object... data) {
+    // if you know the service is local - use same thread
+    // to call directly
+    ServiceInterface si = Runtime.getService(name);
```
kwatters, line 1710: i'm consistently getting the arduino unit test stuck...  i suspect its related to this change

src/test/java/org/myrobotlab/arduino/ArduinoServoConfigTest.java
```diff
@@ -25,12 +24,10 @@
  * @author kwatters
  *
  */
+
+@Ignore
 public class ArduinoServoConfigTest {
```
kwatters, line 29: why is this ignored now?

PR #1012, opened 2022-01-15
Way to supply sets of configuration defaults

supertick:
new way to provide sets of configuration defaults baked into the services themselves
Since its working with the POJO config objects - this will be supportable because field names or type changes will also be sync'd with the defaults.

### DIFF
--- a/src/main/java/org/myrobotlab/framework/Service.java
+++ b/src/main/java/org/myrobotlab/framework/Service.java
@@ -2813,5 +2813,14 @@ public abstract class Service implements Runnable, Serializable, ServiceInterfac
   protected void registerForInterfaceChange(Class<?> clazz) {
     Runtime.getInstance().registerForInterfaceChange(getClass().getCanonicalName(), clazz);
   }
+  
 
+  /**
+   * default emtpy getDefaultConfigs - will be overriden by services which provide sane defaults
+   * @return
+   */
+ static public Map<String, Map<String, ServiceConfig>> getDefaultConfigs() {
+   Map<String, Map<String, ServiceConfig>> ret = new HashMap<String, Map<String, ServiceConfig>>();
+   return ret;
+ }
 }
--- a/src/main/java/org/myrobotlab/service/Runtime.java
+++ b/src/main/java/org/myrobotlab/service/Runtime.java
@@ -3818,11 +3818,10 @@ public class Runtime extends Service implements MessageListener, ServiceLifeCycl
    * their requested interfaces - interfaces they are interested in.
    * 
    * This data should be published whenever new "Type" definitions are found
-   *  
+   * 
    * @param targetedInterface
-   *          - interface this    
-   *          add new interface to requested interfaces - add current names of
-   *          services which fulfill that interface "IS ASKING"
+   *          - interface this add new interface to requested interfaces - add
+   *          current names of services which fulfill that interface "IS ASKING"
    * 
    */
   public void registerForInterfaceChange(String targetedInterface) {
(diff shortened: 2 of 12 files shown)


src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3864,5 +3863,86 @@ public void addServiceToRequestedInteface(String serviceName) {
...
   }
+  
+  /**
+   * Saves a "sane" set of embedded defaults constructed for this service
+   * @param className - the class whos defaults will be saved
+   * @return - returns the set of configuration sets successfully saved
+   */
+  static public Set<String> saveDefaults(String className) {
+    try {
+      return saveDefaults(className, null, null);
```
kwatters, line 3874: you should not pass null for overwrite here.  be explicit.

kwatters: same for the configPrefixPath .. be explicit so it's easier to read the code.. o/w you have to dig through to see that it gets defaulted in the other saveDefaults method.

src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -235,12 +235,15 @@ public ServiceConfig load(ServiceConfig c) {
     if (config.autoDisable) {
       disable();
     }
-    
-    mapper = new MapperLinear(config.minIn, config.maxIn, config.minOut, config.maxOut);
+    if (config.minIn != null && config.maxIn != null && config.minOut != null && config.maxOut != null) {
```
kwatters, line 238: this seems unrelated to this PR.. no?

src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -235,12 +235,15 @@ public ServiceConfig load(ServiceConfig c) {
...
-    
-    mapper = new MapperLinear(config.minIn, config.maxIn, config.minOut, config.maxOut);
+    if (config.minIn != null && config.maxIn != null && config.minOut != null && config.maxOut != null) {
+      mapper = new MapperLinear(config.minIn, config.maxIn, config.minOut, config.maxOut);
+    }
     mapper.setInverted(config.inverted);
     mapper.setClip(config.clip);
     enabled = config.enabled;
-    idleTimeout = config.idleTimeout;
+    if (config.idleTimeout != null) {
```
kwatters, line 244: unrelated to this PR.

src/main/java/org/myrobotlab/service/config/OpenCVConfig.java
```diff
@@ -11,8 +11,8 @@
   public String grabberType;
   public String inputSource;
   public String inputFile;
-  public boolean nativeViewer;
-  public boolean webViewer;
+  public boolean nativeViewer = true;
```
kwatters, line 14: unrelated to this PR.

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -2813,5 +2813,14 @@ public boolean isPeerStarted(String peerKey) {
   protected void registerForInterfaceChange(Class<?> clazz) {
     Runtime.getInstance().registerForInterfaceChange(getClass().getCanonicalName(), clazz);
   }
+  
 
+  /**
+   * default emtpy getDefaultConfigs - will be overriden by services which provide sane defaults
+   * @return
+   */
+ static public Map<String, Map<String, ServiceConfig>> getDefaultConfigs() {
```
kwatters, line 2822: I thought we were going to drop the outer map ?

kwatters: I still think that this method signature should return a simple  List<ServiceConfig>  and that Service config should contain the "name" of the service as a member.  If we don't want to include name in the service config, then we should return something like  List<Pair<String, ServiceConfig>>   where the pair is a name/config pair.  The order of this list should determine the runtime start order..  and runtime should not be included in this returned list of service configs.

src/main/java/org/myrobotlab/service/Tracking.java
```diff
@@ -303,6 +309,67 @@ public ServiceConfig getConfig() {
     return config;
   }
 
+  static public Map<String, Map<String, ServiceConfig>> getDefaultConfigs() {
+
+    Map<String, ServiceConfig> config = new HashMap<>();
+
+    RuntimeConfig runtime = new RuntimeConfig();
+    runtime.registry = new String[] { "controller", "cv", "tilt", "pan", "pid", "tracking" };
```
kwatters, line 317: this shouldn't be defined here.  we should change the return signature so that defining the runtime registery is unnecessary here.  see comment on the base class.

PR #1014, opened 2022-01-16
create start then load

supertick:


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -2813,5 +2815,15 @@ public boolean isPeerStarted(String peerKey) {
...
     Runtime.getInstance().registerForInterfaceChange(getClass().getCanonicalName(), clazz);
   }
+  
 
+  /**
+   * default emtpy getDefaultConfig - will be overriden by services which provide sane defaults
+   * @return
+   */
+ static public Map<String, ServiceConfig> getDefaultConfig() {
+   Map<String, ServiceConfig> ret = new HashMap<>();
```
kwatters: i feel like returning null is better than returning an empty config..

src/main/java/org/myrobotlab/service/Joystick.java
```diff
@@ -177,12 +177,14 @@ public void poll() {
         if (pollingController == null) {
           error("controller is null - can not poll");
           stopPolling();
+          break;
```
kwatters, line 180: does this kill the polling thread?  is this change related to tracking?

src/main/java/org/myrobotlab/service/Joystick.java
```diff
@@ -177,12 +177,14 @@ public void poll() {
...
           error("controller is null - can not poll");
           stopPolling();
+          break;
         }
 
         // get the data
         if (isPolling && !pollingController.poll()) {
           error("failed to poll controller");
           stopPolling();
+          break;
```
kwatters, line 187: see comment above

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -16,6 +16,8 @@
 import java.util.HashMap;
```
kwatters: this file shouldn't have any changes in this PR.. perhaps we need to merge develop back into this branch?

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -2813,5 +2815,15 @@ public boolean isPeerStarted(String peerKey) {
   protected void registerForInterfaceChange(Class<?> clazz) {
     Runtime.getInstance().registerForInterfaceChange(getClass().getCanonicalName(), clazz);
   }
+  
 
+  /**
+   * default emtpy getDefaultConfig - will be overriden by services which provide sane defaults
+   * @return
+   */
+ static public Map<String, ServiceConfig> getDefaultConfig() {
```
kwatters: I thought we were going to change the return signature here to force a LinkedHashMap ?

src/main/java/org/myrobotlab/service/Tracking.java
```diff
@@ -303,6 +305,66 @@ public ServiceConfig getConfig() {
     return config;
   }
 
+  static public Map<String, ServiceConfig> getDefaultConfig() {
+
+    Map<String, ServiceConfig> config = new HashMap<>();
+
+    RuntimeConfig runtime = new RuntimeConfig();
```
kwatters: i thought we were going to get rid of runtime config here because we are using a linked hash map for the name/service config pairs ..

src/main/java/org/myrobotlab/framework/Service.java
supertick: a bunch of npe safety checks then ... not sure i feel the same way

src/main/java/org/myrobotlab/service/Joystick.java
supertick: 10K ways of implementing a worker ..
it was one of my early implementations ...  it seems kind of goofy ... i should refactor it into my most recent standards ...
but yes the thread will leave run

src/main/java/org/myrobotlab/service/WebGui.java
supertick: missed import or ctrl-O'd it out in another commit :shrug:

src/main/java/org/myrobotlab/framework/Service.java
supertick: heh .. upcoming commit - we're on the same page

src/main/java/org/myrobotlab/service/config/ArduinoConfig.java
```diff
@@ -1,8 +1,32 @@
...
+
 public class ArduinoConfig extends ServiceConfig {
+  
 
   public String port;
   public boolean connect;
+  public String serial; // name of serial service 
+  
+  @Override
+  public LinkedHashMap<String, ServiceConfig> getDefault(String name) {
```
kwatters: shoudln't this be getDefaultConfig on the Arduino class?   why is it on the ArduinoConfig ?

src/main/java/org/myrobotlab/framework/Service.java
kwatters: This method... should be fixed.. to have the correct return signature and it should be moved to the ServiceInterface.

src/main/java/org/myrobotlab/service/config/TrackingConfig.java
```diff
@@ -9,4 +14,69 @@
   public boolean enabled;
   public long lostTrackingDelayMs = 1000;
 
+  @Override
+  public LinkedHashMap<String, ServiceConfig> getDefault(String name) {
```
kwatters: Why is this method here if you have getDefaultConfig on Tracking.java?

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -2814,4 +2817,8 @@ protected void registerForInterfaceChange(Class<?> clazz) {
     Runtime.getInstance().registerForInterfaceChange(getClass().getCanonicalName(), clazz);
   }
 
+  public Map<String, ServiceConfig> getDefault() {
```
kwatters: please correct the return type here. it should be   LinkedHashMap<String, ServiceConfig>   otherwise services could create an unordered configset.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3865,4 +3852,104 @@ public void addServiceToRequestedInteface(String serviceName) {
...
+   * @param configPrefixPath
+   *          - prefix location to where the configuration sets will be saved
+   * @param overwrite
+   *          - force overwriting existing config sets
+   * @return - set of successfully saved configuration sets
+   * @throws IOException
+   */
+  static public Set<String> saveDefault(String name, String type, String configPrefixPath, Boolean overwrite) throws IOException {
+
+    Set<String> savedPaths = new HashSet<>();
```
kwatters, line 3900: Shoudl this be a TreeSet instead of a HashSet?  This is unordered.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3865,4 +3852,104 @@ public void addServiceToRequestedInteface(String serviceName) {
...
+        String path = FileIO.gluePaths(dir.getAbsolutePath(), service + ".yml");
+        FileIO.toFile(path, CodecUtils.toYaml(config.get(service)));
+      }
+
+      if (!config.containsKey("runtime")) {
+        // config did not come with an explicit runtime
+        // therefore we will create one with the order of the keyset
+        Map<String, ServiceConfig> runtimeConfig = ServiceInterface.getDefault("runtime", "Runtime");
+        RuntimeConfig rconfig = (RuntimeConfig) runtimeConfig.get("runtime");
+        rconfig.registry = config.keySet().toArray(new String[] {});
```
kwatters, line 3940: This is unordered here...

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3865,4 +3852,104 @@ public void addServiceToRequestedInteface(String serviceName) {
...
+
+      for (String service : config.keySet()) {
+        String path = FileIO.gluePaths(dir.getAbsolutePath(), service + ".yml");
+        FileIO.toFile(path, CodecUtils.toYaml(config.get(service)));
+      }
+
+      if (!config.containsKey("runtime")) {
+        // config did not come with an explicit runtime
+        // therefore we will create one with the order of the keyset
+        Map<String, ServiceConfig> runtimeConfig = ServiceInterface.getDefault("runtime", "Runtime");
```
kwatters: to be explicit about the ordering of the keyset here, you probably should make this a LinkedHashMap

src/main/java/org/myrobotlab/service/Tracking.java
```diff
@@ -303,6 +305,68 @@ public ServiceConfig getConfig() {
     return config;
   }
 
+  static public Map<String, ServiceConfig> getDefaultConfig() {
```
kwatters, line 308: This return type is incorrect,  you should add the @Override annotation here.

kwatters: This method should be removed...  you have getDefault() below..   don't put default configs in 2 places on this service.

src/main/java/org/myrobotlab/service/Tracking.java
```diff
@@ -329,64 +393,106 @@ public ServiceConfig load(ServiceConfig c) {
...
+    ServoConfig tiltConfig = (ServoConfig) config.get(trackingConfig.tilt);
+    tiltConfig.pin = "5";
+    tiltConfig.autoDisable = true;
+    tiltConfig.idleTimeout = 3000;
+    tiltConfig.controller = controller;
+
+    PidConfig pidConfig = (PidConfig) config.get(trackingConfig.pid);
+    PidData panData = new PidData();
+    panData.kp = 0.015;
+    panData.ki = 0.005;
```
kwatters: you should be explicit and also define kd here.

src/main/java/org/myrobotlab/service/Tracking.java
```diff
@@ -329,64 +393,106 @@ public ServiceConfig load(ServiceConfig c) {
...
+
+    pidConfig.data.put(trackingConfig.pan, panData);
+
+    PidData tiltData = new PidData();
+    tiltData.kp = 0.015;
+    tiltData.ki = 0.005;
```
kwatters: you should be explicit and also define kd here.

src/main/java/org/myrobotlab/service/Tracking.java
```diff
@@ -329,64 +393,106 @@ public ServiceConfig load(ServiceConfig c) {
...
-      Runtime.start(pid, "Pid");
-      attach(pid);
-    }
-
-    if (cv == null) {
-      cv = String.format("%s.cv", getName());
-      Runtime.start(cv, "OpenCV");
-      attach(cv);
-    }
+  static public LinkedHashMap<String, ServiceConfig> getDefault(String name) {
```
kwatters, line 339: add the @ Override annotation to this method... you are overriding.

src/main/java/org/myrobotlab/service/config/ServiceConfig.java
```diff
@@ -1,20 +1,33 @@
 package org.myrobotlab.service.config;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.util.LinkedHashMap;
+import java.util.Map;
```
kwatters: cleanup / remove unused imports here.

src/main/java/org/myrobotlab/service/Arduino.java
```diff
@@ -2241,12 +2235,36 @@ public ServiceConfig getConfig() {
...
+      serial = (Serial)Runtime.start(config.serial);
+    }
 
     if (config.port != null) {
       connect(config.port);
     }
     return c;
   }
+  
+  static public LinkedHashMap<String, ServiceConfig> getDefault(String name) {
```
kwatters, line 2250: Where is the @ Override annotation..  the method signature is wrong.. it should include the type...  as per the interface.

src/main/java/org/myrobotlab/service/Tracking.java
kwatters: Also this method signature isn't correct...  it's missing the Type...   (as per ServiceInterface.)

src/main/java/org/myrobotlab/framework/interfaces/ServiceInterface.java
```diff
@@ -162,4 +172,55 @@
 
   public int getCreationOrder();
 
+  static public LinkedHashMap<String, ServiceConfig> getDefault(String name, String type) {
```
kwatters, line 175: Why do you need to pass Type here?  the service interface contains getType()  ...  you should be able to make that static and use it here also to avoid needing to pass the type in.

kwatters: i stand semi-corrected...  ignore this.

src/main/java/org/myrobotlab/framework/Service.java
kwatters: this could also be "final" method so that no subclass could override it.

src/main/java/org/myrobotlab/framework/interfaces/ServiceInterface.java
kwatters: actually.. don't ignore this.. you can derive the type of the implementing class with
`public static String getType() {
    return MethodHandles.lookup().lookupClass().getSimpleName();
  };`

PR #1017, opened 2022-01-23
Fully async ui

supertick:


src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -95,6 +96,8 @@ public void handle(AtmosphereResource r) {
     }
   }
 
+  private final transient IncomingMsgQueue inMsgQueue = new IncomingMsgQueue();
```
kwatters, line 99: this should probably have a max message size. but if it gets large enough to run the jvm out of memory, there are likely other problems.

PR #1028, opened 2022-01-30
fixes for mouthcontrol and neopixel

supertick:


src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -907,12 +878,7 @@ public static void main(String[] args) throws InterruptedException {
 
       NeoPixel neopixel = (NeoPixel) Runtime.start("neopixel", "NeoPixel");
 
-      boolean done = true;
-      if (done) {
-        return;
-      }
-
-      neopixel.setPin(8);
+      neopixel.setPin(26);
```
hairygael, line 881: why pin 26?

supertick: Hi Gael,
The main() method inside a service is used for diagnostic purposes - its never called when myrobotlab.jar is running.
So it's a arbitrary, just depending on the developers hardware at the time.

PR #1045, opened 2022-02-08
omni push

supertick:


src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
```diff
@@ -387,6 +387,9 @@
   /**
    * disable speed control and move the servos at full speed.
    */
+  @Deprecated
   void fullSpeed();
+  
+  void setMaxSpeed();
```
kwatters: please remove this method .. if we want setMaxSpeed , it should take the value of the max speed to be set.. this value should be honored by the servo speed control slider in the webgui and by the TimerEncoder.

PR #1048, opened 2022-02-12
Python Config

supertick:


src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1476,7 +1474,7 @@ synchronized public static void unregister(String inName) {
       }
 
       if (updatedServiceLists) {
-        runtime.broadcast("publishInterfaceToPossibleServices");
+        runtime.invoke("publishInterfaceToPossibleServices");
```
kwatters, line 1477: this isn't part of the service lifecycle publisher / listener interfaces.. if we are putting it here.. it should be defined in the interface also.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3660,8 +3652,16 @@ static public ServiceConfig load(String name) throws IOException {
 
     ServiceConfig config = CodecUtils.readServiceConfig(filename);
     si.load(config);
+
+    if (runtime != null) {
+      runtime.invoke("publishConfigLoaded", name);
```
kwatters, line 3657: also.. this isn't part of the service lifecycle interfaces.

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -1320,4 +1322,24 @@ public static void main(String[] args) {
       log.error("main threw", e);
     }
   }
+
+  @Override
+  public void onCreated(String name) {
```
kwatters, line 1327: shouldn't there be some implementation here?  if not, why is WebGui implementing the ServiceLifeCycleListener?

src/main/java/org/myrobotlab/service/Runtime.java
supertick: this is not part of the ServiceLifeCycleListener interface and should not be

src/main/java/org/myrobotlab/service/WebGui.java
supertick: webgui has to keep the ui in sync with any new service, it does so by subscribing to lifecycle events..
it only needs the onStarted

if you remember synching the availability of services for attaching has gone through a couple iterations ...
I'm not positive that WebGui does not need the onStarted events but I didn't want to keep adding changes to this pr

src/main/java/org/myrobotlab/service/Runtime.java
supertick: it shouldn't be either - its currently an very optimized method that only is useful for the ui

PR #1058, opened 2022-02-20
Many Arduino and Pir Fixes

supertick:


src/main/java/org/myrobotlab/service/Arduino.java
```diff
@@ -471,6 +472,26 @@ public void reattach(DeviceMapping dm) {
...
+      // on reattach get back to its previous state enabled/disabled
+      if (attachable instanceof Pir) {
+        Pir pir = (Pir) attachable;
+        if (pir.isEnabled()) {
+          pir.enable();
+        }
+      }
+
+    } else if (attachable instanceof I2CControl) {
+
```
kwatters: no implementation here.. a log message should be added to say this isn't supported yet.

src/main/java/org/myrobotlab/service/Pir.java
```diff
@@ -88,6 +73,22 @@ public void attachPinArrayControl(PinArrayControl control) {
...
+  @Override
+  public void detach(String name) {
+    ServiceInterface si = Runtime.getService(name);
+    if (si instanceof PinArrayControl) {
+      // FIXME - problem - what if someone else is using this pin ?
+      // FIXME - should disable in the context of this service's name
+      ((PinArrayControl) si).disablePin(pin);
+      detachPinArrayControl((PinArrayControl) si);
+    } else {
+      error("do not know how to attach to %s of type %s", name, si.getSimpleName());
```
kwatters: log message is wrong.. this is "detach" not "attach"

src/main/java/org/myrobotlab/service/Pir.java
```diff
@@ -18,25 +17,21 @@
...
   private static final long serialVersionUID = 1L;
 
-  boolean isActive = false;
+  String controllerName;
+
+  /**
+   * yep there are 3 states to a binary sensor true/false .... and unknown - we
+   * start with "unknown"
+   */
+  Boolean isActive = null;
```
kwatters: This should be a primitive to prevent it from being null.

kwatters: Also please rename "isActive" to just "active" ...   "is" prefix should be reserved for the getter method for a boolean of isActive()

src/main/java/org/myrobotlab/service/Arduino.java
```diff
@@ -471,6 +472,26 @@ public void reattach(DeviceMapping dm) {
       if (servo.isEnabled()) {
         msg.servoAttachPin(dm.getId(), pin);
       }
+    } else if (attachable instanceof UltrasonicSensorControl) {
+      // reattach logic
```
kwatters, line 477: no implementation of reattach for ultrasonic sensor.. please add log warning message or remove this branch of the if statement.

PR #1059, opened 2022-02-22
Grog gestures random 1

supertick:


src/main/java/org/myrobotlab/service/Pir.java
```diff
@@ -18,25 +17,23 @@
...
-  boolean isActive = false;
+  String controllerName;
+
+  /**
+   * yep there are 3 states to a binary sensor true/false .... and unknown - we
+   * start with "unknown"
+   */
+  Boolean active = null;
 
   boolean isEnabled = false;
```
kwatters: i know it's not changed in this PR.. but "isEnabled" should be renamed to just "enabled"

src/main/java/org/myrobotlab/service/Arduino.java
```diff
@@ -471,6 +472,26 @@ public void reattach(DeviceMapping dm) {
       if (servo.isEnabled()) {
         msg.servoAttachPin(dm.getId(), pin);
       }
+    } else if (attachable instanceof UltrasonicSensorControl) {
+      // reattach logic
```
kwatters: I think there needs to be a merge with develop ... there's a missing log message here.

PR #1067, opened 2022-03-12
Business logic of peers + config 

supertick:


src/main/java/org/myrobotlab/framework/Plan.java
```diff
@@ -1,34 +1,130 @@
...
+  }
+
+  public ServiceConfig addPeer(String name, ServiceConfig sc) {
+    config.put(name, sc);
+    return sc;
   }
 
-  public String get(String serviceName) {
-    return services.get(serviceName);
+  public ServiceConfig getPeer(String key) {
```
kwatters: should this method be removed?

src/main/java/org/myrobotlab/framework/interfaces/ServiceInterface.java
```diff
@@ -172,59 +167,7 @@
...
-        serviceConfig.type = type;
-        config.put(name, serviceConfig);
-      }
-
-    } catch (Exception e) {
-      log.error("ServiceConfig.getDefault({},{}) threw", name, type, e);
-    }
-    return config;
-  }
+  public Plan setPlan(Plan plan);
```
kwatters: add javadoc please.  it's not clear when this should be called, or that once you call it, you can't call it again..

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1405,7 +1341,7 @@ public ServiceConfig getConfig() {
   }
 
   public ServiceConfig load() throws IOException {
-    ServiceConfig config = Runtime.load(getName());
+    ServiceConfig config = Runtime.load(null, getName(), getClass().getSimpleName());
```
kwatters: it's confusing to pass null here..  just rely on the load signature that doesn't require the plan first.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -302,19 +311,114 @@ static public String getPublicGateway() {
...
+      if ("runtime".equals(peerName)) {
+        // already handled runtime
+        // apply the config
+        runtime.apply(plan.get("runtime"));
+        continue;
+      }
+
+      ServiceConfig sc = plan.get(peerName);
+      if (sc.autoStart) {
+        ServiceInterface si = createService(peerName, sc.type, null);
```
kwatters: what's the 3rd argument here?  why is it null?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -345,7 +449,7 @@ public final static void createAndStartServices(List<String> services) {
 
         log.info("attempting to invoke : {} of type {}", name, type);
 
-        ServiceInterface s = Runtime.create(name, type);
+        ServiceInterface s = Runtime.create(null, name, type);
```
kwatters: confusing to pass a null plan.. omit the null in this call and rely on the String name, String type definition instead.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1904,46 +1975,89 @@ public void addRoute(String remoteId, String uuid, int metric) {
...
+    return start(null, name, type);
+  }
 
-    return null;
+  static public ServiceInterface start(String name) {
+    return start(null, name, null);
   }
 
-  static public ServiceInterface start(String name, String type) {
+  static public ServiceInterface start(Plan inPlan, String name) {
```
kwatters: why is the argument called "inPlan"  ? why not just "plan"  ?  add javadoc to describe why you've called it "in"Plan

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1904,46 +1975,89 @@ public void addRoute(String remoteId, String uuid, int metric) {
...
+   * </pre>
+   * 
+   * if name and type are null - then process whatever is currently in the Plan
+   * 
+   * @param name
+   * @param type
+   *          - can be null, if file exists on the file system or already loaded
+   * @return
+   */
+  static public ServiceInterface start(Plan inPlan, String name, String type) {
```
kwatters: explain why not just "plan" ? why called inPlan?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1904,46 +1975,89 @@ public void addRoute(String remoteId, String uuid, int metric) {
...
     }
 
+    // clear the plan
+    Runtime.clear();
+
     return si;
   }
 
+  public static ServiceConfig load(String name, String type) {
+    return load(null, name, type);
```
kwatters: don't pass null as the first arg.. that's confusing.. if you're not passing a plan here. .just call load(name,type)

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1904,46 +1975,89 @@ public void addRoute(String remoteId, String uuid, int metric) {
...
+  }
+
+  public static ServiceConfig load(String name) {
+    return load(null, name, null);
```
kwatters: dont' pass null's ..  call  a load(String name) signature if that's what you want.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3601,7 +3659,7 @@ public String setConfigDir(String dir) {
   static public ServiceInterface loadAndStart(String name, String type) {
     ServiceInterface s = null;
     try {
-      s = create(name, type);
+      s = create(null, name, type);
```
kwatters: dont' pass null as the first arg.. use create(name, type) ...

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3639,34 +3697,91 @@ public ServiceConfig getConfig() {
...
+
+    if (inPlan == null) {
+      // root plan
+      plan = new Plan(runtime.serviceType);
+    } else {
+      plan = inPlan;
+    }
+
+    if (name == null && type == null) {
+      log.info("no new name or type information provided - we will use existing plan");
```
kwatters, line 3745: double log message.. tighten up.

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3639,34 +3697,91 @@ public ServiceConfig getConfig() {
...
 
-    if (runtime != null) {
-      runtime.invoke("publishConfigLoaded", name);
+    if (sc == null) {
+      // no file based yml config
+      if (type == null) {
+        error("%s could not load type %s", name, type);
+        return null;
+      } else {
+        log.info("could not find config in file trying default of type {}", type);
```
kwatters, line 3784: should this be a warning/error?  the log message tends to indicate some problem.. maybe re-word the message so it doesn't sound like an error

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3824,12 +3878,12 @@ public boolean save() {
     return false;
   }
 
-  public void setConfigName(String configName) {
+  public String setConfigName(String configName) {
     if (configName == null || configName.trim().length() == 0) {
```
kwatters, line 3899: StringUtils.isEmpty is probably a bit cleaner than this if test here.

src/main/java/org/myrobotlab/service/config/ServoConfig.java
```diff
@@ -2,18 +2,18 @@
 
 public class ServoConfig extends ServiceConfig {
 
-  public boolean autoDisable;
+  public boolean autoDisable = false;
```
kwatters: not related to this PR.. this should have been done on develop... bad grog!

src/main/java/org/myrobotlab/framework/Plan.java
supertick: yeah ... i de-normalized when I shouldn't have - so peers has some more work to do 
the following will still be true
Runtime is in charge of a plan
the plan says what to create and start
its the "only" plan

src/main/java/org/myrobotlab/framework/interfaces/ServiceInterface.java
supertick: this method is going to go away

src/main/java/org/myrobotlab/framework/Service.java
supertick: passing in the "plan" is going to go away - the parameter will be removed

src/main/java/org/myrobotlab/service/Runtime.java
supertick: id of the mrl instance ... its "always" been there, since first tried to do distributed mrl
null == local instance

src/main/java/org/myrobotlab/service/Runtime.java
supertick: yeah - its going away

src/main/java/org/myrobotlab/service/Runtime.java
supertick: this commit - peers needed to send in their plan
removing this ..
there is only "one" plan

src/main/java/org/myrobotlab/service/Runtime.java
supertick: this commit - peers needed to send in their plan
removing this ..
there is only "one" plan

src/main/java/org/myrobotlab/service/Runtime.java
supertick: this commit - peers needed to send in their plan
removing this ..
there is only "one" plan

src/main/java/org/myrobotlab/service/Runtime.java
supertick: this commit - peers needed to send in their plan
removing this ..
there is only "one" plan

src/main/java/org/myrobotlab/service/Runtime.java
supertick: this commit - peers needed to send in their plan
removing this ..
there is only "one" plan

src/main/java/org/myrobotlab/service/Runtime.java
supertick: fixed

src/main/java/org/myrobotlab/service/Runtime.java
supertick: its an error - couldn't get type information for the requested service

src/main/java/org/myrobotlab/service/Runtime.java
supertick: we don't pack StringUtils as part of base jar

src/main/java/org/myrobotlab/service/config/ServoConfig.java
supertick: heh ... hotfix fallout ..
the clip boolean wasn't set ...
come on 1 line ! ;) ... can I hotfix it too ?  if i back it out I'll forget to explicitly set

PR #1069, opened 2022-03-25
the javacv frame converter object isn't thread safe

kwatters:
removing the close call for the frame converters to avoid thread safety issues.

if you reuse the same frame converter .. it will null out the previously converted mat/frame/iplimage... the result is a hard to find null pointer address exception from javacv...


src/main/java/org/myrobotlab/opencv/OpenCVFilterSolr.java
```diff
@@ -58,12 +58,11 @@
...
 
-  private CvFont font = cvFont(CV_FONT_HERSHEY_PLAIN);
-  private CvFont fontWarning = cvFont(CV_FONT_HERSHEY_PLAIN);
+  transient private CvFont font = cvFont(CV_FONT_HERSHEY_PLAIN);
 
   private String formattedSearchResult = "No Result";
 
-  private Solr solr = null;
+  transient private Solr solr = null;
   private String solrUrl = "http://localhost:8983/solr/wikipedia";
```
supertick, line 66: good catch on the new transients,
I would like to see publishing or sending msgs
e.g. send(solar, 'doSomethingSolr', params) vs a direct reference
solar in this case being a String fieldname

supertick (approved):

src/main/java/org/myrobotlab/opencv/OpenCVFilterSolr.java
kwatters: future pr.. this pr was more about making sure all the filters are serializable and to fix places where the frame converters don't behave as expected.

PR #1074, opened 2022-04-02
Grog config 8

supertick:


src/main/java/org/myrobotlab/service/abstracts/AbstractSpeechSynthesis.java
```diff
@@ -1139,22 +1134,24 @@ public void attachSpeechControl(SpeechSynthesisControl control) {
...
-
-  // Hacky way to get normalized code to work - inheritance not so helpful here
-  public SpeechSynthesisConfig getConfig(SpeechSynthesisConfig config) {
-    config.mute = mute;
-    config.blocking = blocking;
+  
+  
+  @Override
+  public ServiceConfig getConfig() {
+    SpeechSynthesisConfig c = (SpeechSynthesisConfig)config;
```
kwatters, line 1141: why don't the specific services have their own subclass of speech synthesis config?  When loading this.. it seems like this will serialize out as a generic SpeechSynthesisConfig.. instead of the specific config type for the specific service...

src/main/java/org/myrobotlab/service/config/SpeechSynthesisConfig.java
```diff
@@ -4,10 +4,11 @@
 
 public class SpeechSynthesisConfig extends ServiceConfig {
```
kwatters, line 5: this class should probably be abstract to make sure no one instantiates it .. it should always have a subclass instance for the specific service that is using it.

supertick: If I was lazy .... (or smarter) ... I wouldn't have had all the subclassed config  - at the moment all the subclasses do nothing..
because it all handled in the SpeeechSynthesisConfig ... but I was scared and did the stupid thing of subclassing it all, without derived differences

PR #1081, opened 2022-04-04
Grog worky save and start

supertick:


src/test/java/org/myrobotlab/service/RuntimeProcessTest.java
```diff
@@ -82,7 +81,8 @@ public void cliTest() throws Exception {
     assertTrue("testing re-entrant - expecting runtime service", Arrays.toString(Runtime.getServiceNames()).contains("runtime"));
     
     // removing all 
-    Runtime.releaseAll(true, true);
+    Runtime.releaseAll();
+    sleep(100);
```
kwatters, line 85: Why is this sleep needed?

src/test/java/org/myrobotlab/test/AbstractTest.java
```diff
@@ -152,7 +152,9 @@ public static void releaseServices() {
     log.warn("end of test - id {} remaining services {}", Platform.getLocalInstance().getId(), Arrays.toString(Runtime.getServiceNames()));
     
     // release all including runtime - be careful of default runtime.yml
-    Runtime.releaseAll(true, true);
+    // Runtime.releaseAll(true, true);
+    Runtime.releaseAll();
+    sleep(100);
```
kwatters, line 157: why is this sleep needed/added?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -329,20 +333,18 @@ static public synchronized ServiceInterface create(String name) {
   // there is no point ... and it just makes it more complicated, if you want to
   // adjust
   // configuration adjust config in the plan before starting
-  static public synchronized ServiceInterface create(String name, String type) {
+  static public synchronized ServiceInterface create(String configName, String name, String type) {
```
kwatters, line 336: should the be private?

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1736,7 +1737,7 @@ public void setThisThread(Thread thisThread) {
 
   public ServiceInterface startPeer(String reservedKey) {
     String actualName = getPeerName(reservedKey);
-    return Runtime.start(actualName);
+    return Runtime.start(actualName, null);
```
kwatters, line 1740: what does it mean to pass a null type here?

PR #1082, opened 2022-04-06
Grog worky save and start 2

supertick:


src/test/java/org/myrobotlab/test/AbstractTest.java
```diff
@@ -152,7 +152,9 @@ public static void releaseServices() {
     log.warn("end of test - id {} remaining services {}", Platform.getLocalInstance().getId(), Arrays.toString(Runtime.getServiceNames()));
     
     // release all including runtime - be careful of default runtime.yml
-    Runtime.releaseAll(true, true);
+    // Runtime.releaseAll(true, true);
+    Runtime.releaseAll();
+    sleep(100);
```
kwatters, line 157: Why is this sleep statement here?  what fails without it?

src/test/java/org/myrobotlab/service/RuntimeProcessTest.java
```diff
@@ -82,7 +81,8 @@ public void cliTest() throws Exception {
     assertTrue("testing re-entrant - expecting runtime service", Arrays.toString(Runtime.getServiceNames()).contains("runtime"));
     
     // removing all 
-    Runtime.releaseAll(true, true);
+    Runtime.releaseAll();
+    sleep(100);
```
kwatters, line 85: Why is this sleep statement needed?  what fails without it?

src/main/java/org/myrobotlab/opencv/ImageFileFrameGrabber.java
```diff
@@ -21,14 +24,15 @@
...
+  protected transient Mat image;
+  transient private Mat lastImage;
 
   // transient private HashMap<String, IplImage> cache = new HashMap<String,
   // IplImage>();
   private int frameCounter = 0;
   public String path;
-  transient OpenCVFrameConverter.ToIplImage converter = new OpenCVFrameConverter.ToIplImage();
+  // transient OpenCVFrameConverter.ToIplImage converter = new OpenCVFrameConverter.ToIplImage();
+  OpenCVFrameConverter.ToMat converter = new OpenCVFrameConverter.ToMat();
```
kwatters: Use the new CloseableFrameConverter please and make sure it's closed in the release method.

src/main/java/org/myrobotlab/opencv/ImageFileFrameGrabber.java
```diff
@@ -21,14 +24,15 @@
   private ArrayList<File> imageFiles = new ArrayList<File>();
   private int grabCount = 0;
 
-  protected transient IplImage image;
-  transient private IplImage lastImage;
+  protected transient Mat image;
+  transient private Mat lastImage;
```
kwatters: Why does the a frame grabber need a handle to the lastImage?

PR #1085, opened 2022-04-07
misc updates

supertick:


src/main/java/org/myrobotlab/service/interfaces/EncoderControl.java
```diff
@@ -23,6 +23,8 @@
    * @return encoder data
    */
   public EncoderData publishEncoderData(EncoderData data);
+  
+  public void attachEncoderController(EncoderController controller) throws Exception;
```
kwatters: Why does this throw an exception?

PR #1086, opened 2022-04-13
Service Cleanup !

supertick:


src/main/java/org/myrobotlab/service/Tracking.java
```diff
@@ -255,6 +255,9 @@ public Stats publishStats() {
 
   public void disable() {
     state = TrackingState.IDLE;
+    for (String filter : trackingFilters) {
+      send(cv, "removeFilter", filter);
```
kwatters: why remove?  why not just disable the filters?

src/main/java/org/myrobotlab/service/interfaces/TextListener.java
```diff
@@ -6,6 +6,24 @@ public interface TextListener extends NameProvider {
 
   public void onText(String text) throws Exception;
 
-  public void attachTextPublisher(TextPublisher service) throws Exception;
+  /**
+   * Attach a text listener
+   * 
+   * @param service
+   */
+  default public void attachTextPublisher(TextPublisher service) {
```
kwatters, line 14: This is backwards, we agreed that publishers would attach listeners.  Please clean this up to make it consistent with the design pattern for UtterancePublisher / UtteranceListener.

kwatters, line 14: This is backwards, we agreed that publishers would attach listeners.  Please clean this up to make it consistent with the design pattern for UtterancePublisher / UtteranceListener.

src/main/java/org/myrobotlab/service/AzureTranslator.java
```diff
@@ -12,96 +12,118 @@
...
-    }
-  }
+  /**
+   * location
+   */
+  String location = "eastus";
 
-  public AzureTranslator(String n, String id) {
-    super(n, id);
+  transient OkHttpClient client = new OkHttpClient();
```
kwatters, line 56: Why can't we just use apache httpclient?  Seems silly to include yet another http client library for a straight forward http request.

src/main/java/org/myrobotlab/service/config/ImageDisplayConfig.java
```diff
@@ -0,0 +1,59 @@
...
+
+  
+    public static class Display{
+      // problematic - make transient
+      String name;
+      
+      public Integer x, y;
+      public Integer width, height;
+      
+      public Boolean fullscreen = null;
```
kwatters: this should have a default value.. we should avoid nulls as defaults.

src/main/java/org/myrobotlab/service/interfaces/ImagePublisher.java
```diff
@@ -0,0 +1,33 @@
...
+   */
+ public ImageData publishImage(ImageData data);
+ 
+ default public void attachImageListener(ImageListener display) {
+   attachImageListener(display.getName());
+ }
+ 
+ // Default way to attach an image listener so implementing classes need 
+ // not worry about these details.
+ default public void attachImageListener(String name) {
```
kwatters, line 24: for completeness, please add detach/remove ImageListener.

src/main/java/org/myrobotlab/service/interfaces/SearchPublisher.java
```diff
@@ -13,7 +14,7 @@ public interface SearchPublisher extends TextPublisher, LocaleProvider {
 
   SearchResults publishResults(SearchResults results);
```
kwatters, line 18: please specify which methods are published.. and add the default implementation of the attach/remove SearchListener

kwatters, line 18: please specify which methods are published.. and add the default implementation of the attach/remove SearchListener

src/main/java/org/myrobotlab/service/interfaces/TextPublisher.java
```diff
@@ -1,22 +1,48 @@
...
   /**
-   * the root for any other attachTextListener type
+   * Default way to attach an utterance listener so implementing classes need
+   * not worry about these details.
    * 
    * @param name
-   *          - the name of the text listener
    */
-  public void attachTextListener(String name);
+  default public void attachTextListener(String name) {
```
kwatters, line 35: please add removeTextListener for completeness of this interface.

kwatters, line 35: please add removeTextListener for completeness of this interface.

src/main/java/org/myrobotlab/service/config/ImageDisplayConfig.java
supertick: config potentially can have 3 values - the semantics are simple
true config force true behavior
false config force false behavior
null config do not influence current behavior
so concrete example - if a image is currently being shown, and we don't want to change this display attribute, leave it null

src/main/java/org/myrobotlab/service/interfaces/TextListener.java
supertick: No, I disagree  ...  if appropriate pub/sub is followed bi-directional attachment is trivial.
This interface implements the correct pub/sub call 
send(name, "attachTextListener", getName());
It works well without the burden of the user having to know which is the publisher and which is the user.
I would prefer you update your utterance listeners to do the same.

supertick: No, I disagree  ...  if appropriate pub/sub is followed bi-directional attachment is trivial.
This interface implements the correct pub/sub call 
send(name, "attachTextListener", getName());
It works well without the burden of the user having to know which is the publisher and which is the user.
I would prefer you update your utterance listeners to do the same.

src/main/java/org/myrobotlab/service/interfaces/TextPublisher.java
supertick: you probably mean detachTextListener .. but ya I agree

src/main/java/org/myrobotlab/service/Tracking.java
supertick: hmm ... I agree

src/main/java/org/myrobotlab/service/AzureTranslator.java
supertick: because okhttp is the better library  (also more cloud services are providing examples with okhttp)

https://stackoverflow.com/questions/42392778/okhttp-or-httpclient-which-offers-better-functionality-and-more-efficiency

src/main/java/org/myrobotlab/service/Tracking.java
supertick: updated

src/main/java/org/myrobotlab/service/interfaces/ImagePublisher.java
supertick: good call - updated

src/main/java/org/myrobotlab/service/interfaces/SearchPublisher.java
supertick: good call - updated

src/main/java/org/myrobotlab/service/interfaces/TextPublisher.java
supertick: not part of my pr ... but I'll do the penance because i really like the pattern :)
updated

supertick: updated

src/main/java/org/myrobotlab/service/interfaces/TextListener.java
kwatters: sorry.. we agreed directionality is important and we should have a convention.  this has ramifications in the config generation...  publishers are supposed to attach listeners in their configs.

kwatters: ramifications for config need to be acknowledged.  there should be convention , and that should be the convention that we agreed upon many moons ago.

src/main/java/org/myrobotlab/service/config/ImageDisplayConfig.java
kwatters: so long as getConfig doesn't provide any nulls in the config and fully populates all values with non null values , i'm ok with it.

src/main/java/org/myrobotlab/service/interfaces/SearchPublisher.java
kwatters: still is missing the publishMethods and the attachSearchListener / removeSearchListener methods.

src/main/java/org/myrobotlab/service/interfaces/ImagePublisher.java
kwatters: I don't see the removeImageListener method added...

supertick: correct - because removeImageListener isn't standard
detachImageListener is

src/main/java/org/myrobotlab/service/interfaces/SearchPublisher.java
supertick: search replace error - fixed

supertick: search replace error corrected

src/main/java/org/myrobotlab/service/Tracking.java
kwatters: This still says  "removeFilter" .. did you forget to check in some changes?

src/main/java/org/myrobotlab/service/interfaces/SearchPublisher.java
kwatters: I still don't see the publishMethods / attachSearchListener / removeSearchListener methods.  Did you miss some files when you checked them in?

src/main/java/org/myrobotlab/service/Tracking.java
supertick: yeah ... noticed i didn't get all the updates I thought I did - will take me a little to organize

supertick: updated

src/main/java/org/myrobotlab/service/interfaces/TextListener.java
supertick: configs are different than calling methods
publisher.attach(listener) works just as well as listener.attach(publisher)
publisher is the only one saving the attachment as config
all listener does is send a message through the default interface to add itself to the publisher
there is no ambiguity, or confusion - its simple and all implemented as pub/sub message sending

supertick: configs are different than calling methods
publisher.attach(listener) works just as well as listener.attach(publisher)
publisher is the only one saving the attachment as config
all listener does is send a message through the default interface to add itself to the publisher
there is no ambiguity, or confusion - its simple and all implemented as pub/sub message sending

src/main/java/org/myrobotlab/service/config/ImageDisplayConfig.java
supertick: updated to boolean, Integer x, y, width height as null means in the center of the screen as the actual image size

src/main/java/org/myrobotlab/service/interfaces/SearchPublisher.java
supertick: you should see it now ..

PR #1087, opened 2022-04-13
Shared peers not released

supertick:


src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -368,6 +368,7 @@ private static ServiceInterface createServicesFromPlan(String configName, String
       return null;
     }
 
+    Set<String> autoStartedPeers = new HashSet<>();
```
kwatters, line 371: why is this defined locally if it's also defined on the base Service class?  that's confusing.

src/main/java/org/myrobotlab/service/Wikipedia.java
```diff
@@ -188,6 +185,10 @@ public int setMaxImages(int cnt) {
     return cnt;
   }
 
+  @Override
+  public void attach(String serviceName) throws Exception {
+    attachImageListener(serviceName);
```
kwatters: shouldn't this do a type check to see that the service being attached is an image listener?  is that the only thing that wikipedia publishes/ can attach ?  what if it's a search publisher or a text publisher being attached?  what if it's 2 out of 3 of those... or all 3?

src/main/java/org/myrobotlab/service/Runtime.java
supertick: because when Runtime is requested to start a composite service that has peers, the current contract is that it starts the peers first "before" the composite service is started.   You can't use si.addAutoStartedPeer(peerName) "before" si is created

src/main/java/org/myrobotlab/service/Wikipedia.java
supertick: updated

PR #1089, opened 2022-04-23
Peer config binding

supertick:
Composite service now know their peer's state started vs released
Additionally, they correctly can find their peer's actual name from the peerKey
This pr also contains what will be a very useful publishPeerStarted(peerKey) and publishPeerReleased(peerKey) so scripts or services can handle details when a peer is started or released

### DIFF
--- a/src/main/java/org/myrobotlab/codec/CodecUtils.java
+++ b/src/main/java/org/myrobotlab/codec/CodecUtils.java
@@ -8,6 +8,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
+import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
@@ -750,6 +751,18 @@ public class CodecUtils {
     return (ServiceConfig)yaml.load(data);
   }
 
+  public static void setField(Object o, String field, Object value) {
+    try {
+      // TODO - handle all types :P
+     Field f =  o.getClass().getDeclaredField(field);
+     f.setAccessible(true);
+     f.set(o, value);
+    } catch (Exception e) {
+      /** don't care - if its not there don't set it */
+    }
+  }
+  
+  
   public static void main(String[] args) {
     LoggingFactory.init(Level.INFO);
 
@@ -770,4 +783,7 @@ public class CodecUtils {
     }
   }
 
+  
+  
+  
 }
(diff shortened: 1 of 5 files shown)


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1737,14 +1738,35 @@ public void setThisThread(Thread thisThread) {
...
+    String actualName = getPeerName(peerKey);
+    ServiceInterface si = Runtime.start(actualName, null);
+    if (si != null) {
+      ServiceReservation sr = serviceType.getPeer(peerKey);
+      if (sr != null) {
+        sr.state = "STARTED";
+        sr.actualName = actualName;
+        sr.type = si.getSimpleName();
+        if (si != null) {
+          CodecUtils.setField(this, peerKey, si);
```
kwatters, line 1751: why does this need to do fancy introspection ?  I'm missing something that is causing this additional level of complexity.

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -750,6 +751,18 @@ public static ServiceConfig readServiceConfig(String filename) throws IOExceptio
     return (ServiceConfig)yaml.load(data);
   }
 
+  public static void setField(Object o, String field, Object value) {
```
kwatters, line 754: Seems like you'll always pass a Service class in here instead of object... what's the rational for the fancy  reflection going on below?

src/main/java/org/myrobotlab/framework/Service.java
supertick: good question :)
It establishes a convention and contract ... 
If a service has a field named the same as a peer 
it will be set when the parent is started ...

Already, the convention/contract was established that peers start before their parents.
there is a bunch of npe's and various manual conventions that were established which tried to assign the reference of a peer to a class member.

This takes the random developers implementation out of the mix.

Rule is:
* if you went through the effort to define a peer for a service, and you have a local reference named the same thing
e.g. transient SerialDevice serial ... then when the parent is started, the framework assigns the reference

src/main/java/org/myrobotlab/codec/CodecUtils.java
supertick: utilities should be generalized, and not dictate the way they are used

src/main/java/org/myrobotlab/framework/Service.java
kwatters: fancy...

PR #1092, opened 2022-04-23
Polly with ssml support

supertick:


src/main/java/org/myrobotlab/service/config/PollyConfig.java
```diff
@@ -2,4 +2,12 @@
 
 public class PollyConfig extends SpeechSynthesisConfig {
 
+  
+  public boolean ssml = false;
+  public boolean autoDetectSsml = true;
+  
+  public PollyConfig() {
+    voice = "Brian";
```
kwatters, line 10: seems odd to default this here? what's that rational?

kwatters: merging anyway..  but we should clean this up.. i don't understand why we have a specific default being set here.

PR #1093, opened 2022-04-23
InMoov refactor

supertick:
massive refactor of inmoov .. not "complete" but a good reduction of code


pom.xml
```diff
@@ -351,6 +351,15 @@
   </dependency>
 <!-- Elasticsearch end -->
 
+<!-- FiniteStateMachine begin -->
```
kwatters, line 348: This doesn't appear to be in the FiniteStateMachineMeta.java...  
Please generate the pom file and make sure all changes are updated in the appropriate places.

pom.xml
```diff
@@ -1083,25 +1093,25 @@
   <dependency>
     <groupId>com.fasterxml.jackson.core</groupId>
     <artifactId>jackson-core</artifactId>
-    <version>2.10.1</version>
+    <version>2.12.6</version>
```
kwatters: I don't see this updated in the PollyMeta or the WikiDeataFetcherMeta files..  please generate the pom.

pom.xml
supertick: the metas were updated and regenned in a later pr

pom.xml
supertick: the metas were updated and regenned in a later pr

kwatters:
I'm getting a null pointer exception when I click on the inmoov button from the Intro service.

```10:25:40.904 [webgui-incoming-msg-queue] ERROR class org.myrobotlab.service.meta.abstracts.MetaData - getting meta data failed on org.myrobotlab.service.meta.InMoov2Meta
java.lang.NullPointerException: null
	at org.myrobotlab.service.meta.InMoov2Meta.getDefault(InMoov2Meta.java:109) ~[myrobotlab.jar:1.1.202205191915]
	at org.myrobotlab.service.meta.abstracts.MetaData.getDefault(MetaData.java:345) [myrobotlab.jar:1.1.202205191915]
	at org.myrobotlab.service.Runtime.loadService(Runtime.java:3868) [myrobotlab.jar:1.1.202205191915]
	at org.myrobotlab.service.Runtime.loadService(Runtime.java:2155) [myrobotlab.jar:1.1.202205191915]
	at org.myrobotlab.service.Runtime.create(Runtime.java:342) [myrobotlab.jar:1.1.202205191915]
	at org.myrobotlab.service.Runtime.startInternal(Runtime.java:2132) [myrobotlab.jar:1.1.202205191915]
	at org.myrobotlab.service.Runtime.start(Runtime.java:2092) [myrobotlab.jar:1.1.202205191915]
```

src/main/java/org/myrobotlab/service/Polly.java
```diff
@@ -185,6 +190,7 @@ public String[] getKeyNames() {
    */
   @Override
   public AudioData generateAudioData(AudioData audioData, String toSpeak) throws IOException {
+    PollyConfig c = (PollyConfig) config;
```
kwatters: you shouldn't reference the config from here.  instead when you are loading/applying the config, pull the appropriate values into member variables that you can reference .. "autoDetectSsml" in particular, should be a member variable.. and you won't need the config here.  even further, the config object shouldn't exist outside of the load/apply method.  i would prefer to not see it exist at all as a member on the service classes.

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -596,82 +486,55 @@ public String get(String key) {
...
   public InMoov2Arm getArm(String side) {
-    if ("left".equals(side)) {
-      return leftArm;
-    } else if ("right".equals(side)) {
-      return rightArm;
-    } else {
-      log.error("can not get arm {}", side);
-    }
-    return null;
+    return (InMoov2Arm) getPeer(side + "Arm");
```
kwatters, line 493: there's no error checking that "side' value is valid here anymore.

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -596,82 +486,55 @@ public String get(String key) {
...
   public InMoov2Hand getHand(String side) {
-    if ("left".equals(side)) {
-      return leftHand;
-    } else if ("right".equals(side)) {
-      return rightHand;
-    } else {
-      log.error("can not get arm {}", side);
-    }
-    return null;
+    return (InMoov2Hand) getPeer(side + "Hand");
```
kwatters, line 501: no error checking that "side" is a valid value here anymore.

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -596,82 +486,55 @@ public String get(String key) {
...
 
-    if (rightArm != null) {
-      lastActivityTime = Math.max(lastActivityTime, rightArm.getLastActivityTime());
-    }
+      Long lastActivityTime = 0L;
 
-    if (head != null) {
-      lastActivityTime = Math.max(lastActivityTime, head.getLastActivityTime());
-    }
+      Long head = (Long) sendToPeerBlocking("head", "getScript", getName());
```
kwatters: what is the reason for sending a blocking call for "getScript" here ?  This seems odd/wrong .. the point of this method is to see the last time there was activity.

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -596,82 +486,55 @@ public String get(String key) {
...
-    if (lastPirActivityTime != null) {
-      lastActivityTime = Math.max(lastActivityTime, lastPirActivityTime);
-    }
+      return lastActivityTime;
 
-    if (lastActivityTime == 0) {
-      error("invalid activity time - anything connected?");
-      lastActivityTime = System.currentTimeMillis();
+    } catch (Exception e) {
+      error(e);
```
kwatters, line 534: add an actual error message if we actually need this exception handling.  what would throw this?

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -841,47 +638,24 @@ public String captureGesture(String gestureName) {
...
-
-    if (leftHand != null) {
-      script.append(indentSpace);
-      script.append(leftHand.getScript(getName()));
-    }
-    if (rightHand != null) {
-      script.append(indentSpace);
-      script.append(rightHand.getScript(getName()));
-    }
+    try {
```
kwatters, line 655: this should be replaced with a pose from the ServoMixer...

src/main/java/org/myrobotlab/service/Polly.java
```diff
@@ -203,31 +219,43 @@ public AudioData generateAudioData(AudioData audioData, String toSpeak) throws I
     return audioData;
   }
 
+  public boolean setSsml(boolean ssml) {
+    PollyConfig c = (PollyConfig) config;
```
kwatters: this shouldn't reference the config here...  please remove the config as a member for these classes.

src/main/java/org/myrobotlab/service/Polly.java
```diff
@@ -203,31 +219,43 @@ public AudioData generateAudioData(AudioData audioData, String toSpeak) throws I
...
   }
 
+  public boolean setSsml(boolean ssml) {
+    PollyConfig c = (PollyConfig) config;
+    c.ssml = ssml;
+    return ssml;
+  }
+
+  public boolean setAutoDetectSsml(boolean ssml) {
+    PollyConfig c = (PollyConfig) config;
```
kwatters: this shouldn't reference config here.. please remove config as a member from these classes.

supertick: The more variables you make the more error prone it becomes, if I was to pull it out of config then I'd still need to maintain it in config, and additionally copy it into then copy it out of that config.   Your recommending more maintenance in the future, which I am against.

src/main/java/org/myrobotlab/service/Polly.java
supertick: The more variables you make the more error prone it becomes, if I was to pull it out of config then I'd still need to maintain it in config, and additionally copy it into then copy it out of that config. Your recommending more maintenance in the future, which I am against.

src/main/java/org/myrobotlab/service/Polly.java
supertick: configuration affects behavior
and denormalization you want to avoid, because it costs maintenance to keep things in sync.
Why not use the same configuration object to affect behavior directly.
This way the mindless copy out from the service, and copy into the service when applying can be avoided

src/main/java/org/myrobotlab/service/InMoov2.java
supertick: fixed

src/main/java/org/myrobotlab/service/InMoov2.java
supertick: fixed

src/main/java/org/myrobotlab/service/InMoov2.java
supertick: copy paste error - good catch thanks ..

supertick: updated to getLastActivityTime

src/main/java/org/myrobotlab/service/InMoov2.java
supertick: send blocking potentially can timeout

supertick: error(e) should be enough - it should give a message and a stack trace

src/main/java/org/myrobotlab/service/InMoov2.java
supertick: yeah ... i don't like the script processing in InMoov - but wasn't ready to do that change yet...

PR #1101, opened 2022-06-04
MqttBroker Config

supertick:


src/main/java/org/myrobotlab/service/MqttBroker.java
```diff
@@ -654,5 +656,25 @@ public void setKey(String keyName, String keyValue) {
...
+    c.wsPort = wsPort;
+    c.username = username;
+    c.password = password;
+    return c; 
+  }
+
+  
+  @Override
+  public ServiceConfig apply(ServiceConfig c) {
+    MqttBrokerConfig config = (MqttBrokerConfig) c;
```
kwatters, line 674: You don't seem to be applying any configuration here... i'd expect that you should be setting the address / wsport / mqttport / username / password here...

PR #1124, opened 2022-07-25
Add Proxy service generation for external services

### LINKED ISSUE #1123: Add dynamic proxy generation for remote services
In order to support services that only exist in external processes, we need to add support for dynamically generating a proxy class if we don't know about a remote class being registered.

Such functionality will require several enhancements:
1. A new subclass of `Registration` called `ExternalRegistration` that contains additional information such as implemented interfaces and a list of public methods. Need to investigate whether the deserializer can handle such a situation
2. Enable `Runtime.register()` to determine whether the service being registered is known
3. If the service is not known, attempt to cast to `ExternalRegistration` and use the contained information to generate a proxy using `java.lang.reflect.Proxy`.
4. The proxy will require an `InvocationHandler`; thus, we need a handler class that forwards all method calls to the remote service. We can optionally derive some method implementations such as the `ServiceInterface` methods

An alternative would be creating a `Service` class called `ProxyService` that overrides multiple methods to forward to the remote service. This has several disadvantages, however, such as the inability to use dot notation to call remote service methods (which will not function well with python scripts), an inability to use the service object in place of an interface instance, and a requirement to treat the proxy service specially in several framework code paths.

**Additional context**
This issue is to aid in the implementation of `mrlpy` and other foreign language interfaces.

AutonomicPerfectionist:
This PR fixes #1123.

TODO:

- [ ] Include list of fields and/or public methods in Registration (can be done in separate PR)
- [x] Create a specific type key format for determining an external service's type. Currently use `py:` prefix since it is not a valid Java class or package identifier. Once stabilized, can remove usage of `Class.forName()` for determining whether a type key is internal or external


src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -1,12 +1,14 @@
...
 
+
 import java.io.BufferedOutputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
+import java.lang.reflect.*;
```
supertick: please organize imports

src/main/java/org/myrobotlab/service/meta/RuntimeMeta.java
```diff
@@ -26,6 +26,9 @@ public RuntimeMeta() {
     // apache 2.0 license
     addDependency("com.google.code.gson", "gson", "2.8.5");
 
+    // for proxy generation
```
supertick, line 29: add license info in the comment // apache 2.0 license & ASM 3 clause BSD
:thinking: we should have a field in the addDependency to specify ... later ...

src/main/java/org/myrobotlab/framework/Registration.java
```diff
@@ -27,6 +28,14 @@ public class Registration {
   protected String name;
   protected String typeKey;
```
supertick, line 31: please add JavaDoc for name and typeKey

supertick (approved):

PR #1131, opened 2022-07-26
Add cached threadpool executor for service inbox processing

### LINKED ISSUE #1127: Use cached executor service for service inbox processing
Each service has a single thread dedicated to processing messages within its inbox. Using a single thread can introduce deadlocks if the processing of such a message requires a call to the service's own `Service.sendBlocking()` or other methods that require waiting for the completion of message processing.

Using a thread pool that creates new threads on-demand for inbox processing would alleviate the potential for deadlocks; however, this can introduce significant overhead. Using a caching thread pool executor should alleviate most of the overhead, but if many messages are received at once, the thread pool may grow abnormally large. Perhaps we could limit the caching thread pool to a maximum size.

An alternative would be redesigning a few key methods that cause deadlocks, so they no longer do. However, doing so would be a considerable effort and does not protect against unknown deadlock scenarios.

**Additional context**
Deadlocks were observed during the design of proxy services for #1124. Executing `Runtime.ls()` results in an invocation to each service's `getId()` method; in order to forward that to the remote service, the proxy must call `Runtime.getInstance().sendBlocking()`. However, doing so blocks the inbox processing thread (because it is processing `Runtime.ls()`) and results in a timeout.

AutonomicPerfectionist:
This PR fixes #1127 by executing inbox message processing on a dedicated executor service. The executor is constructed with a thread group dedicated to the inbox, and each new thread will have a name describing what it's doing (service name + "-inbox-worker-thread").

### DIFF
--- a/src/main/java/org/myrobotlab/framework/Service.java
+++ b/src/main/java/org/myrobotlab/framework/Service.java
@@ -1,7 +1,7 @@
 /**
- *                    
+ *
  * @author grog (at) myrobotlab.org
- *  
+ *
  * This file is part of MyRobotLab (http://myrobotlab.org).
  *
  * MyRobotLab is free software: you can redistribute it and/or modify
@@ -18,9 +18,9 @@
  * All libraries in thirdParty bundle are subject to their own license
  * requirements - please refer to http://myrobotlab.org/libraries for 
  * details.
- * 
+ *
  * Enjoy !
- * 
+ *
  * */
 
 package org.myrobotlab.framework;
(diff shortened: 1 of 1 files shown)


kwatters:
Won't this create a situation where messages could be processed out of order?  I think that is problematic.

AutonomicPerfectionist:
@kwatters yes that could happen. Are there any areas where message order is required/assumed? From what I could tell the only place where that would be is in the webgui client (haven't looked at it yet), everything else I've seen uses a sequencing approach where one message triggers another in the callback, or a sendBlocking()/services API call is used.

The reasoning behind the change is in the attached issue, if a thread pool is too concerning I can probably figure out an alternative solution

kwatters:
I would say that message ordering is very important.  Let's take an example like someone attaching and detaching a servo.  If those messages are processed out of order, then the servo will be left in the wrong state.

AutonomicPerfectionist:
@kwatters I managed to find an alternative solution, thanks for your advice!

Issue #1121, opened 2022-07-23
Tooling for code quality improvement


holgerfriedrich:
When I started to work with the MRL codebase using my favourite development IDE (MS Code), I found a few things which could be improved. There seem to be files with inconsistent line endings (like pom.xml), a lot of unused imports in Java files, etc.

I have some time during my robot is printed ;-), so I could support in cleaning things up, add automated tools like spotless, spotbugs/pmd, etc.

Would you be willing to merge such PRs?
WDYT?

AutonomicPerfectionist:
@holgerfriedrich that sounds great; I was thinking about cleaning up the code just the other day. I'm sure any PRs made would be gladly welcomed

PR #1134, opened 2022-07-27
Fix problems reported by spotbugs, refs #1121

holgerfriedrich:
just playing around with spotbugs, these are the first out of many findings....

### DIFF
--- a/src/main/java/org/myrobotlab/opencv/OpenCVData.java
+++ b/src/main/java/org/myrobotlab/opencv/OpenCVData.java
@@ -543,7 +543,7 @@ public class OpenCVData extends CvData {
 
   public PointCloud getPointCloud() {
     List<PointCloud> pcs = getPointCloudList();
-    if (pcs == null && pcs.size() != 0) {
+    if (pcs == null || pcs.size() == 0) {
       return null;
     }
     return pcs.get(0);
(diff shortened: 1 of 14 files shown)


src/main/java/org/myrobotlab/opencv/OpenCVData.java
```diff
@@ -543,7 +543,9 @@ public List<PointCloud> getPointCloudList() {
 
   public PointCloud getPointCloud() {
     List<PointCloud> pcs = getPointCloudList();
-    if (pcs == null && pcs.size() != 0) {
+    if (pcs == null) {
```
AutonomicPerfectionist: This could be done with a boolean OR instead (`pcs == null || pcs.size() == 0`), short circuiting will prevent any NPEs

AutonomicPerfectionist (changes requested):
Most of it looks pretty good to me! Just one conditional should probably be a short-circuited `or` to keep the related return paths together. Thanks for the contribution!

src/main/java/org/myrobotlab/opencv/OpenCVData.java
holgerfriedrich: thanks for reviewing, done

PR #1135, opened 2022-07-29
Add virtual class field to json to inform deserialization

### LINKED ISSUE #1130: Add class field for json to help in serialization/deserialization
Currently, MRL deserializes JSON messages by looking up the method in the method cache. However, this causes failures when using polymorphic types or when the method does not exist, as in #1128. Additionally, deserializing in external processes requires inferring the type by the field names, which is a brittle procedure.

I'd like to add an additional serializer/deserializer to the CodecUtils GSON object that injects a `class` field for all JSON objects. The deserializer can also use this field to fall back on if the method cache lookup fails. This will also enable much simpler deserialization in external processes.

This does require an update to the webgui client code, as the `class` field is currently unknown to it and causes errors while deserializing.

An alternative would be to keep the JSON as-is and add fallback paths to fix method cache lookup failures in #1128 

**Additional context**
#1124 currently requires the `class` field in json

AutonomicPerfectionist:
This PR adds custom serializers and deserializers for both GSON and Jackson that add a virtual `class` field to json recursively and then uses that field to deserialize polymorphic types correctly.

Fixes #1130 

Also adds documentation for most of CodecUtils cause it was bugging me

Relies on #1138 

TODO

- [x] Write tests for the Gson polymorphic stuff


pom.xml
```diff
@@ -114,7 +114,7 @@
         <repository>
           <id>sarxos</id>
           <name>sarxos</name>
-          <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+          <url>http://oss.sonatype.org/content/repositories/snapshots</url>
```
supertick: You don't want to revert this - also was this generated.. if it was that's great, that's how we manage dependencies.  Since I refreshed your branch, I hope a regen of the pom will the https, cuz it was put into the template

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -34,51 +24,210 @@
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.Constructor;
 
-import com.google.gson.Gson;
-import com.google.gson.GsonBuilder;
-import com.google.gson.internal.LinkedTreeMap;
+import java.io.*;
```
supertick: Should CTRL+SHIFT+O to organize imports, wildcards are not needed

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -1,28 +1,18 @@
...
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
-
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.PropertyAccessor;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
```
supertick, line 6: is the idea that fasterxml.jackson replaces gson - is that the strategy everyone is doing moving forward?
why is gson still included?

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -34,51 +24,210 @@
...
+   * A description of an API type
+   */
   public static class ApiDescription {
-    String key;
-    String path; // {scheme}://{host}:{port}/api/messages
-    String exampleUri;
-    String description;
+    /**
+     * The string to use after {@link #PARAMETER_API}
+     * in URIs to select this API.
```
supertick: nice comments :+1:

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -34,51 +24,210 @@
...
+     * @param exampleUri {@link #exampleUri}
+     * @param description {@link #description}
+     */
     public ApiDescription(String key, String uriDescription, String exampleUri, String description) {
       this.key = key;
       this.path = uriDescription;
       this.exampleUri = exampleUri;
       this.description = description;
     }
+
```
supertick, line 259: equals and hashCode for a utils ?

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -34,51 +24,210 @@
...
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types">MIME types</a>
+   */
   public final static String MIME_TYPE_JSON = "application/json";
 
-  private transient static Gson gson = new GsonBuilder().setDateFormat("yyyy-MM-dd HH:mm:ss.SSS").disableHtmlEscaping().create();
-  private transient static Gson prettyGson = new GsonBuilder().setDateFormat("yyyy-MM-dd HH:mm:ss.SSS").setPrettyPrinting().disableHtmlEscaping().create();
+  /**
+   * Whether we are using the GSON JSON backend or not. If false,
+   * use Jackson.
+   * TODO Replace with enum to allow extension for multiple backends
```
supertick: when will we be using gson and when jackson ?

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -34,51 +24,210 @@
...
+   * @see #USING_GSON
+   */
+  private static final TypeFactory typeFactory = TypeFactory.defaultInstance();
+
+
+  //Class initializer to setup mapper when the class is loaded
+  static {
+    //This allows Jackson to work just like GSON when no default constructor is available
+    mapper.registerModule(new NoCtorDeserModule());
+
```
supertick: ok cool, trying to establish some backward compatibility ?

src/main/java/org/myrobotlab/framework/Message.java
```diff
@@ -26,10 +26,7 @@
 package org.myrobotlab.framework;
 
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
supertick: ctrl+shift+o

src/main/java/org/myrobotlab/framework/Message.java
```diff
@@ -286,4 +283,18 @@ public String getMethod() {
     return method;
   }
 
+  @Override
+  public boolean equals(Object o) {
```
supertick, line 292: Good - makes sense here :+1:

src/main/java/org/myrobotlab/framework/Message.java
```diff
@@ -286,4 +283,18 @@ public String getMethod() {
     return method;
   }
 
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    Message message = (Message) o;
+    return msgId == message.msgId && Objects.equals(name, message.name) && Objects.equals(sender, message.sender) && Objects.equals(sendingMethod, message.sendingMethod) && Objects.equals(historyList, message.historyList) && Objects.equals(properties, message.properties) && Objects.equals(status, message.status) && Objects.equals(encoding, message.encoding) && Objects.equals(method, message.method) && Arrays.equals(data, message.data);
```
supertick: :+1:

src/main/java/org/myrobotlab/framework/Message.java
```diff
@@ -286,4 +283,18 @@ public String getMethod() {
...
+  }
+
+  @Override
+  public int hashCode() {
+    int result = Objects.hash(msgId, name, sender, sendingMethod, historyList, properties, status, encoding, method);
+    result = 31 * result + Arrays.hashCode(data);
+    return result;
+  }
 }
```
supertick, line 319: needs a newline

src/main/java/org/myrobotlab/framework/MethodCache.java
```diff
@@ -553,7 +547,7 @@ public Object[] getDecodedJsonParameters(Class<?> clazz, String methodName, Obje
       Class<?>[] paramTypes = possible.get(p).getParameterTypes();
       try {
         for (int i = 0; i < encodedParams.length; ++i) {
-          if (encodedParams[i].getClass() == LinkedTreeMap.class) {
+          if (CodecUtils.JSON_DEFAULT_OBJECT_TYPE.equals(encodedParams[i].getClass())) {
```
supertick, line 554: is it a LinkedTreeMap ?

src/main/java/org/myrobotlab/framework/Registration.java
```diff
@@ -87,4 +89,17 @@ public boolean hasInterface(Class<?> interfaze) {
     // TODO Auto-generated method stub
     return false;
   }
+
```
supertick, line 93: :+1:

src/test/java/org/myrobotlab/codec/CodecUtilsTest.java
```diff
@@ -0,0 +1,48 @@
...
+        Locale mrlLocale;
+
+        // String code construction
+        String code = null;
+        String json = null;
+
+        java.util.Locale check = new java.util.Locale("zh-cmn-Hans-CN");
+
+
+        code = "-";
```
supertick, line 22: So "class" is in our json now ?

src/test/java/org/myrobotlab/codec/json/GsonPolymorphicTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+public class GsonPolymorphicTest extends AbstractTest {
+    public final static Logger log = LoggerFactory.getLogger(GsonPolymorphicTest.class);
+
+    private static Gson polymorphicGson;
+
+    private static Gson regularGson;
+
+    @BeforeClass
+    public static void setup() {
```
supertick, line 26: unit tests :+1:

supertick (changes requested):

pom.xml
AutonomicPerfectionist: Yeah that was generated, should be fixed with a regen

src/main/java/org/myrobotlab/codec/CodecUtils.java
AutonomicPerfectionist: Whoops, Intellij did that on an autoformat, I'll have to adjust my settings

src/main/java/org/myrobotlab/codec/CodecUtils.java
AutonomicPerfectionist: GSON is still included in case I missed some behavior that Jackson does differently, so if there's a problem we can quickly switch over to Gson again with a single-line hotfix

src/main/java/org/myrobotlab/codec/CodecUtils.java
AutonomicPerfectionist: From what I understand, ApiDescription is a dataclass whose only purpose is to hold information, it's generally considered required to add equals and hashcode for dataclasses so if we ever wanted to check for equality or put them in a map it won't rely on the identity of the object. It is more boilerplate though, I don't believe we're actively using equals and hashcode except in tests so if we need to they can be removed

src/main/java/org/myrobotlab/codec/CodecUtils.java
AutonomicPerfectionist: Right now Gson is the selected backend for compatibility reasons, I'd like to switch to Jackson though because it's more flexible. This boolean switch was made in case there's an issue we discover with the Jackson backend down the line so we can quickly switch over to Gson with a one-line hotfix

src/main/java/org/myrobotlab/framework/MethodCache.java
AutonomicPerfectionist: Gson uses a LinkedTreeMap while Jackson uses some other tree map, this code here ensures we don't rely on a specific implementation's default type

src/test/java/org/myrobotlab/codec/CodecUtilsTest.java
AutonomicPerfectionist: Yes, "class" was chosen since it's an invalid identifier in most languages but it is configurable

pom.xml
AutonomicPerfectionist: @supertick Regen has fixed that regression, but it did also reorganize some of the dependencies due to my recent change to the generation code. Say, for example, we had service 1 and service 2 that both had the Apache commons dependency, before this PR service 2 had the dependency in the pom.xml and service 1 had a comment that it was a duplicate. Now, some of them are reversed, so service 1 has the dependency, and service 2 has the comment about a duplicate. So functionally no difference, but the diff is very big and scary-looking

src/main/java/org/myrobotlab/codec/CodecUtils.java
AutonomicPerfectionist: Should be fixed now

src/main/java/org/myrobotlab/framework/Message.java
AutonomicPerfectionist: Should be fixed

pom.xml
supertick: :+1:

src/main/java/org/myrobotlab/framework/MethodCache.java
```diff
@@ -2,13 +2,7 @@
...
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeMap;
-import java.util.TreeSet;
+import java.util.*;
```
supertick: more ctrl+shift+o

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -20,770 +43,1252 @@
...
+        //Disables Jackson's automatic property detection
+        mapper.setVisibility(PropertyAccessor.ALL, JsonAutoDetect.Visibility.NONE);
+        mapper.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
+        mapper.setVisibility(PropertyAccessor.SETTER, JsonAutoDetect.Visibility.ANY);
 
-  public final static String makeFullTypeName(String type) {
-    if (type == null) {
-      return null;
+        //Make jackson behave like gson in that unknown properties are ignored
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
supertick, line 203: :+1:  yeah good call

src/main/java/org/myrobotlab/framework/Message.java
AutonomicPerfectionist: Should be fixed now

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -20,770 +43,1252 @@
...
+     *                      of genericClass.
+     * @param <T>           The type of the target class.
+     * @return An object of the specified class (or a subclass of) with the state
+     * given by the json.
+     * @throws JsonDeserializationException if an error during deserialization occurs.
+     * @see #USING_GSON
+     */
+    public static <T extends Object> T fromJson(String json, Class<?> genericClass, Class<?>... parameterized) {
+        try {
+            if (USING_GSON)
```
supertick: stylistically @kwatters and I never do conditionals without curly braced blocks.  I think this adds clarity to someone say that came from a Python background and assuming indention determines blocks.
Just want to call it out, because I know an argument can be made saying it reduced eye clutter.

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -20,770 +43,1252 @@
...
+        }
+    }
+
+    /**
+     * Converts a topic method name to the name of the method that is
+     * used for callbacks. Usually this involves prepending the string
+     * "on", removing any "get" or "publish" prefix, and converting
+     * it all to proper camelCase.
+     *
+     * @param topicMethod The topic method name, such as "publishState"
```
supertick, line 408: very nice job on all the comments :+1:

src/main/java/org/myrobotlab/framework/Message.java
```diff
@@ -286,4 +288,27 @@ public String getMethod() {
...
+            && Arrays.equals(data, message.data);
+  }
+
+  @Override
+  public int hashCode() {
+    int result = Objects.hash(msgId, name, sender, sendingMethod, historyList, properties, status, encoding, method);
+    result = 31 * result + Arrays.hashCode(data);
+    return result;
+  }
 }
```
supertick, line 319: newline ;)

supertick (changes requested):

src/main/java/org/myrobotlab/framework/MethodCache.java
AutonomicPerfectionist: Should be fixed

src/main/java/org/myrobotlab/codec/CodecUtils.java
AutonomicPerfectionist: Noted, I tried to find all instances of that style but I use it a lot so it's probably snuck in in a lot of other places

src/main/java/org/myrobotlab/framework/Message.java
AutonomicPerfectionist: Should be fixed

src/test/java/org/myrobotlab/service/MqttBrokerTest.java
```diff
@@ -31,7 +31,7 @@ public void testService() throws Exception {
     assertTrue(mqtt.isConnected());
 
     // fire mqtt msg to execute python test and set value of test_value
-    mqtt.publish("api/service/python/exec/test('worky!')");
+    mqtt.publish("api/service/python/exec/\"test(\\\"worky!\\\")\"");
```
supertick, line 34: heh .. :)

PR #1116, opened 2022-07-21
Change sonatype repo from http to https

holgerfriedrich:
Allow building with maven >=3.8.1, maven-default-http-blocker

PR #1138, opened 2022-08-01
Make POM generation respect includeInOneJar when duplicate dependencies exist

### LINKED ISSUE #1137: MavenWrapper does not select correct scope of dependency if it's found at a different scoep earlier
**Describe the bug**
When MavenWrapper generates the pom.xml, it loops over all of the services to find their dependencies. It will then determine what their scope should be by looking at the `includeServiceInOneJar()` property. If it finds a dependency that was already written, it will skip the dependency. However, if the first one had a different scope, then it will not be updated to match the scope of the skipped one.

**To Reproduce**
1. Add Jackson dependencies to RuntimeMeta
2. Generate pom.xml with MavenWrapper
3. Observe the jackson dependencies still listed as `provided` due to MQTT being processed first

**Expected behavior**
If a dependency that was already written is found but with a higher scope, then the scope should be updated.

AutonomicPerfectionist:
Fixes #1137 by adding a preprocessing stage to scan for duplicate dependencies and to select the one that requires inclusion in the final jar as the one to put in the pom.xml file.


pom.xml
```diff
@@ -111,7 +111,7 @@
         <repository>
           <id>sarxos</id>
           <name>sarxos</name>
-          <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+          <url>http://oss.sonatype.org/content/repositories/snapshots</url>
```
holgerfriedrich: Pls do not change back to http... This will break compatibility to newer mvn versions (maven-default-http-blocker)

AutonomicPerfectionist: Oops, not sure how that happened. I'll take a look at it in a bit

AutonomicPerfectionist: @holgerfriedrich I see what happened, MRL dynamically generates a POM file from a template, #1116 only changed pom.xml and not the configuration used to generate the repositories, so when it was regenerated the change was wiped out. I think a new PR should be opened to change to https, that would be in `org.myrobotlab.framework.repo.Repo`:

https://github.com/MyRobotLab/myrobotlab/blob/6e24be329e36209e5d4a9af89c56637faa504b12/src/main/java/org/myrobotlab/framework/repo/Repo.java#L135-L152

AutonomicPerfectionist: @holgerfriedrich I updated the repo config and regenerated the pom, it should be using https again

PR #1142, opened 2022-08-16
programab session focus wake and sleep

supertick:


src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1329,7 +1342,8 @@ public void onUtterance(Utterance utterance) throws Exception {
         // TODO: don't talk to bots.. it won't go well..
         // TODO: the discord api can provide use the list of mentioned users.
         // for now.. we'll just see if we see Mr. Turing as a substring.
-        if (utterance.text.contains(botName)) {
+        sleep = (sleep || utterance.text.contains("@")) && !utterance.text.contains(botName);
```
kwatters, line 1369: i'd like to preserve existing behavior as default and make this behavior a configuration option exposed in the config

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1296,6 +1298,16 @@ synchronized public void addCategoryToFile(Bot bot, Category c) {
       error(e);
     }
   }
+  
+  boolean sleep = false;
```
kwatters: member variables should go to the top of the class.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1296,6 +1298,16 @@ synchronized public void addCategoryToFile(Bot bot, Category c) {
       error(e);
     }
   }
+  
+  boolean sleep = false;
+  
+  public void wake() {
```
kwatters, line 1325: are you intending these to be invoked via aiml ?

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: agreed fixed

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: agreed fixed

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: yes - it worky ... tried first with single method and boolean ... and that failed ...   :(

src/main/java/org/myrobotlab/service/config/ProgramABConfig.java
```diff
@@ -15,5 +15,15 @@ public class ProgramABConfig extends ServiceConfig {
   // search engine - currently wikipedia
   // search for sraix
   public String search;
+  
+  /**
+   * enables global session and sleep/wake functionality
+   */
+  public boolean useGlobalSession;
```
kwatters: you've added this configuration variable, but it isn't actually used anywhere. please remove.

pom.xml
```diff
@@ -349,23 +349,11 @@
 <!-- DocumentPipeline end -->
 
 <!-- Elasticsearch begin -->
-  <dependency>
```
kwatters: Why are there pom changes in this PR?  please revert pom.xml

supertick: pretty sure the current was out of spec not being generated ... but whatever, i reverted

PR #1143, opened 2022-08-16
Discord image publish

supertick:


src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -261,4 +279,68 @@ public void setBotName(String name) {
...
   }
 
+  @Override
+  public void onImage(ImageData img) {
+    sendImage(img, null, null);
+  }
+
+  public void sendImage(ImageData img, String id, String channelName) {
+    if (channelName == null) {
+      channelName = "general";
```
kwatters, line 266: this seems wrong..  why general?  that's not correct..

supertick: that's funny ... its not even part of my pr .. its part of your original commit ... 
¯\_(ツ)_/¯

src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -153,21 +163,29 @@ public void setToken(String token) {
     this.token = token;
   }
 
-  public static void main(String[] args) throws Exception {
-    // Brief example of starting a programab chatbot and connecting it to
-    // discord
+  public void runMrT() throws LoginException, IOException {
```
kwatters: this should not be on the DiscordBot.

src/main/java/org/myrobotlab/service/DiscordBot.java
kwatters: this is part of your PR.  the sendImage method is net new  here.

src/main/java/org/myrobotlab/service/DiscordBot.java
kwatters: i see what you are referring to.. in the sendUtterance method.. that default logic should be removed and an exception should be thrown if the channel isn't known.

supertick: that's increasing scope - i lifted what was previously done in sendUtterance

supertick: I didn't put runMrT in it..  although i agree ... i might remove it

src/main/java/org/myrobotlab/service/DiscordBot.java
supertick: I copied paste sendImage from pre-existing sendUtterance .. not in this scope to refactor it out

src/main/java/org/myrobotlab/service/DiscordBot.java
```diff
@@ -261,4 +256,81 @@ public void setBotName(String name) {
     this.botName = name;
   }
 
+  @Override
+  public void onImage(ImageData img) {
+    sendImage(img, null, null);
```
kwatters, line 261: this means that images by default will always be sent to "general" ... no?  sorry, we need to fix this somehow.

supertick: leaving it as is - "fixed the problem" by embedding in text urls, this ImageListener interface should not be taken away however - perhaps improved in the future with options of binary data and certainly channel can be added to ImageData if necessary

PR #1146, opened 2022-08-24
Pom gen 1

supertick:


kwatters:
This PR should only have pom.xml and changes to pom.xml.template in here...

PR #1153, opened 2022-09-03
AudioFile Update

supertick:


src/main/java/org/myrobotlab/service/AudioFile.java
```diff
@@ -508,6 +530,7 @@ public ServiceConfig getConfig() {
     config.currentPlaylist = currentPlaylist;
     config.volume = volume;
     config.playlists = playlists;
+    config.peakMultiplier = peakMultiplier ;
```
kwatters, line 531: we should add peakSampleInterval here also.

supertick: nice ! :+1:

src/main/java/org/myrobotlab/audio/AudioProcessor.java
```diff
@@ -195,21 +203,52 @@ public AudioData play(AudioData data) {
...
+              }
+
+              rms = (float) Math.sqrt(rms / samples.length);
+
+              if (lastPeak > peak) {
+                peak = lastPeak; 
+              }
+
+              lastPeak = peak;
+              audioFile.invoke("publishPeak", peak * (float) audioFile.getPeakMultiplier());
```
kwatters, line 224: what's the point of computing the rms if all you are doing is publishing the peak value?

src/main/java/org/myrobotlab/audio/AudioProcessor.java
```diff
@@ -195,21 +199,52 @@ public AudioData play(AudioData data) {
...
-
-          // the buffer of raw data could be published from here
-          // if a reference of the service is passed in
-
           if (audioFile.isMute()) {
             // NoOp for a mute audioFile.
           } else {
+            AudioFileConfig config = (AudioFileConfig)audioFile.getConfig();
+            if (cnt % config.peakSampleInterval == 0) {
+              float[] samples = new float[buffer.length / 2];
```
kwatters: you shouldn't be creating this array here...  you should just scan over the "sample" and find the peak.

src/main/java/org/myrobotlab/audio/AudioProcessor.java
```diff
@@ -195,21 +199,52 @@ public AudioData play(AudioData data) {
...
+                sample |= buffer[i++] & 0xFF; // (reverse these two lines
+                sample |= buffer[i++] << 8; // if the format is big endian)
+
+                // normalize to range of +/-1.0f
+                samples[s++] = sample / 32768f;
+              }
+
+              // float rms = 0f;
+              float peak = 0f;
+              for (float sample : samples) {
```
kwatters: remove this loop.. and do it in the loop above.

src/main/java/org/myrobotlab/service/AudioFile.java
kwatters, line 531: you missed peakSampleInterval ...

PR #1164, opened 2022-09-15
ProgramAB has htmlfilter and config

supertick:
A way to get html filtering on ProgramAB without breaking anything, and giving it the flexibility to add more filters through config.


kwatters:
How would you get both the filtered and the raw response with this?  It seems like you have removed that ability.

PR #1165, opened 2022-09-16
Fix to WebKitSpeechRecognition to set language

supertick:


src/main/resources/resource/WebGui/app/service/js/WebkitSpeechRecognitionGui.js
```diff
@@ -181,14 +181,24 @@ angular.module('mrlapp.service.WebkitSpeechRecognitionGui', []).controller('Webk
         }
     }
 
+    $scope.setLanguageFromService = function(lang) {
+        // recognizer.lang = $scope.selectedLanguage
+        recognizer.lang = lang
+        // sync'ing runtime platform locale
+        // msg.send("setLocale", $scope.selectedLanguage)
+        if ($scope.isRecording) {
```
kwatters, line 189: why does this stop the recognizer if you change the language?

supertick: its required to change languages

PR #1166, opened 2022-09-16
Added some notes to JavaDocs

Cyber-One:


src/main/java/org/myrobotlab/service/Adafruit16CServoDriver.java
```diff
@@ -621,12 +630,17 @@ public void motorReset(MotorControl motor) {
 
   }
 
-    /*
+   /*
    * Set Output Power of a pin.
```
supertick, line 634: to be javadocs you need double star at top like this /**

supertick (changes requested):

PR #1175, opened 2022-09-18
Fixes POST calls to service API

supertick:
service api now accepts json record of parameters in a HTTP POST request


kwatters:
1175 was the number of the house that i lived at when I was growing up.. Anyway. looks good, though, I wonder what encoding is going to be implied when we get the bytes and create a string out of it.  I suspect that would default to the local platform, so we might have some oddities on different platfroms if they're not using utf-8...

PR #1176, opened 2022-09-22
Update Pcf8574.java

Cyber-One:
Made a number of changes to match the way the hardware works. Added more JavaDocs as well.

### DIFF
--- a/src/main/java/org/myrobotlab/service/Pcf8574.java
+++ b/src/main/java/org/myrobotlab/service/Pcf8574.java
@@ -30,13 +30,14 @@ import org.slf4j.Logger;
  * PCF8574 / PCF8574A Remote I/O expander for i2c bus with interrupt ( interrupt
  * not yet implemented )
  * 
- * @author Mats
+ * References: http://www.digikey.com/product-detail/en/nxp-semiconductors/PCF8574T-3,518/568-1077-1-ND/735791
  * 
- *         References:
- *         http://www.digikey.com/product-detail/en/nxp-semiconductors
- *         /PCF8574T-3,518/568-1077-1-ND/735791
+ * A quasi-bidirectional I/O is an input or output port without using a direction control register. 
+ * Whenever the master reads the register, the value returned to master depends on the actual voltage or status of the pin. 
+ * At power on, all the ports are HIGH with a weak 100 A internal pull-up to VDD, but can be driven LOW by an internal transistor, or an external signal. 
+ * The I/O ports are entirely independent of each other, but each I/O octal is controlled by the same read or write data byte.
  * 
- *         FIXME - this is a i2c controller as well as an i2c device
+ * @author Mats Onnerby modified by Ray Edgley.
  * 
  */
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/myrobotlab/service/Pcf8574.java
```diff
@@ -499,23 +492,17 @@ public void onRegistered(Registration s) {
...
   public void pinMode(int address, int mode) {
     PinDefinition pinDef = getPin(address);
-    if (mode == INPUT) {
-      pinDef.setMode("INPUT");
-      directionRegister &= ~(1 << address);
-    } else {
-      pinDef.setMode("OUTPUT");
-      directionRegister |= (1 << address);
-    }
+    // There is no direction register in the PCF8574 it is always BIDRECTIONAL.
```
supertick, line 495: might be worth logging a warning or info - saying if its != "BIDIRECTIONAL"
Pcf8574 only does BIDIRECTIONAL

src/main/java/org/myrobotlab/service/Pcf8574.java
```diff
@@ -303,7 +296,7 @@ public Map<String, PinDefinition> createPinList() {
       pindef.setPwm(false);
       pindef.setPinName(name);
       pindef.setAddress(i);
-      pindef.setMode("INPUT");
+      pindef.setMode("BIDIRECTIONAL");
```
supertick, line 299: ok :+1:

src/main/java/org/myrobotlab/service/Pcf8574.java
```diff
@@ -575,6 +563,34 @@ int readRegister() {
...
+  /**
+   * This returns the last set value on the pin.
+   * When this is set to True, a read value of false indicates the pin is pulled low by external influence.
+   * 
+   * @param address Integer
+   * The pin to be looked at
+   * @return current state of the output register for the pin
+   */
+  public boolean readOutputPin(int address) {
+    int value = (writeRegister >> address) & 1;
```
supertick, line 578: I think it should return an int .. it would be really good if it would match an interface.
Mrl is too Arduino specific.  In the big picture this should follow file reading interface although imo Java's not the best spec for this.
perhaps 
```java
public int read(int address) {
```

src/main/java/org/myrobotlab/service/Pcf8574.java
```diff
@@ -575,6 +563,34 @@ int readRegister() {
...
+   * This returns the last set value on the pin.
+   * When this is set to True, a read value of false indicates the pin is pulled low by external influence.
+   * 
+   * @param pinName String
+   * @return current state of the output register for the pin
+   */
+  public boolean readOutputPin(String pinName) {
+    return readOutputPin(getPin(pinName).getAddress());
+  }
+
```
supertick, line 592:
```java
public int read(String pinName) {
```

src/main/java/org/myrobotlab/service/Pcf8574.java
```diff
@@ -590,19 +606,27 @@ public void setBus(String bus) {
     setDeviceBus(bus);
   }
 
+  /**
+   * Set the I2C Address of the device.
```
supertick, line 609: nice ! :+1:

supertick (changes requested):

supertick (approved):

PR #1178, opened 2022-09-24
Update Hd44780.java

Cyber-One:
Lots of changes to make it more reliable.
Also added more functions and lots of javaDocs :-)
Also removed the delay value, this can be done in the user program.

### DIFF
--- a/src/main/java/org/myrobotlab/service/Hd44780.java
+++ b/src/main/java/org/myrobotlab/service/Hd44780.java
@@ -32,7 +32,16 @@ public class Hd44780 extends Service {
 
   protected boolean backLight;
 
-  public final byte En = (byte) 0b00000100; // Enable bit
+  public final byte Rs = (byte) 0b00000001; // Register select bit
+  public final byte Rw = (byte) 0b00000010; // Read/Write bit
+  public final byte En = (byte) 0b00000100; // Enable/Strobe bit
+  public final byte Bl = (byte) 0b00001000; // Backlight bit
+  public final byte DB = (byte) 0b11110000; // Data Bits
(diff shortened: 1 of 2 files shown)


src/main/java/org/myrobotlab/service/Hd44780.java
```diff
@@ -196,97 +193,288 @@ public void display(String string, int line) {
...
-        lcdWrite((byte) 0xC0);
+      setDDRAMaddress((byte) 0x40);
         break;
       case 2:
-        lcdWrite((byte) 0x94);
+      setDDRAMaddress((byte) 0x14);
         break;
       case 3:
-        lcdWrite((byte) 0xD4);
+      setDDRAMaddress((byte) 0x3C);
```
supertick: The convention we follow with the myrobotlab is camel back case including abbreviations
So setDDRAMaddress should be setDdramAddress 
just like setXML would be setXml

src/main/java/org/myrobotlab/service/Hd44780.java
```diff
@@ -296,6 +484,10 @@ public void setBackLight(boolean status) {
    *          c
    * 
    */
+  // Why are we exposing the writeRegister publicly here.
+  // This is a dangerous method to the HD44780 as anything written here can destabilize it.
+  // not yet removed but should be depreciated or removed if not being used.
+  @Deprecated // use lcdWriteCmd(byte cmd) or lcdWriteData(byte data)
   public void writeRegister(byte cmd) {
```
supertick: its used internally so it shouldn't be deprecated unless its mis-used,
but you should change the public to private

supertick (changes requested):

supertick (approved):

PR #1211, opened 2022-12-03
Delegate WebGui message deser and execution to reduce code duplication

### LINKED ISSUE #1128: Service.sendBlocking() causes method cache lookup failure when sender doesn't have callback topic method
**Describe the bug**
When calling `Service.sendBlocking()` with a remote target, the return message has the `method` field set to the result of `CodecUtils.getCallbackTopicName(method)`. This message is intercepted within the sender's inbox, but before it reaches there, the message must be deserialized at one of the gateways. Usually, the datatypes of the message parameters are deserialized by using the method cache and looking up the target method. However, in this case, the message target method does not always exist, as it will be intercepted and is never actually executed. Thus, the deserialization may fail; without that stage, the message never reaches the sender's inbox.

**To Reproduce**
1. Run two MRL instances and connect them over HTTP
2. Start a service on one, such as Python
3. On the other, execute `Runtime.getInstance().sendBlocking()` with the full name of the first's Python service, the method name `getId()`, and an empty list of `Object`s for the data
4. Observe a `TimeoutException` on the second instance and multiple errors regarding method cache lookup failure

**Expected behavior**
Invoking `sendBlocking()` should not require the sender to have the callback method since it is never called.

**Additional context**
#1124 uses `sendBlocking()` for forwarding messages to the remote service, the message method is arbitrary, so it triggers this bug.

AutonomicPerfectionist:
This PR adds a centralized method in CodecUtils to handle deserializing Messages along with their double-encoded parameters. This ensures that the code to do so is not duplicated in multiple gateways, such as WebGui.

It also forces the WebGui to delegate message execution to a service's inbox. Before, the WebGui would manually find the service method and invoke it on its own servicing thread, but this resulted in code duplication and as a result when `sendBlocking()` was added this message execution code was not updated. Meaning `sendBlocking()` would timeout when traversing the WebGui gateway.

Finally, this PR adds some nullability annotations to the public API of CodecUtils and MethodCache. This is just a preliminary addition before making more widespread additions in a later PR but already they've helped me locate two potential NPE errors that are also fixed in this PR.

Fixes #1128


pom.xml
```diff
@@ -1319,6 +1309,11 @@
     <artifactId>async-http-client</artifactId>
     <version>2.12.3</version>
   </dependency>
+  <dependency>
+    <groupId>com.google.code.findbugs</groupId>
```
supertick: How is this used, is Runtime using it or is it used in testing ?
Findbugs looks like it was updated to SpotBugs ?

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -241,11 +246,11 @@ public static String capitalize(final String line) {
...
      * @param <T>   The type of the target class.
      * @return An object of the specified class (or a subclass of) with the state
-     * given by the json.
+     * given by the json. Null is an allowed return object.
      * @throws JsonDeserializationException if an error during deserialization occurs.
      * @see #USING_GSON
      */
-    public static <T extends Object> T fromJson(String json, Class<T> clazz) {
+    public static <T> @Nullable T fromJson(@Nonnull String json, @Nonnull Class<T> clazz) {
         try {
```
supertick, line 262: I'd be curious of json decoding null.
I know by definition null IS a valid json primitive, so is a string, so perhaps fromJson(null, String.class) should return null ?  The spec is clear about what are valid primitives, its probably not clear on what a decoder should do.
More importantly, how does gson decoding or the CodecUtils previously handle it ?

src/main/java/org/myrobotlab/service/meta/RuntimeMeta.java
```diff
@@ -49,6 +49,12 @@ public RuntimeMeta() {
     // ws client sockets
     addDependency("org.asynchttpclient", "async-http-client", "2.12.3");
 
+    // Nullability annotations
+    // Contrary to what its group ID implies, this is
+    // the reference implementation for javax.annotation.Nullable / NotNull
+    // Apache 2.0 license
+    addDependency("com.google.code.findbugs", "jsr305", "3.0.2");
+
```
supertick, line 52: should it be updated to spotbugs ?

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -544,6 +544,7 @@ public void handle(AtmosphereResource r) {
       String apiKey = getApiKey(r.getRequest().getRequestURI());
 
       // the mrl "id" of the client
+      // TODO remove since not used
       String id = r.getRequest().getParameter("id");
```
supertick: if not used please remove it

src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -995,7 +1108,9 @@ static public Double makeDouble(String data) {
     static public Boolean isBoolean(String data) {
         try {
             Boolean.parseBoolean(data);
-            return true;
+            // The above will return false and not throw
+            // exception for most cases
+            return "true".equals(data) || "false".equals(data);
```
supertick, line 1165: :+1:

src/main/java/org/myrobotlab/framework/MRLListener.java
```diff
@@ -78,9 +78,13 @@ public MRLListener(String topicMethod, String callbackName, String callbackMetho
...
-    if (callbackName.equals(other.callbackName) && callbackMethod.equals(other.callbackMethod) && topicMethod.equals(other.topicMethod)) {
-      return true;
+  @Override
+  public boolean equals(final Object other) {
+    if (other instanceof MRLListener) {
+      MRLListener listener = (MRLListener) other;
+      return callbackName.equals(listener.callbackName)
+              && callbackMethod.equals(listener.callbackMethod)
+              && topicMethod.equals(listener.topicMethod);
     }
```
supertick, line 88: :+1:

src/main/java/org/myrobotlab/framework/MethodCache.java
```diff
@@ -529,26 +531,50 @@ public List<MethodEntry> getRemoteOrdinalMethods(Class<?> object, String methodN
...
+   * <p>
+   *     FIXME Change encodedParameters to String[] to enforce type safety
+   * </p>
+   *
+   * @param clazz The class to lookup methods for
+   * @param methodName The name of the method to decode parameters for
+   * @param encodedParams The encoded parameters in JSON format.
+   * @return The decoded parameters according to a matched method, or null
+   *  if no such method could be found.
+   */
```
supertick, line 550: what is the info returned in the error when methodName parameter is null ? does the exception contain the npe parameter name ?

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -665,7 +666,11 @@ public void handle(AtmosphereResource r) {
         // decoding 1st pass - decodes the containers
         Message msg = null;
         try {
-          msg = CodecUtils.fromJson(bodyData, Message.class);
+          msg = CodecUtils.jsonToMessage(bodyData);
+          if (msg == null) {
+            log.error("Got null message from client, check client code for bugs");
```
supertick, line 668: :+1:

src/test/java/org/myrobotlab/codec/CodecUtilsTest.java
```diff
@@ -24,27 +25,106 @@ public void testLocale() {
         code = "-";
         mrlLocale = new Locale(code);
         json = CodecUtils.toJson(mrlLocale);
-        assertEquals("{\"language\":null,\"displayLanguage\":null,\"country\":null,\"displayCountry\":null,\"tag\":null,\"class\":\"org.myrobotlab.service.data.Locale\"}", json);
+        if(!CodecUtils.USING_GSON) {
+            assertEquals("{\"language\":null,\"displayLanguage\":null,\"country\":null,\"displayCountry\":null,\"tag\":null,\"class\":\"org.myrobotlab.service.data.Locale\"}", json);
```
supertick: USING_GSON=true fail these tests ?

src/test/java/org/myrobotlab/codec/CodecUtilsTest.java
```diff
@@ -24,27 +25,106 @@ public void testLocale() {
...
+                                )
+                        }
+                );
+        String json = CodecUtils.toJsonMsg(message);
+        Message deserMessage = CodecUtils.jsonToMessage(json);
+        assertEquals(message, deserMessage);
+    }
+
+    @Test
+    public void testDecodeMessageParams() {
```
supertick, line 78: :+1:  Good .. clear unit tests

src/test/java/org/myrobotlab/codec/CodecUtilsTest.java
```diff
@@ -24,27 +25,106 @@ public void testLocale() {
...
+                }
+        );
+
+        Message decodedMessage = Message.createMessage(
+                "testing-sender",
+                "testService",
+                "testMethod",
+                new Object[]{
+                        "testParam1",
+                        2,
```
supertick, line 117: heh.. it be nice to see all primitive types and a few null values :)

supertick (changes requested):

src/main/java/org/myrobotlab/service/meta/RuntimeMeta.java
AutonomicPerfectionist: That depends, the FindBugs version adheres to jsr305 while spot bugs does not, but the spot bugs version is actively developed and comes with additional useful annotations

pom.xml
AutonomicPerfectionist: It's a set of nullability annotations that are used in places within this PR to help with tracking down NPEs

src/main/java/org/myrobotlab/codec/CodecUtils.java
AutonomicPerfectionist: `gson.fromString(null, Object.class)` returns null while `mapper.readValue(null, Object.class)` throws an exception, and the same goes for any target class. In my opinion, the difference between a JSON null value and an invalid input is that if the JSON contains null, the string should be `"null"` and not actually null. Passing in a null would indicate some failure upstream and shouldn't be taken as intentional

src/main/java/org/myrobotlab/framework/MethodCache.java
AutonomicPerfectionist: Hmm, it appears the documentation there is mistaken, no NPE is thrown, instead only null is returned and an error logged. In that case should I keep the `@Nonnull` and fix the documentation or change it to nullable?

supertick: :+1:

src/main/java/org/myrobotlab/service/meta/RuntimeMeta.java
supertick: what do you think ?

src/main/java/org/myrobotlab/codec/CodecUtils.java
supertick: null is valid json and null String is a valid reference to a non-existent string.  "null" is not null its just a string, and if we ever see "null" we probably have a problem to fix..
IMO null is valid input and a null input should be a null output - and should not throw.
A simple string is valid JSON too, and should not throw as input.
This is an interesting post describing the validity of a string only - https://stackoverflow.com/questions/7487869/is-this-simple-string-considered-valid-json.
I think null input should not throw, we should simply return null - and let downstream code handle the consequences

pom.xml
supertick: If you think its better than spotbugs I'm fine with that - just want to know the reasoning.

src/main/java/org/myrobotlab/codec/CodecUtils.java
AutonomicPerfectionist: > "null" is not null its just a string

I think this is where some confusion lies, it's correct that in `{"someval": "null"}`, `"null"` is just a string. But when it reaches `fromJson()`, that json is encoded as `"{\"someval\": \"null\"}"` in Java parlance. Likewise, the json value `null` is encoded in Java as `"null"`, so that `CodecUtils.toJson(null)` returns `"null"`, and not a null value (String containing `null` versus a `null` value entirely). If the JSON is intended to be null, we would pass `CodecUtils.fromJson("null", Object.class)`. This stuff definitely does get confusing, but if we look at Jackson we'll see it throws, and it seems to adhere pretty well to the JSON spec.

But, I can indeed allow that if it's what is desired, would just need to add a return in the catch

supertick: The context of this is a message's method parameter data right ?
I'm thinking Java method signatures which that take Booleans, Integers, Strings, Doubles, and Objects certainly allow null.
If its the Message is json encoded the non-string objects are should not be in quotes.
Heh, they are when its double encoded.  But that's another topic that can make the null topic more complex :)

My concern is if the decoder throws more, are some methods might no longer accessible with null values :shrug: 
The following should work without throwing...

```
{
  "method": "someMethod", 
  "data":[null, 123]
}

public void someMethod(Boolean b, int) {

```
Honestly, I'm not positive of the behavior now ... but whatever we decide to do, we should figure out how it currently behaves - make a decision on it, and lock that decision with a unit test.

AutonomicPerfectionist: > My concern is if the decoder throws more, are some methods might no longer accessible with null values shrug
The following should work without throwing...

That example would indeed work, because this:
```json
{
  "method": "someMethod", 
  "data":[null, 123]
}
```

Is a JSON-encoded String, including the null value embedded inside of it. There's already a test to ensure that works in CodecUtilsTest, the one called `testJsonMessageDeserNull`. Copied below for context:

```java
@Test
public void testJsonMessageDeserNull() {
    Message deserMessage = CodecUtils.jsonToMessage("null");
    assertNull(deserMessage);

}

```


When we pass JSON to the deserializer it is always in the form of a non-null String object, even if it represents a null value. Take the body of a POST request, for example: if you want the data to deserialize to null, you'd put `null` in the body, and that null is read as a String and passed to the deserializer. But if you put nothing, the body itself is null and will not deserialize at all because there's nothing there to deserialize.

Take this test here:
```java
Message deserMessage = CodecUtils.jsonToMessage(null);
```

Note how this is the same test mentioned above, but the null is not quoted, making it a null object instead of a String representing null. My intention is to make this throw because this is different from the test mentioned above and is only possible when there is a fault somewhere in the caller. This does not represent a null Message object like the previous test, but rather a complete lack of JSON at all, and interpreting it as a null Message can be dangerous.

Another example: passing `10` to `fromJson()` will not compile because `10` is not a String but rather an integer. But if we do this: `fromJson("10", Integer.class)`, we get the correct object. If we want to be consistent, then passing a null object should also fail because it is not a String, Java's type system just doesn't enforce that.

How it behaves right now differs depending on whether jackson or gson is used, that was unintentional, I thought the `@Nonnull` annotation would insert assertions to guarantee it's not null, but apparently, it does not do that. When GSON is selected, passing in a null object will return null, and when jackson is selected doing the same will throw an exception. Note that jackson is still able to deserialize messages that contain null data parameters, and the first test mentioned above succeeds with both. I will insert manual null checks / null returns once we come to a decision to keep the behavior between the two in parity.

supertick: Cool, having a null parameter in as a method parameter is a common thing, and sounds like jacksons and the updates can handle that.  null as input as a json string to decode explodes.  As you said, it would a problem with the client. :+1:

supertick (approved):

src/main/java/org/myrobotlab/service/meta/RuntimeMeta.java
AutonomicPerfectionist: I've looked through a few different options, and it seems that the most widely-used nullability annotations are from the Checker Framework, which also includes some pretty nifty annotations to help with logic bugs. I think that's probably the best option I can see at the moment, it seems a lot of large projects like Guava use it. I'm not sure integrating it in this PR is a good idea though, so I'll probably remove JSR305 and the annotations for now and make a new PR for Checker later

src/test/java/org/myrobotlab/codec/CodecUtilsTest.java
AutonomicPerfectionist: Gson by default does not serialize null values while Jackson does, I think serializing nulls is probably best since we use it in a lot of places as a special value, and I could probably configure gson to do that

src/main/java/org/myrobotlab/service/WebGui.java
AutonomicPerfectionist: Should be removed now

src/main/java/org/myrobotlab/framework/MethodCache.java
AutonomicPerfectionist: Nonnull removed and documentation updated

src/test/java/org/myrobotlab/codec/CodecUtilsTest.java
AutonomicPerfectionist: Added some more types and a list and nested list. In the process found a bug with how GSON behaves so it's good that I added those

src/test/java/org/myrobotlab/codec/CodecUtilsTest.java
AutonomicPerfectionist: I've made GSON serialize nulls as well so this test now doesn't need to differentiate between the serializers

src/main/java/org/myrobotlab/service/meta/RuntimeMeta.java
AutonomicPerfectionist: JSR305 removed and the annotations commented out (so I can convert them to Checker later), I'll make a new PR to add Checker stuff later

pom.xml
AutonomicPerfectionist: Removed and will be changed to Checker in a different PR

kwatters:
I'm consistently getting a failure in postTest() in WebGuiTest ... I also noticed, after kicking the build, that Jenkins only ran the first 43 unit tests.  The jvm crashed, and jenkins still reported the build as successful, even though it didn't make it through the unit tests...

AutonomicPerfectionist:
Oh, that's a problem... I just ran the tests locally and got that failure, but upon running it alone it succeeded, so probably something to do with being ran at the same time as other tests

AutonomicPerfectionist:
I can consistently trigger the segfault by running all of the OpenCVTest tests, but by running each individually, I only get a null pointer exception on a few of them (null pointer in native code, not because of a null java value). Surefire still lists the tests as succeeding; no idea why. Actually, looks like it might because the exception occurs in a different thread

```
17:44:25.613 ERROR [cv-video-processor-3] c.m.s.OpenCV [OpenCV.java:219] getting grabber failed
java.lang.NullPointerException: This pointer address is NULL.
	at org.bytedeco.opencv.opencv_core.Mat.cols(Native Method)
	at org.bytedeco.javacv.OpenCVFrameConverter.convert(OpenCVFrameConverter.java:167)
	at org.bytedeco.javacv.OpenCVFrameConverter$ToMat.convert(OpenCVFrameConverter.java:55)
	at org.myrobotlab.opencv.CloseableFrameConverter.toImage(CloseableFrameConverter.java:66)
	at org.myrobotlab.opencv.OpenCVFilterFaceDetectDNN.process(OpenCVFilterFaceDetectDNN.java:198)
	at org.myrobotlab.service.OpenCV.processVideo(OpenCV.java:1118)
	at org.myrobotlab.service.OpenCV$VideoProcessor.run(OpenCV.java:210)
	at java.base/java.lang.Thread.run(Thread.java:829)

```

Can't trigger the segfault by just running individual tests though, only by running all OpenCV tests

AutonomicPerfectionist:
As for the postTest(), I got it to succeed when running all other webgui tests by removing the Service.sleep() call, increasing the delay further caused a connection pool shutdown exception so for whatever reason the HTTP client we are using is shutting down before we were finished with the tests

kwatters:
. @AutonomicPerfectionist  Which Service.sleep() call are you referring to?  Did you happen to push a change?  I'm cool with the rest of the PR.  If we want to just ignore this test for now, we can do that too...

PR #1213, opened 2022-12-05
Config with peers and subscriptions

supertick:
Peers are now defined within configuration.
Subscriptions or "listeners" have been defined within config as well.
The methods getConfig and apply use super methods to process common subscription handling.

PR #1219, opened 2022-12-09
added listeners to config

supertick:


src/main/resources/resource/Adafruit16CServoDriver/adafruit16cservodriver.yml
```diff
@@ -2,5 +2,6 @@
 controller: null
 deviceAddress: null
 deviceBus: null
+listeners: null
```
AutonomicPerfectionist: Is there a reason the default is null instead of an empty list?

supertick: Thinking it was aesthetically more simple for users, than an empty list ... probably not true :shrug: 
All the npe details have been done to the best of knowledge,  e.g. ServiceConfig.getConfig() 
I don't have much issue making it an empty list ... biggest issue still is ...
**PROs**
- its awesome that pub/sub and routes of messages are exposed to user modification
- its great in theory anyone creating a new service doesn't need to add code to handle listeners/publishers/subscribers
- no special handling of these listeners - its all handled with framework code


**CONs**
- the ui subscriptions need to be filtered out 
- the subscribing to ServiceLifeCycle should be filtered out too 

itemizing this the PROs imo far outweigh the cons - but to minimize the CONs .. I currently have hardcoded filtering based on `callbackName.endsWith("@webgui-client")`

AutonomicPerfectionist: As long as null is checked in all of the required places I suppose it's fine here, I would prefer an empty list but that's just me.

For pros versus cons, I agree the pros outweigh the cons, but I think we need to have an adequate filtering system in place before merging. And just to confirm, this PR requires #1213 before it can be merged, correct?

supertick: No its not required, because this pr is documentation only ... its documentation on what the default config for all services would be.

AutonomicPerfectionist: So attempting to load these configs won't cause an exception because the listeners field doesn't exist in the config classes yet?

supertick: they aren't used in any form yet - only documentation

PR #1220, opened 2022-12-10
localspeech fixes

supertick:


src/main/java/org/myrobotlab/service/LocalSpeech.java
```diff
@@ -301,15 +309,20 @@ public boolean setFestival() {
    *          chars to filter.
    * 
    */
-  public void setFilter(String filter) {
-    filterChars = filter;
+  public void setFilter(String target, String replace) {
```
kwatters: The name of this method is misleading because it adds to the filter, it doesn't set it.   Please consider renaming to addFilter

PR #1221, opened 2022-12-15
Fix jackson being too strict when encountering strings without quotes

AutonomicPerfectionist:
This PR fixes a bug where switching to jackson causes the cli API to fail on strings without quotes. It also adds support for URL encoded characters in the service API


src/main/java/org/myrobotlab/codec/CodecUtils.java
```diff
@@ -838,8 +857,8 @@ static public Message cliToMsg(String contextPath, String from, String to, Strin
         cmd = cmd.trim();
 
         // remove uninteresting api prefix
-        if (cmd.startsWith("/api/service")) {
-            cmd = cmd.substring("/api/service".length());
+        if (cmd.startsWith(API_SERVICE_PATH)) {
+            cmd = cmd.substring(API_SERVICE_PATH.length());
         }
```
supertick, line 1027: :+1:

src/main/java/org/myrobotlab/service/WebGui.java
```diff
@@ -637,7 +639,11 @@ public void handle(AtmosphereResource r) {
 
       } else if (apiKey.equals(CodecUtils.API_SERVICE)) {
 
-        Message msg = CodecUtils.cliToMsg(null, getName(), null, r.getRequest().getPathInfo());
+        Message msg = CodecUtils.cliToMsg(
+                null,
+                getName(),
+                null,
+                URLDecoder.decode(r.getRequest().getPathInfo(), StandardCharsets.UTF_8));
```
supertick, line 644: Great ! ...
what's the use case that broke this, that you fixed ?
(should be added to unit test)

supertick (changes requested):

src/main/java/org/myrobotlab/service/WebGui.java
AutonomicPerfectionist: This section adds the ability to use URI path encoding schemes in the service API. So double quotes are %22 and will be transparently decoded to double quotes by this section. I'll add it to the test suite

PR #1222, opened 2022-12-18
Make subscribe and unsubscribe use full names

AutonomicPerfectionist:
This PR makes `Service.subscribe()` and related methods use the full service name when creating the `MRLListener` data object, as otherwise, the listener target can be ambiguous when traversing instances.

- [x] Fix merge conflict
- [x] Add tests


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1888,38 +1888,38 @@ public void subscribe(String topicName, String topicMethod, String callbackName,
       List<String> tnames = Runtime.getServiceNames(topicName);
       for (String serviceName : tnames) {
         MRLListener listener = new MRLListener(topicMethod, callbackName, callbackMethod);
-        send(Message.createMessage(getName(), serviceName, "addListener", listener));
+        send(Message.createMessage(getFullName(), serviceName, "addListener", listener));
       }
```
supertick, line 2016: getFullName() in subscriptions will break ServiceConfig.listeners - unless its fixed
That change is coming .. but there are some other filtering questions that need to be answered too

supertick (changes requested):

src/main/java/org/myrobotlab/framework/Service.java
AutonomicPerfectionist: Added a simple filter to `Service.getConfig()` that removes the local ID

supertick (approved):

PR #1227, opened 2022-12-23
Rename Service.getType() to Service.getTypeKey() and make webgui use it instead

### LINKED ISSUE #1212: WebGui relies on protected serviceClass field
**Describe the bug**
The WebGui currently uses `Service.serviceClass` to generate a panel, but there are three problems with using that field:

1. It is only present in classes directly inheriting from Service, not those that implement ServiceInterface in a different way
2. The field is protected with no public accessors, so the WebGui is technically relying on an implementation detail
3. There is an equivalent property that *is* present in ServiceInterface: `getType()`. However, this property is currently not serialized because it does not have a backing field and getter-only property serialization is disabled since Jackson misinterprets some methods as getters. Adding a backing field breaks downstream services that already have a "type" field.

These 3 problems combine to cause an impossibility for external services to work correctly with the WebGui, since they do not directly inherit from Service.

**Recommended Solution**
I recommend adding a `getServiceClass()` method to `ServiceInterface`, this would be the quickest and easiest way to solve the above problems. Eventually, I'd recommend enabling getter-only property serialization (and switching the WebGui to use getType() instead), but since this requires scouring the code and adding `@JsonIgnore` to everything that isn't a getter, it will take time

AutonomicPerfectionist:
Fixes #1212.

This is really only a temporary fix, a more permanent solution would be to choose either `getServiceClass()` or `getType()`, standardize on it, and ensure the chosen method is serialized correctly.

**Additional Context**
PR #1124 requires this for proper serialization of proxies for use in the webgui.

**EDIT**
This PR has been repurposed to rename `getType()` as `getTypeKey()` and standardize on that, removing `getServiceClass()`


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1100,6 +1100,17 @@ public String getType() {
...
+   * Equivalent to {@link #getType()} but required
+   * because the WebGui relies on this field.
+   *
+   * @return The fully qualified class name of this service.
+   */
+  @Override
+  public String getServiceClass() {
+    return serviceClass;
+  }
+
```
supertick: This is also used to pull the class name off of json "class" ?

supertick (changes requested):

AutonomicPerfectionist:
This PR has been redone, instead removing `serviceClass` entirely, using the newly-renamed `Service.getTypeKey()` instead. No more conflicts with downstream services, no more duplicated class fields. It will also be used to store the foreign type key of services in other languages

PR #1245, opened 2023-01-15
worky start.yml

supertick:


kwatters:
Why would you need it want a start.yml?  What is this for?

supertick:
To persist 2 data elements across restarts, id and config
https://discord.com/channels/887362610360643664/1009121322133098518/1064224889168465970

kwatters:
Seems like you should just pass the config that you want to start name on the command line...  and the "id" value should probably be in runtime.yml...

supertick:
It doesn't negate the capability of passing via command line ...  The command line should be a higher priority.
But the concept of this is to allow a restart to get back to previous state without the need of modifying command line.
Effectively, it should be have like this 
image

supertick:
but perhaps there needs to be an actual UI option to switch it off and on ...
:thinking: and then there should be a unit test to verify that command line still works and has a higher priority ...
I'll look into implementing these features/tests

PR #1248, opened 2023-01-20
mostly worky gpt3

supertick:


src/main/java/org/myrobotlab/service/config/Gpt3Config.java
```diff
@@ -0,0 +1,25 @@
+package org.myrobotlab.service.config;
+
+public class Gpt3Config extends ServiceConfig {
+
+  public String currentBotName = "Alice";
```
kwatters: Why alice?

supertick: There are probably vars and behaviors that probably should work the same way between ProgramAB and ChatGpt3.  
I guess this is a reminder to start heading in that direction.  Heh, a little experimentation came up with a non-personalized response. 
image

supertick: removing it for now .. but it would be nice to name it what you want ...

kwatters: Alice refers to a specific AIML set.  IF we want to give this one a name.. it should simply be GPT-3...

src/main/java/org/myrobotlab/service/Gpt3.java
```diff
@@ -0,0 +1,255 @@
...
+
+  public Gpt3(String n, String id) {
+    super(n, id);
+  }
+
+  public Response getResponse(String text) {
+
+    try {
+
+      Gpt3Config c = (Gpt3Config) config;
```
kwatters, line 72: I don't like that this needs to be type casted here on each response.

supertick: The framework sets the value ... all the services don't need to worry about handling the config
the getConfig/apply is handled completely by the framework.
To do so its a single field in the Service.java definition is needed.   If not this,
what would you suggest ?

src/main/java/org/myrobotlab/service/config/Gpt3Config.java
supertick: I just removed it for now ... there are some meaningful features  (sleep, wake, bot name, session name, user name)
which would be useful to implement - but this is only the initial implementation

PR #1251, opened 2023-02-18
testing jackson replacement of gson

supertick:


supertick:
Thanks @AutonomicPerfectionist  !

PR #1258, opened 2023-04-15
localspeech update

supertick:


lecagnois (approved):
Is ok

src/main/java/org/myrobotlab/service/LocalSpeech.java
```diff
@@ -302,6 +309,33 @@ public boolean setFestival() {
...
+    addVoice("de-DE", "female", "de-DE", "pico2wav");
+    addVoice("en-GB", "female", "en-GB", "pico2wav");
+    addVoice("en-US", "female", "en-US", "pico2wav");
+    addVoice("es-ES", "female", "es-ES", "pico2wav");
+    addVoice("fr-FR", "female", "fr-FR", "pico2wav");
+    addVoice("it-IT", "female", "it-IT", "pico2wav");
+
+    setTtsCommand("pico2wave -l {voice_name} -w {filename} \"{text}\" ");
+    if (!Runtime.getPlatform().isLinux()) {
+      error("festival only supported on Linux");
```
AutonomicPerfectionist: I think that should say pico2wav right?

src/main/java/org/myrobotlab/service/LocalSpeech.java
```diff
@@ -281,6 +284,8 @@ public void removeExt(boolean b) {
   public boolean setEspeak() {
     LocalSpeechConfig c = (LocalSpeechConfig)config;
     c.speechType = "Espeak";
+    voices.clear();
+    addVoice("Linus", "male", "en-US", "festival");
```
AutonomicPerfectionist: Not sure this is meant to be in setEspeak()

src/main/java/org/myrobotlab/service/LocalSpeech.java
```diff
@@ -302,6 +309,33 @@ public boolean setFestival() {
     }
     return true;
   }
+  
+  
+  /**
+   * @return setFestival sets the Linux tts to festival template
```
AutonomicPerfectionist: Should be setPico2wav. Also should actually say what the return value does, does it mean success?

src/main/resources/resource/WebGui/app/service/views/LocalSpeechGui.html
```diff
@@ -2,15 +2,19 @@
...
+        
         <select ng-model="service.config.speechType" ng-change="setType(service.config.speechType)" class="form-control" title="select a type of local voice control">
             <option ng-repeat="type in service.types" ng-value="{{type}}">{{type}}</option>
         </select>
-
-        
+        <div ng-show="service.config.speechType === 'Pico2Wav'">
+            <b>Install pico2wave</b><br/>
+            https://doc.ubuntu-fr.org/svoxpico<br/>
+            <pre>sudo apt-get install libttspico-utils</pre>
```
AutonomicPerfectionist: apt-get is considered legacy, most systems recommend plain `apt`

AutonomicPerfectionist (changes requested):

src/main/resources/resource/WebGui/app/service/views/LocalSpeechGui.html
supertick: i'm dating myself .. ok will change

src/main/java/org/myrobotlab/service/LocalSpeech.java
supertick: the "festival" is a copy paste error, but I did want to put a voice in each voice subtype, because our voice definition should be the way to further distinguish other voices or language support in the subtypes ... I did this to pico2wav so thought I would add at least 1 voice to each. Correcting the Espeak festival part though.  Hmm, i guess since we only have one (at the moment) we'll call it espeak voice .. :shrug: 
fixed

src/main/java/org/myrobotlab/service/LocalSpeech.java
supertick: yep fixed

src/main/java/org/myrobotlab/service/LocalSpeech.java
supertick: yup updated and hopefully more clear

AutonomicPerfectionist (approved):

PR #1260, opened 2023-04-16
OpenCV list valid cameras

supertick:


src/main/java/org/myrobotlab/service/OpenCV.java
```diff
@@ -714,6 +715,44 @@ public void capture(String filename) {
...
+    for (int i = 0; i < 8; i++) {
+      try {
+        FrameGrabber grabber = new OpenCVFrameGrabber(i);
+        grabber.start();
+        Frame frame = grabber.grab();
+        if (frame != null) {
+          cameraIndexes.add(i);
+        }
+        grabber.stop();
+        sleep(1000);
```
kwatters, line 743: this is excessive, no?  Is there no other way to list video capture devices ?   8 seconds minimum to run this method.  Also, the upper limit of 8 should probably be pulled out as a constant and pulled up as a member of this class instead .. this way a javadoc comment can be added easily so that we can describe the behavior better.

PR #1262, opened 2023-04-17
Py4j

supertick:


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -663,16 +665,20 @@ public Service(String reservedKey, String inId) {
     this.inbox = new Inbox(getFullName());
     this.outbox = new Outbox(this);
 
+    if (name.equals("i01")) {
```
kwatters: this is debugging code that probably shouldn't be included in this PR.

supertick: sure, removed

src/main/java/org/myrobotlab/framework/interfaces/MessageInvoker.java
```diff
@@ -0,0 +1,10 @@
+package org.myrobotlab.framework.interfaces;
+
+import org.myrobotlab.framework.Message;
+
+public interface MessageInvoker {
+  public Object invoke(Message msg);
+
+  public Object invokeOn(boolean blockLocally, Object obj, String method, Object... params);
```
AutonomicPerfectionist, line 24: Would like a Java doc on this method since there's so many parameters

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -26,40 +37,79 @@ public Py4j(String n, String id) {
...
+      gateway.addListener(this);
+      gateway.start();
+      info("server started listening on %s:%d", gateway.getAddress(), gateway.getListeningPort());
+      handler = (Invoker) gateway.getPythonServerEntryPoint(new Class[] { Invoker.class });
     } else {
       log.info("Py4j gateway server already started");
     }
   }
+  
+  public void newScript() {
```
AutonomicPerfectionist, line 51: As I understand this method will only open a new script.py file if one isn't open already. Wouldn't it be better if we were able to create multiple new scripts, maybe adding an index to the name to differentiate them (script2.py, etc)?

src/main/java/org/myrobotlab/framework/interfaces/MessageInvoker.java
supertick: yup .. updated

src/main/java/org/myrobotlab/service/Py4j.java
supertick: Yes, I agree ... this was copied over from the Python/Jython service .. I want them to be in sync.  An interface isn't currently defined, but should be in the future.  but I did add:

```java
  public void newScript(String scriptName) {
  ```
  I'm avoiding file operations though which has been a long thorn in the side ... for a new release Ogre .. because I'm hoping at that time the webgui will be under npm package manager and we can use a file operations library for user interaction to save/open/save as etc

File: src/main/java/org/myrobotlab/service/Py4j.java [hunk 4]
```diff
@@ -26,40 +37,83 @@ public Py4j(String n, String id) {
...
    */
   public void stop() {
-    if (server != null) {
-      server.shutdown();
-      server = null;
+    if (gateway != null) {
+      gateway.shutdown();
+      gateway = null;
     } else {
       log.info("Py4j gateway server already started");
```
AutonomicPerfectionist: Should say already stopped

AutonomicPerfectionist (approved):

src/main/java/org/myrobotlab/service/Py4j.java
supertick: :+1:

PR #1264, opened 2023-04-18
function to query working cameras of a system

supertick:


src/main/java/org/myrobotlab/service/OpenCV.java
```diff
@@ -715,6 +721,45 @@ public void captureFromResourceFile(String filename) throws IOException {
...
+    for (int i = 0; i < 8; i++) {
+      try {
+        FrameGrabber grabber = new OpenCVFrameGrabber(i);
+        grabber.start();
+        Frame frame = grabber.grab();
+        if (frame != null) {
+          cameraIndexes.add(i);
+        }
+        grabber.stop();
+        sleep(1000);
```
kwatters: this seems like an excessively long time to wait here...  do we really need a sleep here?

supertick: yeah .. took me a while to try to figure out if there was a correct blocking call that released the appropriate resources.
removed the sleep... now it uses stop() previously it used release() (wrong)  .. the thing that takes a huge amount of time is the failure ... ultimately, its a timeout error somewhere in opencv I suspect trying to read a frame that's not available

PR #1269, opened 2023-04-21
updates to the PCF8574 service and WebGUI

Cyber-One:


Cyber-One (commented):
This appears to be working, but not the prettiest UI

supertick:
i think this was covered in the cyber one recent merge

PR #1274, opened 2023-04-27
Auto start yml

supertick:
to be able to control and automatically start a config through the ui on next start


src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -598,9 +598,16 @@ static public boolean setAllVirtual(boolean b) {
   /**
    * start
    * @param b
+   * @throws IOException
```
AutonomicPerfectionist: Under what circumstances is it thrown? Also the javadoc should explain what auto starting means, especially what file it will be using

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -598,9 +598,16 @@ static public boolean setAllVirtual(boolean b) {
   /**
    * start
    * @param b
+   * @throws IOException 
    */
-  static void setAutoStart(boolean b) {
+  public void setAutoStart(boolean b) throws IOException {
+    log.error("setAutoStart {}", b);
```
AutonomicPerfectionist: Shouldn't be an error, maybe don't log at all

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1341,6 +1350,16 @@ public static synchronized List<ServiceInterface> getServicesFromInterface(Class
     }
     return ret;
   }
+  
+  /**
+   * Because startYml is required to be a static variable, since it's needed "before"
+   * a runtime instance exists it will be null in json serialization.  This method is needed
+   * so we can serialize the data appropriately.
```
AutonomicPerfectionist, line 1361: Has this been verified with both gson and jackson? I'm not sure both will pick up on the static getter. Quite frankly I'm not sure either will get it right, we had a lot of issues with Jackson thinking things were getters so I might've disabled that

AutonomicPerfectionist (changes requested):

src/main/java/org/myrobotlab/service/Runtime.java
supertick: updated :+1:

src/main/java/org/myrobotlab/service/Runtime.java
supertick: changed it to debug

src/main/java/org/myrobotlab/service/Runtime.java
supertick: verified it currently with jackson, was the case with gson as well..
in fact its the case we general serialization in Java ... probably not worth changing at this point as I'm sure other things will break

PR #1276, opened 2023-04-29
Programab get Predicates

supertick:
image


src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1452,4 +1466,19 @@ public Utterance publishUtterance(Utterance utterance) {
...
+    return response.msg;
+  }
+
+  @Override
+  public String translate(String text, String fromLang, String toLang) {
+    // FIXME implement - same mapping that current inmoov has ? en-US etc ..
+    // although all of those are localizations
+    // not languages only the 1st part is "en" in en-US
+    // return getResponse(text)
+    return null;
```
kwatters: this should be implemented.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1452,4 +1466,19 @@ public Utterance publishUtterance(Utterance utterance) {
     return utterance;
   }
 
+  @Override
+  public String translate(String text) {
+    Response response = getResponse(text);
```
kwatters: This should call the translate method with the max complexity args.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -608,6 +609,19 @@ public void reloadSession(String userName, String botName) throws IOException {
     }
   }
 
+  /**
+   * Get all current predicates names and their values
+   * for the current session
+   * @return
+   */
+  public Map<String, String> getPredicates() {
```
kwatters, line 616: there should be a getPredicates(session) method.. and this method should call that with the default session.

src/main/java/org/myrobotlab/service/interfaces/Translator.java
```diff
@@ -1,8 +1,29 @@
...
+   * Translate the incoming text from incoming language to target language This
+   * method will rely on the internal state of the translation service to
+   * determine incoming and target language
+   * 
+   * @param text
+   * @return
+   */
+  public String
+
+      translate(String text);
```
kwatters, line 15: This interface should allow you to set the source/ destination language .. o/w the interface won't support changing which language you are translating to by default.

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: agreed fixed

src/main/java/org/myrobotlab/service/interfaces/Translator.java
supertick: but thats really setting the bot

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: removed

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: removed

src/main/java/org/myrobotlab/service/interfaces/Translator.java
supertick: removed

PR #1277, opened 2023-04-29
Small cleaning pr - look in commits for info

supertick:


src/main/java/org/myrobotlab/service/Arduino.java
```diff
@@ -365,20 +365,20 @@ public void attachI2CControl(I2CControl control) {
     // the i2c bus here and in MrlComm
     // This will only handle the creation of i2cBus.
     if (i2cBus == null) {
-      i2cBus = new I2CBus(String.format("I2CBus%s", control.getDeviceBus()));
-      i2cBusAttach(i2cBus, Integer.parseInt(control.getDeviceBus()));
+      i2cBus = new I2CBus(String.format("I2CBus%s", control.getBus()));
```
AutonomicPerfectionist, line 368: What's with the changes from getDeviceBus/Address to getBus/Address? Is there any difference between them or just a naming change?

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1496,4 +1497,19 @@ public TopicChange publishTopic(TopicChange topicChange) {
   
   
 
+  @Override
+  public String translate(String text) {
```
AutonomicPerfectionist: What determines the to and from languages for this translation call?

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1496,4 +1497,19 @@ public TopicChange publishTopic(TopicChange topicChange) {
...
 
+  @Override
+  public String translate(String text) {
+    Response response = getResponse(text);
+    return response.msg;
+  }
+
+  @Override
+  public String translate(String text, String fromLang, String toLang) {
+    // FIXME implement - same mapping that current inmoov has ? en-US etc ..
```
AutonomicPerfectionist: This comes from a more general translation interface, and I think having localization strings for translation is beneficial since different regions have different dialects. Like US English and British English

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1496,4 +1497,19 @@ public TopicChange publishTopic(TopicChange topicChange) {
...
+    // although all of those are localizations
+    // not languages only the 1st part is "en" in en-US
+    // return getResponse(text)
+    return null;
```
AutonomicPerfectionist: If possible I think a full implementation would be better than a stub returning null, but at the same time I'm not totally sure how to do that with program AB

src/main/java/org/myrobotlab/service/Arduino.java
supertick: naming change only ...

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: pre existing config in the translator ...  setFromLanguage("en") setToLanguage("fr")

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: Sure, "if supported" in the implementation .. since its free form string input nothing prevents it at this level

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: Yeah ... I'll need to address this in the short future ..
As @kwatters mentioned, the translation layer will be implemented as an overlay with key templates and a key driven bot.
I'm still not completely sure if ProgramAB should or should not support translation semantics ..

PR #1278, opened 2023-04-30
ProgramAB now publishes topic

supertick:


src/main/java/org/myrobotlab/service/data/TopicChange.java
```diff
@@ -0,0 +1,21 @@
+package org.myrobotlab.service.data;
+
+public class TopicChange {
```
AutonomicPerfectionist, line 12: Would like javadoc on this class and all its members

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1320,21 +1321,27 @@ synchronized public void onChangePredicate(Chat chat, String predicateName, Stri
...
+        invoke("publishPredicate", s, chat, predicateName, result);
         s.savePredicates();
         return;
       }
     }
     error("could not find session to save predicates");
   }
 
-  public PredicateEvent publishChangePredicate(Session session, Chat chat, String name, String value) {
+  public PredicateEvent publishPredicate(Session session, Chat chat, String name, String value) {
```
AutonomicPerfectionist: Why the rename? Would this break any services or scripts subscribing to it?

Also would appreciate a Java doc on it, having difficulty figuring out what the parameters do

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1320,21 +1321,27 @@ synchronized public void onChangePredicate(Chat chat, String predicateName, Stri
...
     PredicateEvent event = new PredicateEvent();
     event.id = String.format("%s<->%s", session.userName, session.botInfo.name);
     event.userName = session.userName;
     event.botName = session.botInfo.name;
     event.name = name;
     event.value = value;
+    
+    if (name.equals("topic")) {
```
AutonomicPerfectionist: Should probably do `"topic".equals(name)` to avoid NPE

src/main/java/org/myrobotlab/service/data/TopicChange.java
supertick: fixed

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: agree

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: Checked nothing depends on it, and its new functionality I've accounted for in the UI
Added docs

AutonomicPerfectionist: Not sure changing the name has enough of a benefit, considering it's publishing a change of a predicate. User scripts may rely on it too (unlikely, but the possibility is there)

supertick: good point, but its relatively new .. and in the discussion with @kwatters we agreed on publishTopic .. so it would be following the same convention

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1320,21 +1321,35 @@ synchronized public void onChangePredicate(Chat chat, String predicateName, Stri
...
+  public PredicateEvent publishPredicate(Session session, String name, String value) {
     PredicateEvent event = new PredicateEvent();
     event.id = String.format("%s<->%s", session.userName, session.botInfo.name);
     event.userName = session.userName;
     event.botName = session.botInfo.name;
     event.name = name;
     event.value = value;
+    
+    if ("topic".equals(name)) {
+      invoke("publishTopic", new TopicChange(session.userName, session.botInfo.name, value, session.currentTopic));
```
kwatters, line 1349: we should probably also test that the new value is actually different from the old value here, o/w we could falsely detect a change in topic, maybe?

supertick: updated to filter out non changes

PR #1279, opened 2023-04-30
Sorted and removed a few not used inmoov methods 

supertick:
Process the formatting only pr first !
Some refactoring and removal of methods.
Added switch statement for peer processing


src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -880,21 +875,21 @@ public void moveHead(Integer neck, Integer rothead, Integer rollNeck) {
...
-  public void moveHeadBlocking(Double neck, Double rothead, Double eyeX, Double eyeY, Double jaw) {
+  public void moveHeadBlocking(Double neck, Double rothead, Double eyeX, Double eyeY, Double jaw) throws InterruptedException, TimeoutException {
     moveHeadBlocking(neck, rothead, eyeX, eyeY, jaw, null);
   }
 
-  public void moveHeadBlocking(Double neck, Double rothead, Double eyeX, Double eyeY, Double jaw, Double rollNeck) {
-    // the "right" way
-    sendToPeer("head", "moveToBlocking", neck, rothead, eyeX, eyeY, jaw, rollNeck);
+  public void moveHeadBlocking(Double neck, Double rothead, Double eyeX, Double eyeY, Double jaw, Double rollNeck) throws InterruptedException, TimeoutException {
+    sendBlocking(getPeerName("head"), "moveToBlocking", neck, rothead, eyeX, eyeY, jaw, rollNeck);
```
AutonomicPerfectionist: Bubbling up the checked exceptions is probably not needed; the try/catch can probably be done here at a minimum if not in sendBlocking itself

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -880,21 +875,21 @@ public void moveHead(Integer neck, Integer rothead, Integer rollNeck) {
...
+    log.info("here");
```
AutonomicPerfectionist: Info log probably should be removed

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -988,8 +983,9 @@ public void onJointAngles(Map<String, Double> angleMap) {
 
   @Override
   public void onJoystickInput(JoystickData input) throws Exception {
-    // TODO Auto-generated method stub
-
+    // TODO timer ? to test and not send an event
+    // switches to manual control ?
+    invoke("publishEvent", "joystick");
```
AutonomicPerfectionist, line 884: What's the purpose of this?

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -1092,83 +1098,113 @@ public void onStarted(String name) {
...
+        configStarted = true;
       }
 
-      // mapping a channel to a single end point depending on peer
-      actualName = getPeerName("rightArm");
-      if (actualName.equals(name)) {
-        addListener("publishMoveRightArm", actualName, "onMoveArm");
+      invoke("publishEvent", "STARTED " + peerKey);
+
+      switch (peerKey) {
```
AutonomicPerfectionist, line 1009: From what I can tell, this is autowiring peers together. Can this not be handled by the config system?

src/main/java/org/myrobotlab/service/InMoov2.java
supertick: the library method sendBlocking always need to throw on timeout, but I agree this stuff doesn't

src/main/java/org/myrobotlab/service/InMoov2.java
supertick: done

src/main/java/org/myrobotlab/service/InMoov2.java
supertick: the onJoystickInput is required, because  InMoov2 is a joystick listener.
publishEvent is a generalized InMoov2 publish which will help in managing its state machine
I don't know exactly what will happen in InMoov2 Joystick interactions .. alot would need to be defined by he user.
The user should be getting the actual Joystick data from the Joystick service, but I wanted to make the event available if it helps in managing the state of InMoov.
e.g. switching from gestures to manual mode (locking out gestures) :shrug:

src/main/java/org/myrobotlab/service/InMoov2.java
supertick: yup .. exactly ..
this was started before config was capable .. it should be removed, but that will come in a later pr

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -182,16 +183,16 @@ public static void main(String[] args) {
 
   protected List<String> configList;
 
+  /**
+   * Configuration from runtime has started. This is when runtime starts
+   * processing a configuration set for the first time since inmoov was started
+   */
+  boolean configStarted = false;
```
AutonomicPerfectionist: Should probably have visibility modifiers on this and all the other fields

supertick: added modifiers and removed unused members

AutonomicPerfectionist (approved):

PR #1282, opened 2023-05-03
moving serviceData.json from data/.myrobotlab to  libraries

supertick:


supertick:
worky... tested locally

ls -alh libraries/serviceData.json 
-rw-rw-r-- 1 139K May  3 06:43 libraries/serviceData.json
 ls -alh data/.myrobotlab/
total 8.0K
drwxrwxr-x 2  4.0K May  3 06:42 .
drwxrwxr-x 5  4.0K May  3 06:42 ..

image

src/main/java/org/myrobotlab/framework/repo/MavenWrapper.java
```diff
@@ -280,8 +280,10 @@ public static void main(String[] args) {
     try {
 
       LoggingFactory.init(Level.INFO);
-
-      File cache = new File("./data/.myrobotlab/serviceData.json");
+      
+      File libraries = new File("libraries");
+      libraries.mkdir();
+      File cache = new File("libraries/serviceData.json");
```
AutonomicPerfectionist: It's probably fine but I've heard you're supposed to use File.separator instead of slashes. As I understand Windows will interpret forward slashes as backslashes anyway though

src/main/java/org/myrobotlab/framework/repo/ServiceData.java
```diff
@@ -53,7 +53,7 @@ public class ServiceData implements Serializable {
 
   private static final long serialVersionUID = 1L;
 
-  static private String serviceDataCacheFileName = FileIO.getCfgDir() + File.separator + "serviceData.json";
+  static private String serviceDataCacheFileName = "libraries" + File.separator + "serviceData.json";
```
AutonomicPerfectionist: There should probably be a constant somewhere to store the string "libraries", don't want a typo to cause problems

src/main/java/org/myrobotlab/framework/repo/MavenWrapper.java
supertick: fixed

src/main/java/org/myrobotlab/framework/repo/ServiceData.java
supertick: fixed

PR #1286, opened 2023-05-18
Solr92

kwatters:
solr 9.2 upgrade with associated program-ab upgrade (to resolve conflicting versions of lucene analysis libraries.)

### DIFF
--- a/src/main/java/org/myrobotlab/service/meta/SolrMeta.java
+++ b/src/main/java/org/myrobotlab/service/meta/SolrMeta.java
@@ -16,8 +16,8 @@ public class SolrMeta extends MetaData {
 
     addDescription("Solr Service - Open source search engine");
     addCategory("search");
-    String solrVersion = "8.11.2";
-    String luceneVersion = solrVersion;
+    String solrVersion = "9.2.0";
+    String luceneVersion = "9.4.2";
     addDependency("org.apache.lucene", "lucene-core", luceneVersion);
     addDependency("org.apache.solr", "solr-core", solrVersion);
     exclude("log4j", "*");
--- a/src/main/resources/resource/Solr/core1/conf/managed-schema
+++ b/src/main/resources/resource/Solr/core1/conf/managed-schema
@@ -414,6 +414,10 @@
       <tokenizer class="solr.WhitespaceTokenizerFactory"/>
     </analyzer>
   </fieldType>
+  
+  <!-- Adding a dense vector field for knn search to do cosine similarity ( 384 dimensions and cosine similarity matches with al-->
+  <fieldType name="knn_vector" class="solr.DenseVectorField" vectorDimension="384" similarityFunction="cosine" knnAlgorithm="hnsw" hnswMaxConnections="10" hnswBeamWidth="40"/>
+  
   <field name="_root_" type="string" docValues="false" indexed="true" stored="false"/>
   <field name="_text_" type="text_general" multiValued="true" indexed="true" stored="false"/>
   <field name="_version_" type="plong" indexed="false" stored="false"/>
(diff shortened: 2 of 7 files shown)


src/main/java/org/myrobotlab/service/meta/SolrMeta.java
```diff
@@ -16,8 +16,8 @@ public SolrMeta() {
 
     addDescription("Solr Service - Open source search engine");
     addCategory("search");
-    String solrVersion = "8.11.2";
-    String luceneVersion = solrVersion;
+    String solrVersion = "9.2.0";
+    String luceneVersion = "9.4.2";
```
AutonomicPerfectionist, line 20: When I updated Solr on my own branch I needed to add the following here:

```
addDependency("org.apache.solr", "solr-scripting", solrVersion);
exclude("com.google.guava", "*");
```

From what I remember, one of the classes used in the solrconfig.xml was moved to this artifact, but I can't remember off the top of my head which one. Do you think this is necessary? I can find which class it was by looking through a diff of my solconfig.xml if you need more info

AutonomicPerfectionist: Nevermind, looks like it was `XSLTResponseWriter` which your changes have removed from the solrconfig.xml

src/main/resources/resource/Solr/core1/conf/managed-schema
```diff
@@ -414,6 +414,10 @@
       <tokenizer class="solr.WhitespaceTokenizerFactory"/>
     </analyzer>
   </fieldType>
+  
+  <!-- Adding a dense vector field for knn search to do cosine similarity ( 384 dimensions and cosine similarity matches with al-->
+  <fieldType name="knn_vector" class="solr.DenseVectorField" vectorDimension="384" similarityFunction="cosine" knnAlgorithm="hnsw" hnswMaxConnections="10" hnswBeamWidth="40"/>
```
AutonomicPerfectionist, line 419: This is what I ended up doing on my branch as well, but I was wondering if there were a better way to accomplish it. Tying ourselves to a fixed vector dimension limits the models we can support with Solr, is there a way to dynamically define the type according to the vector dimension? If not, can we define multiple types for each of the most common dimensions? Also, do we want to use cosine similarity or can we force normalization before indexing so we can use the faster dot product on search?

src/test/java/org/myrobotlab/service/SolrTest.java
```diff
@@ -123,9 +128,42 @@ public void testService() throws Exception {
...
+    
+    query = new SolrQuery("{!knn f=vector topK=10}"+queryVec);
+    resp = solr.search(query);
+    
+    // System.out.println(resp);
+    Assert.assertEquals(1, resp.getResults().getNumFound());
+
+  }
 
+  private String vecToString(ArrayList<Float> v) {
```
AutonomicPerfectionist, line 146: This looks to me like it's reimplementing the standard `List.toString()`. Any particular reason for it, other than ensuring we aren't reliant on potentially overridden `toString()` implementations?

src/main/resources/resource/Solr/core1/conf/managed-schema
kwatters: Good news is that we can programmatically add fields to the schema.  I chose 384 dimensions and cosine similarity because that lines up with the models that are commonly used. ( miniLM-L6-V2 ). https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2. 
Cosine similarity is much more popular compared to dot product similarity.

src/test/java/org/myrobotlab/service/SolrTest.java
kwatters: This is just the unit test to create the proper syntax.  We could create a vector class to hold these embeddings.. but ultimately it's a float array.  

The goal of this PR was really just the upgrade and a small example of doing a knn query.

PR #1287, opened 2023-05-21
Fix incorrect ImageData import in Email service

AutonomicPerfectionist:
This PR is pretty simple; it just changes the import for ImageData to `org.myrobotlab.service.data.ImageData`


supertick:
Thanks AP .. got it :+1:

PR #1292, opened 2023-06-01
Py4j Worky !!!

supertick:
image

### DIFF
--- a/src/main/java/org/myrobotlab/framework/interfaces/Executor.java
+++ b/src/main/java/org/myrobotlab/framework/interfaces/Executor.java
@@ -1,9 +1,0 @@
-package org.myrobotlab.framework.interfaces;
-
-public interface Executor extends Invoker {
-
-  public Object exec(String code);
-  
-  public String setName(String name);
-
-}
--- a/src/main/java/org/myrobotlab/framework/interfaces/Executor.java
+++ b/src/main/java/org/myrobotlab/framework/interfaces/Executor.java
@@ -1,0 +1,9 @@
+package org.myrobotlab.framework.interfaces;
+
+public interface Executor extends Invoker {
+
+  public Object exec(String code);
+  
+  public String setName(String name);
+
+}
--- a/src/main/java/org/myrobotlab/io/StreamGobbler.java
+++ b/src/main/java/org/myrobotlab/io/StreamGobbler.java
@@ -19,12 +19,25 @@ public class StreamGobbler extends Thread {
   protected String name;
   public final static Logger log = LoggerFactory.getLogger(StreamGobbler.class);
 
+  /**
+   * When we do not need to redirect the stream to another stream. E.g.
+   * when we create a process, and are already redirecting std:out and std:err
+   * 
+   * @param name
+   * @param processOut
+   */
   public StreamGobbler(String name, InputStream processOut) {
-    super(name);
-    this.processOut = processOut;
-    this.name = name;
+    this(name, processOut, null);
   }
 
+  /**
+   * This is a general stream gobbler that will consume and input stream and move 
+   * the data to an output stream.
+   * 
+   * @param name
+   * @param processOut
+   * @param processIn
+   */
   public StreamGobbler(String name, InputStream processOut, OutputStream processIn) {
     super(name);
     this.processOut = processOut;
--- a/src/main/java/org/myrobotlab/io/StreamGobbler.java
+++ b/src/main/java/org/myrobotlab/io/StreamGobbler.java
@@ -6,10 +6,24 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 
+import org.myrobotlab.logging.LoggerFactory;
+import org.slf4j.Logger;
+/**
+ * A general stream gobbler, useful when starting processes and handling their streams
+ * @author GroG
+ *
+ */
 public class StreamGobbler extends Thread {
   protected transient InputStream processOut;
   protected transient OutputStream processIn;
   protected String name;
+  public final static Logger log = LoggerFactory.getLogger(StreamGobbler.class);
+
+  public StreamGobbler(String name, InputStream processOut) {
+    super(name);
+    this.processOut = processOut;
+    this.name = name;
+  }
 
   public StreamGobbler(String name, InputStream processOut, OutputStream processIn) {
     super(name);
@@ -24,14 +38,14 @@ public class StreamGobbler extends Thread {
       InputStreamReader isr = new InputStreamReader(processOut);
       BufferedReader br = new BufferedReader(isr);
       String line = null;
-      while ((line = br.readLine()) != null)
-        if (line != null) {
+      while ((line = br.readLine()) != null) {
+        log.info("gobbler - {}", line);
+        if (line != null && processIn != null) {
           processIn.write(String.format("%s\n", line).getBytes());
           processIn.flush();
-        }
+        }}
     } catch (IOException ioe) {
-      System.out.println("gobbler leaving");
-      ioe.printStackTrace();
+      log.info("{} gobbler leaving", name);
     }
   }
 }
--- a/src/main/java/org/myrobotlab/service/Py4j.java
+++ b/src/main/java/org/myrobotlab/service/Py4j.java
@@ -11,13 +11,13 @@ import java.util.Set;
 
 import org.myrobotlab.framework.Message;
 import org.myrobotlab.framework.Service;
-import org.myrobotlab.framework.interfaces.Executor;
 import org.myrobotlab.io.FileIO;
 import org.myrobotlab.io.StreamGobbler;
 import org.myrobotlab.logging.Level;
 import org.myrobotlab.logging.LoggerFactory;
 import org.myrobotlab.logging.LoggingFactory;
 import org.myrobotlab.service.data.Script2;
+import org.myrobotlab.service.interfaces.Executor;
 import org.slf4j.Logger;
 
 import py4j.GatewayServer;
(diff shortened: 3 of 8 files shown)


src/main/java/org/myrobotlab/framework/interfaces/Executor.java
```diff
@@ -0,0 +1,9 @@
+package org.myrobotlab.framework.interfaces;
+
+public interface Executor extends Invoker {
```
AutonomicPerfectionist: Should be documented, and maybe not in Framework since it seems to be Py4j specific. Alternatively, it seems this interface could be implemented by the Python and Py4j services with zero issues, so it could be put in the service.interfaces package

src/main/java/org/myrobotlab/io/StreamGobbler.java
```diff
@@ -6,10 +6,24 @@
...
+ * @author GroG
+ *
+ */
 public class StreamGobbler extends Thread {
   protected transient InputStream processOut;
   protected transient OutputStream processIn;
   protected String name;
+  public final static Logger log = LoggerFactory.getLogger(StreamGobbler.class);
+
+  public StreamGobbler(String name, InputStream processOut) {
```
AutonomicPerfectionist, line 29: If keeping processIn null is intentional, that should be documented somewhere and probably explicitly set to null in this constructor

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -1,148 +1,352 @@
...
 import org.slf4j.Logger;
+
+import py4j.GatewayServer;
 import py4j.GatewayServerListener;
 import py4j.Py4JServerConnection;
-import py4j.GatewayServer;
 
+/**
+ * 
+ * @author GroG
```
AutonomicPerfectionist: The `@author` needs to be at the bottom, otherwise the entire content of the javadoc will be put as the "author"

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -1,148 +1,352 @@
...
+    public Process process;
+    public Py4j py4j;
+
+    public WaitForProcess(Py4j py4j, Process process) {
+      super(String.format("%s-process-signal", py4j.getName()));
+      this.process = process;
+      this.py4j = py4j;
+    }
+
+    public void run() {
```
AutonomicPerfectionist, line 102: Should have an `@Override`

src/main/java/org/myrobotlab/service/data/Script2.java
```diff
@@ -0,0 +1,21 @@
+package org.myrobotlab.service.data;
+
+import java.io.Serializable;
+
+public class Script2 implements Serializable {
```
AutonomicPerfectionist, line 5: Why is it named Script2? And it should be documented

src/main/resources/resource/Py4j/Py4j.py
```diff
@@ -30,33 +20,70 @@
...
+        """
+        try:
+            exec(code)
+        except Exception as e:
+            print(e)
+
     def invoke(self, method, data=None):
         # convert to list
         params = list(data)
         eval(method)(*params)
```
AutonomicPerfectionist, line 77: This `eval()` on line 77 should probably be replaced with a lookup into the `globals()` dictionary, eval() is slow and is considered bad practice since it can execute arbitrary code

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -1,148 +1,352 @@
...
+   * POJO class to tie all the data elements of a external python process
+   * together. Including the process handler, the std out, std err streams and
+   * termination signal thread.
+   * 
+   * @author GroG
+   *
+   */
+  class Py4jClient {
+    // py4j connection
+    public transient Py4JServerConnection connection;
```
AutonomicPerfectionist, line 62: Do these fields need to be public?

src/main/java/org/myrobotlab/io/StreamGobbler.java
```diff
@@ -24,14 +38,14 @@ public void run() {
...
+        log.info("gobbler - {}", line);
+        if (line != null && processIn != null) {
           processIn.write(String.format("%s\n", line).getBytes());
           processIn.flush();
-        }
+        }}
     } catch (IOException ioe) {
-      System.out.println("gobbler leaving");
-      ioe.printStackTrace();
+      log.info("{} gobbler leaving", name);
```
AutonomicPerfectionist, line 61: Is catching an IOException not considered an error here?

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -1,148 +1,352 @@
...
+  }
+
+  /**
+   * A class to wait on a process signal and notify client is disconnected
+   * 
+   * @author GroG
+   *
+   */
+  public class WaitForProcess extends Thread {
+    public Integer exitCode;
```
AutonomicPerfectionist, line 91: These probably don't need to be public, if they need to be accessed later it should be through a getter

src/main/resources/resource/Py4j/Py4j.py
```diff
@@ -30,33 +20,70 @@
...
 class MessageHandler(object):
 
     def __init__(self):
+        global runtime
         # initializing stdout and stderr
+        self.name = None
         self.stdout = sys.stdout
         self.stderr = sys.stderr
         sys.stdout = self
         sys.stderr = self
```
AutonomicPerfectionist, line 34: This merges stdout and stderr into one stream, erasing where the message originally came from. It would probably be better to have two separate handlers that each invoke either `handleStdOut` or a new `handleStdErr`. Also, it might be beneficial to output to stdout and stderr as well as to the java side, so I think we should have the streams cloned instead of replaced

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -1,148 +1,352 @@
...
+      super(String.format("%s-process-signal", py4j.getName()));
+      this.process = process;
+      this.py4j = py4j;
+    }
+
+    public void run() {
+      try {
+        exitCode = process.waitFor();
+      } catch (InterruptedException e) {
+      }
```
AutonomicPerfectionist, line 106: Interruption should at least be logged

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -168,16 +372,138 @@ public void serverStarted() {
...
+      // sets the name of the message handler
+      // calling the handler when the connection is not established
+      handler.setName(getName());
+
+    } catch (Exception e) {
+      error(e);
+    }
   }
 
+  public void startService() {
```
AutonomicPerfectionist, line 437: Should have `@Override`

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -168,16 +372,138 @@ public void serverStarted() {
...
+    start();
+    sleep(300);
+    // start the python process which starts the Py4j.py MessageHandler
+    startPythonProcess();
+  }
 
+  /**
+   * stop the gateway service and teardown of the python process
+   */
+  public void stop() {
```
AutonomicPerfectionist, line 452: Does this need `@Override`?

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -168,16 +372,138 @@ public void serverStarted() {
...
+    if (pythonProcess != null) {
+      log.info("shutting down python process");
+      pythonProcess.process.destroy();
+    }
+  }
+
+  /**
+   * shutdown cleanly
+   */
+  public void stopService() {
```
AutonomicPerfectionist, line 473: `@Override`

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -1,148 +1,352 @@
...
-      gateway = new GatewayServer(this);
-      gateway.addListener(this);
-      gateway.start();
-      info("server started listening on %s:%d", gateway.getAddress(), gateway.getListeningPort());
-      handler = (Invoker) gateway.getPythonServerEntryPoint(new Class[] { Invoker.class });
-    } else {
-      log.info("Py4j gateway server already started");
-    }
-  }
+  public void addScript(String scriptName, String code) throws IOException {
```
AutonomicPerfectionist, line 163: Can this be pulled up into a common interface for both Py4j and Python? Along with the other script-related methods

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -1,148 +1,352 @@
...
+    clients.remove(getClientKey(gatewayConnection));
+    invoke("getClients");
+  }
+
+  /**
+   * One of 3 methods supported on the MessageHandler() callbacks
+   * 
+   * @param code
+   */
+  public void exec(String code) {
```
AutonomicPerfectionist, line 216: Either pull up to a common interface or implement Executor

AutonomicPerfectionist (changes requested):

src/main/java/org/myrobotlab/service/data/Script2.java
supertick: When refactoring, I didn't want to affect Python so I made a Script2 ... its resolved in the next pr

src/main/java/org/myrobotlab/service/Py4j.java
supertick: nope

src/main/java/org/myrobotlab/service/Py4j.java
supertick: its duplicated in the next pr .. but ya that would be the evolution, or a more general file operations - but the more generalized one will need to wait for some more dependencies to be available...

src/main/java/org/myrobotlab/service/Py4j.java
supertick: Yeah I'd like to see how deployment of Py4j and the roadmap for Jython/Python resolve first ...

PR #1293, opened 2023-06-02
chat completions with Gpt3 service

supertick:


supertick:
image
working with dropped in llama docker

src/main/java/org/myrobotlab/service/Gpt3.java
```diff
@@ -99,12 +109,28 @@ public Response getResponse(String text) {
         @SuppressWarnings({ "unchecked", "rawtypes" })
         Map<String, Object> payload = (Map) CodecUtils.fromJson(msg, LinkedHashMap.class);
```
AutonomicPerfectionist: This can be done without casting using StaticType. Like so:
```java
Map<String, Object> payload = CodecUtils.fromJson(msg, new StaticType{}());
```

But, a better way would be to create a POJO that maps directly to the response format, eliminating needs for later casts as well

src/main/java/org/myrobotlab/service/Gpt3.java
```diff
@@ -141,6 +167,22 @@ public Response getResponse(String text) {
     return null;
   }
   
+  /**
+   * Overriden error to also publish the errors
+   * probably would be a better solution to self subscibe to errors and 
+   * have the subscriptions publish utterances/responses/text
+   */
+  public Status error(String error) {
```
AutonomicPerfectionist, line 175: Should have `@Override`

src/main/java/org/myrobotlab/service/config/Gpt3Config.java
```diff
@@ -13,9 +13,9 @@ public class Gpt3Config extends ServiceConfig {
 
   public int maxTokens = 256;
   public float temperature = 0.7f;
-  public String url = "https://api.openai.com/v1/completions";
+  public String url = "https://api.openai.com/v1/chat/completions";
```
AutonomicPerfectionist, line 16: Does changing this also change the default YAML config file?

src/main/java/org/myrobotlab/service/Gpt3.java
```diff
@@ -141,6 +167,22 @@ public Response getResponse(String text) {
...
   
+  /**
+   * Overriden error to also publish the errors
+   * probably would be a better solution to self subscibe to errors and 
+   * have the subscriptions publish utterances/responses/text
+   */
+  public Status error(String error) {
+    Status status = super.error(error);
+    invoke("publishText", error);    
+    Response response = new Response("friend", getName(), error, null);
```
AutonomicPerfectionist, line 178: Why "friend" here?

src/main/java/org/myrobotlab/service/Gpt3.java
```diff
@@ -99,12 +109,28 @@ public Response getResponse(String text) {
...
           Map<String, Object> textObject = (Map) choices.get(0);
           responseText = (String) textObject.get("text");
-          invoke("publishText", responseText);
+          if (responseText != null) {
+            // /completions
+            invoke("publishText", responseText);
+          } else {
+            // /chat/completions
+            @SuppressWarnings({ "unchecked", "rawtypes" })
+            Map<String, Object> content = (Map)textObject.get("message");
```
AutonomicPerfectionist, line 128: You can cast to generic type, like so:
```java
content = (Map<String, Object>) textObject.get("message");
```

I still vote for a POJO based approach but this at least would get rid of using raw types

src/main/java/org/myrobotlab/service/Gpt3.java
supertick: yup

src/main/java/org/myrobotlab/service/config/Gpt3Config.java
supertick: yes it does

src/main/java/org/myrobotlab/service/Gpt3.java
supertick: initially, I was excited to see they have an openapi.yml
but I tried a myriad of different versions of open-api generator and multiple versions of the spec release 2.0.0 and master
nothing worked without lots of errors when it generated the spec 
image

supertick: you mean 
```import org.myrobotlab.framework.StaticType;```
does not seem to work ...
image

AutonomicPerfectionist: Needs to be `new StaticType<>() {}`

Also, you need to remove the `(Map)` cast

src/main/java/org/myrobotlab/service/Gpt3.java
supertick: its a ProgramAB Response object ... until it gets promoted to something more generalized.
And the Response object expect to know the "user" the bot is chatting with ... default is friend

src/main/java/org/myrobotlab/service/Gpt3.java
supertick: still gives you a warning then for unchecked ... yeah POJOs a later pr

PR #1294, opened 2023-06-03
Fixes from Py4j back to Python

supertick:


src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -270,7 +273,7 @@ public Object get(String pythonRefName) {
   Map<String, String> exampleFiles = new TreeMap<String, String>();
 
   transient LinkedBlockingQueue<Message> inputQueue = new LinkedBlockingQueue<Message>();
-  final transient InputQueue inputQueueThread;
+  transient InputQueue inputQueueThread;
```
AutonomicPerfectionist: Should stay final, you can initialize it here even though it references `this`, like so:
```java
final transient InputQueue inputQueueThread = new InputQueue(this);
```

src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -871,54 +871,91 @@ public void stopService() {
     // release the interpeter
     stop();
   }
+  
+  public void init() {
```
AutonomicPerfectionist, line 879: Needs a javadoc

src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -871,54 +871,91 @@ public void stopService() {
     // release the interpeter
     stop();
   }
+  
+  public void init() {
+    
+    log.info("created python {}", getName());
+    createPythonInterpreter();
+    sleep(250);
```
AutonomicPerfectionist, line 883: What's this sleep for? Should have at least a comment explaining it, hopefully it can be removed eventually

src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -871,54 +871,91 @@ public void stopService() {
...
     stop();
   }
+  
+  public void init() {
+    
+    log.info("created python {}", getName());
+    createPythonInterpreter();
+    sleep(250);
+    
+    inputQueueThread = new InputQueue(this);
```
AutonomicPerfectionist: Can be put in the initializer as stated in an above comment

src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -871,54 +871,91 @@ public void stopService() {
...
-      Python python = (Python) Runtime.start("python", "Python");
-      // python.execFile("data/adafruit.py");
+    attachPythonConsole();
 
-      // Runtime.start("i01", "InMoov2");
+    String selfReferenceScript = "from time import sleep\nfrom org.myrobotlab.framework import Platform\n" + "from org.myrobotlab.service import Runtime\n"
+        + "from org.myrobotlab.framework import Service\n" + "from org.myrobotlab.service import Python\n"
+        + String.format("%s = Runtime.getService(\"%s\")\n\n", CodecUtils.getSafeReferenceName(getName()), getName()) + "Runtime = Runtime.getInstance()\n\n"
+        + String.format("runtime = Runtime.getInstance()\n") + String.format("myService = Runtime.getService(\"%s\")\n", getName());
+    // FIXME !!! myService is SO WRONG it will collide on more than 1 python
```
AutonomicPerfectionist, line 904: Is the `myService` variable still needed? As far as I've seen it's never used and instead the `python` variable or service is used

src/main/resources/resource/PythonProxy/PythonProxy.py
```diff
@@ -1,2 +0,0 @@
-# TODO: Implement this script fpr
-pythonproxy = runtime.start("pythonproxy","PythonProxy")...
```
AutonomicPerfectionist, line 2: Didn't know that was still in there, wonder if there's still code elsewhere for that....

src/main/java/org/myrobotlab/service/data/Script.java
```diff
@@ -1,58 +1,21 @@
...
-import java.io.File;
 import java.io.Serializable;
 
 public class Script implements Serializable {
   static final long serialVersionUID = 1L;
   /**
    * unique location &amp; key of the script e.g. /mrl/scripts/myScript.py
    */
-  File file;
+  public String file;
```
AutonomicPerfectionist, line 19: Should probably have getters, and if mutability is not needed then this and `code` should be final with no setters

src/main/java/org/myrobotlab/service/data/Script.java
```diff
@@ -1,58 +1,21 @@
 package org.myrobotlab.service.data;
 
-import java.io.File;
 import java.io.Serializable;
 
 public class Script implements Serializable {
```
AutonomicPerfectionist, line 13: Probably needs javadoc and probably needs to implement equals() and hashcode(), just to be safe

src/main/java/org/myrobotlab/service/Python.java
supertick: its asynchronous initialization ... if i remember correctly :-1:

src/main/java/org/myrobotlab/service/Python.java
supertick: really it should start only with runtime ... this is the same behavior as py4j .. ie runtime reference is all you need to get things started ...
Will remove..

src/main/resources/resource/PythonProxy/PythonProxy.py
supertick: its gone ...  this just a little cleanup

src/main/java/org/myrobotlab/service/Python.java
supertick: ah .. it is needed, its need to handle std:out from the jython ... kind of a critical piece 
This is the same for Py4j, in general the python needs to know the service name to hijack the std:out std:err of the python
In Py4j I had to create a setName on the Executor interface so the Python process can know which mrl service is responsible for creating it.

src/main/java/org/myrobotlab/service/data/Script.java
supertick: It can't be final, and its a bit of a philosophical and experience difference. 
To me POJOs are extremely dumb and simple data contracts, for some reason Java culture thinks setters and getters are always necessary.  They are not. Especially if the understanding of was a POJO is agreed on.  Other languages do not unnecessarily muddle thier POXOs with accessor functions..  POPOs, JavaScript Dictionaries, POCOs etc...
It is an enforced contract in the form of a object in memory.  Additionally, I've run into parsers that have had issues with accessor nuances

PR #1295, opened 2023-06-07
Solrgui

kwatters:
this is a refresh of the solr service and the solr gui to actually be able to search some data ...


src/main/java/org/myrobotlab/service/Solr.java
```diff
@@ -463,6 +467,34 @@ public IplImage bytesToImage(byte[] bytes) throws IOException {
...
+   *           boom
+   * 
+   */
+  public String bytesToBase64Jpg(byte[] bytes) {
+    //
+    // let's assume we're a buffered image .. those are serializable :)
+    ByteArrayOutputStream os = new ByteArrayOutputStream();
+    try {
+      BufferedImage bufImage = ImageIO.read(new ByteArrayInputStream(bytes));
+      ImageIO.write(bufImage, "jpg", os);
```
supertick: I swore I made this in CodecUtils at one point 
It'd be "nice" if it was re-implemented there
```java
public String toBase64Image(byte[] bytes){
         toBase64Image(bytes, "jpg")
}
public String toBase64Image(byte[] bytes, String type)
```
if this was done again in CodecUtils we'd standardize it, and this way anyone can use it for gif tiff png and anything else ImageIO currently supports

src/main/java/org/myrobotlab/service/Solr.java
```diff
@@ -463,6 +467,34 @@ public IplImage bytesToImage(byte[] bytes) throws IOException {
...
+      os.close();
+    } catch (IOException e) {
+      // TODO Auto-generated catch block
+      return "ERROR converting image to jpg base64.";
+    }
+
+    String data = String.format("data:image/%s;base64,%s", "jpg", Base64.getEncoder().encodeToString(os.toByteArray()));
```
supertick: :+1:  would be nice in CodeUtils

src/main/java/org/myrobotlab/service/Solr.java
```diff
@@ -983,52 +1118,67 @@ public void onMessage(Message message) {
...
-        // iterate over the fields on the returned document
-        for (String fieldName : d.getFieldNames()) {
-
-          System.out.print(fieldName + "\t");
-          // fields can be multi-valued
-          for (Object value : d.getFieldValues(fieldName)) {
-            System.out.print(value);
-            System.out.print("\t");
+      boolean testIndex = false;
+      if (testIndex) {
```
supertick, line 1032: should be in unit test ... no ?

src/main/resources/resource/WebGui/app/service/js/SolrGui.js
```diff
@@ -1,25 +1,102 @@
...
+    	// update the start offset and run the search
+    	$scope.startOffset += $scope.pageSize;
+    	if ($scope.startOffset > $scope.numFound) {
+    		$scope.startOffset -= $scope.pageSize;
+    	}
+    	$scope.execSearch();
+    }
+    
+    msg.subscribe('publishResults');
+    msg.subscribe(this);
```
supertick, line 98: :+1:

supertick (approved):

src/main/java/org/myrobotlab/service/Solr.java
kwatters: we have org.myrobotlab.image.Util  .. this is an image specific util.. moved it there.

src/main/java/org/myrobotlab/service/Solr.java
kwatters: there is a SolrTest ....  this is just some example code here.

PR #1300, opened 2023-06-14
to provide search over a directory with media in it ( mp3 , video , etc ) 

kwatters:
uses FileConnector to create documents that are sent to the DocumentPipeline which extracts the metadata/text from the files, and then it sends the results to Solr where the documents can be searched.


src/main/java/org/myrobotlab/service/Solr.java
```diff
@@ -65,6 +69,7 @@
  */
 public class Solr extends Service implements DocumentListener, TextListener, MessageListener {
 
+  private SolrConfig config;
```
AutonomicPerfectionist: I don't think this is correct, Java doesn't support overriding fields so this actually hides the superclass `config` and creates a new field only accessible to methods in this downstream class. Without doing some in depth debugging I can't tell for certain if this will break things or if the getConfig() and applyConfig() methods correctly set this field to the same reference as the superclass field

src/main/java/org/myrobotlab/service/Solr.java
```diff
@@ -75,6 +80,10 @@ public class Solr extends Service implements DocumentListener, TextListener, Mes
   public String solrHome = "Solr";
   // EmbeddedSolrServer embeddedSolrServer = null;
   transient private EmbeddedSolrServer embeddedSolrServer = null;
+  
+  Collection<SolrInputDocument> documentBatch = Collections.synchronizedCollection(new ArrayList<>());
```
AutonomicPerfectionist, line 84: `CopyOnWriteArrayList` would be better here, synchronized collection just wraps the given collection in a big synchronized block which is inefficient compared to a dedicated concurrent implementation

src/main/java/org/myrobotlab/service/Solr.java
```diff
@@ -159,43 +168,51 @@ public void startEmbedded(String path) throws SolrServerException, IOException {
...
-      // TODO : retry?
-      log.warn("An exception occurred when trying to add document to the index.", e);
-    } catch (IOException e) {
-      // TODO : maybe retry?
-      log.warn("A network exception occurred when trying to add document to the index.", e);
-    }
+    // Always batch!
+    ArrayList<SolrInputDocument> docs = new ArrayList<SolrInputDocument>();
+    docs.add(doc);
+    addDocuments(docs);
```
AutonomicPerfectionist: Can just do `addDocuments(List.of(doc))`

supertick (approved):

src/main/java/org/myrobotlab/document/transformer/Configuration.java
```diff
@@ -17,7 +18,11 @@ public class Configuration {
   // workflow /pipeline config
   // connector config
 
-  protected HashMap<String, Object> config = null;
+  /**
+   * 
+   */
+  private static final long serialVersionUID = 1L;
+  public HashMap<String, Object> config = null;
```
AutonomicPerfectionist: Can this be initialized here or is it absolutely necessary to be set elsewhere?

src/main/java/org/myrobotlab/document/transformer/Configuration.java
```diff
@@ -132,6 +137,12 @@ public List<String> getListParam(String name) {
     return null;
   }
 
+  public void setMapProperty(String name, Map map) {
```
AutonomicPerfectionist, line 135: Would Map<String, String> be sufficient or is it expected to also contain other types?

src/main/java/org/myrobotlab/document/workflow/WorkflowWorker.java
```diff
@@ -24,10 +25,12 @@ public class WorkflowWorker extends Thread {
 
   private final LinkedBlockingQueue<Document> queue;
 
-  WorkflowWorker(WorkflowConfiguration workflowConfig, LinkedBlockingQueue<Document> queue, String workerId) throws ClassNotFoundException {
+  private final DocumentPipeline pipeline;
+  WorkflowWorker(WorkflowConfiguration workflowConfig, LinkedBlockingQueue<Document> queue, String workerId, DocumentPipeline pipeline) throws ClassNotFoundException {
```
AutonomicPerfectionist, line 29: Is package level visibility desired?

src/main/java/org/myrobotlab/document/workflow/WorkflowWorker.java
```diff
@@ -63,6 +66,7 @@ public void run() {
           processing = true;
           // process from the start of the workflow
           processDocumentInternal(doc, 0);
+          pipeline.invoke("publishDocument", doc);
```
AutonomicPerfectionist, line 69: If we don't want to maintain a reference to the pipeline and this is the only reason we need it, we could instead pass its name and then use Runtime.invoke() with its name to auto route to the pipeline, I think that works anyway, not at my desktop to check right now

src/main/java/org/myrobotlab/service/Solr.java
```diff
@@ -75,6 +80,10 @@ public class Solr extends Service implements DocumentListener, TextListener, Mes
   public String solrHome = "Solr";
   // EmbeddedSolrServer embeddedSolrServer = null;
   transient private EmbeddedSolrServer embeddedSolrServer = null;
+  
+  Collection<SolrInputDocument> documentBatch = Collections.synchronizedCollection(new ArrayList<>());
+  // The batch size of documents to accumulate before flushing the batch to solr.
```
AutonomicPerfectionist, line 85: Should probably be a javadoc instead of a regular comment

src/main/java/org/myrobotlab/service/Solr.java
```diff
@@ -75,6 +80,10 @@ public class Solr extends Service implements DocumentListener, TextListener, Mes
   public String solrHome = "Solr";
   // EmbeddedSolrServer embeddedSolrServer = null;
   transient private EmbeddedSolrServer embeddedSolrServer = null;
+  
+  Collection<SolrInputDocument> documentBatch = Collections.synchronizedCollection(new ArrayList<>());
+  // The batch size of documents to accumulate before flushing the batch to solr.
+  public transient int batchSize = 100;
```
AutonomicPerfectionist, line 86: Does this have a getter and setter? Might be nice to be able to set this through a message. Side note: might be able to allow setting and getting fields with messages in the future with a bit of tweaking in the method invocation code

src/main/java/org/myrobotlab/service/Solr.java
```diff
@@ -159,43 +172,51 @@ public void startEmbedded(String path) throws SolrServerException, IOException {
...
-      // TODO : retry?
-      log.warn("An exception occurred when trying to add document to the index.", e);
-    } catch (IOException e) {
-      // TODO : maybe retry?
-      log.warn("A network exception occurred when trying to add document to the index.", e);
-    }
+    // Always batch!
+    ArrayList<SolrInputDocument> docs = new ArrayList<SolrInputDocument>();
+    docs.add(doc);
+    addDocuments(docs);
```
AutonomicPerfectionist: Could just do `addDocuments(List.of(doc))`

src/main/java/org/myrobotlab/service/interfaces/DocumentPublisher.java
```diff
@@ -4,10 +4,21 @@
 
 public interface DocumentPublisher {
 
+  public static String[] publishMethods = new String[] { "publishDocument" , "publishFlush"};
```
AutonomicPerfectionist, line 7: Should probably be final as well

src/main/java/org/myrobotlab/service/Solr.java
kwatters: I'd be worried that if the list isn't synchronized, then a document could be lost in the batching logic...  In the scenario where we pick up a batch to send  while another thread is adding to the batch as that happens.  then after we send that batch, we clear this list.. so we don't want it to loose docs that were added in that small period of time.  another implementation could be to use a queue, so we only remove the elements that are in the current batch...

src/main/java/org/myrobotlab/service/Solr.java
supertick: Yeah, that won't work well for serialization ... there is already a Service.config

src/main/java/org/myrobotlab/service/Solr.java
AutonomicPerfectionist: In that case there's still a race condition in between adding the document batch and clearing it, you'd need to synchronize on it for the whole critical section or use a lock, for a synchronized collection the synchronization only applies during operations, so two threads could do the following:

Thread 1: adds the document batch to Solr
Thread 2: attempts to add documents but blocks while waiting for thread 1

Thread 1 completes adding the batch, so thread 2 unblocks and adds the document

Thread 1 calls clear() but blocks until thread 2 finishes

Now, thread 1 unblocks and clears the document batch even when there's a new document that hasn't been sent yet

PR #1301, opened 2023-06-14
implement servo limits

supertick:
limits

### DIFF
--- a/src/main/java/org/myrobotlab/service/Servo.java
+++ b/src/main/java/org/myrobotlab/service/Servo.java
@@ -96,7 +96,7 @@ public class Servo extends AbstractServo implements ServiceLifeCycleListener {
     newPos = (newPos < minLimit) ? minLimit : newPos;
     newPos = (newPos > maxLimit) ? maxLimit : newPos;
     
-    log.error("============== newPos {} ====", newPos);
+    log.debug("{} processMove {}", getName(), newPos);
 
     // This is to allow attaching disabled
     // then delay enabling until the first moveTo command
--- a/src/main/java/org/myrobotlab/service/Servo.java
+++ b/src/main/java/org/myrobotlab/service/Servo.java
@@ -91,8 +91,8 @@ public class Servo extends AbstractServo implements ServiceLifeCycleListener {
       return false;
     }
 
-    double minLimit = (mapper.minX < mapper.maxX) ? mapper.minX : mapper.maxX;
-    double maxLimit = (mapper.maxX > mapper.minX) ? mapper.maxX : mapper.minX;
+    double minLimit = Math.min(mapper.minX, mapper.maxX);
+    double maxLimit = Math.min(mapper.minX, mapper.maxX);
     newPos = (newPos < minLimit) ? minLimit : newPos;
     newPos = (newPos > maxLimit) ? maxLimit : newPos;
 
--- a/src/main/java/org/myrobotlab/service/Servo.java
+++ b/src/main/java/org/myrobotlab/service/Servo.java
@@ -25,7 +25,6 @@
 
 package org.myrobotlab.service;
 
-import java.util.HashSet;
 import java.util.Set;
 
 import org.myrobotlab.logging.Level;
@@ -36,7 +35,6 @@ import org.myrobotlab.sensor.EncoderData;
 import org.myrobotlab.sensor.TimeEncoder;
 import org.myrobotlab.service.abstracts.AbstractServo;
 import org.myrobotlab.service.config.ServiceConfig;
-import org.myrobotlab.service.config.ServiceConfig.Listener;
 import org.myrobotlab.service.config.ServoConfig;
 import org.myrobotlab.service.data.ServoMove;
 import org.myrobotlab.service.interfaces.ServiceLifeCycleListener;
@@ -87,6 +85,16 @@ public class Servo extends AbstractServo implements ServiceLifeCycleListener {
    */
   @Override
   protected boolean processMove(Double newPos, boolean blocking, Long timeoutMs) {
+
+    if (newPos == null) {
+      log.info("servo processMove(null) not valid position");
+      return false;
+    }
+
+    double minLimit = (mapper.minX < mapper.maxX) ? mapper.minX : mapper.maxX;
+    double maxLimit = (mapper.maxX > mapper.minX) ? mapper.maxX : mapper.minX;
+    newPos = (newPos < minLimit) ? minLimit : newPos;
+    newPos = (newPos > maxLimit) ? maxLimit : newPos;
 
     // This is to allow attaching disabled
     // then delay enabling until the first moveTo command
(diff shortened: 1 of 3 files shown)


src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -88,6 +86,16 @@ public Servo(String n, String id) {
   @Override
   protected boolean processMove(Double newPos, boolean blocking, Long timeoutMs) {
 
+    if (newPos == null) {
+      log.info("servo processMove(null) not valid position");
+      return false;
+    }
+
+    double minLimit = (mapper.minX < mapper.maxX) ? mapper.minX : mapper.maxX;
```
AutonomicPerfectionist: Can be shortened to Math.min()

src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -88,6 +86,16 @@ public Servo(String n, String id) {
   @Override
   protected boolean processMove(Double newPos, boolean blocking, Long timeoutMs) {
 
+    if (newPos == null) {
+      log.info("servo processMove(null) not valid position");
+      return false;
+    }
+
+    double minLimit = (mapper.minX < mapper.maxX) ? mapper.minX : mapper.maxX;
+    double maxLimit = (mapper.maxX > mapper.minX) ? mapper.maxX : mapper.minX;
```
AutonomicPerfectionist: Math.max()

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -1073,7 +1071,7 @@ public ServoEvent publishServoStarted(String name, Double position) {
    */
   @Override
   public ServoEvent publishServoStopped(String name, Double position) {
-    log.debug("publishServoStopped({}, {})", name, position);
+    log.error("{} publishServoStopped({}, {})", System.currentTimeMillis(), name, position);
```
AutonomicPerfectionist: Why error instead of debug?

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -1031,17 +1030,16 @@ public void sync(String name) {
...
   }
 
   @Override
   public void waitTargetPos() {
-    //
-    // while (this.pos != this.targetPos) {
-    // Some sleep perhaps?
-    // TODO:
-    // }
+     while (this.getCurrentInputPos() != this.targetPos) {
```
AutonomicPerfectionist, line 1040: How does this work for servos that don't have access to the raw encoder?

AutonomicPerfectionist (changes requested):

src/main/java/org/myrobotlab/service/Servo.java
supertick: :+1:  updated

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
supertick: Thank you ... :roll_eyes:

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
supertick: All servos currently have access to a time encoder ... heh, its the cheapest and least accurate sensor ever .. but it doesn't require additional hardware and provides the estimate of position at any given time.  "Real" hardware can substitute a real encoder instead of a time estimation - that's kinda what @kwatters DiyServo did

src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -88,6 +86,18 @@ public Servo(String n, String id) {
   @Override
   protected boolean processMove(Double newPos, boolean blocking, Long timeoutMs) {
 
+    if (newPos == null) {
```
AutonomicPerfectionist, line 89: Is there any actual code that passes a null position? If not, it would probably be best to just change the parameter to a primitive and get rid of this check entirely

src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -88,6 +86,18 @@ public Servo(String n, String id) {
...
   protected boolean processMove(Double newPos, boolean blocking, Long timeoutMs) {
 
+    if (newPos == null) {
+      log.info("servo processMove(null) not valid position");
+      return false;
+    }
+
+    double minLimit = Math.min(mapper.minX, mapper.maxX);
+    double maxLimit = Math.max(mapper.minX, mapper.maxX);
+    newPos = (newPos < minLimit) ? minLimit : newPos;
```
AutonomicPerfectionist, line 96: Can be replaced with Math.max()

Or, we add a forceToRange() method in some utility class to take care of this and the next line

src/test/java/org/myrobotlab/service/ServoTest.java
```diff
@@ -273,10 +276,43 @@ public void moveToBlockingTest() throws Exception {
...
+
+    // wait for the move to complete using waitTargetPos
+    new Thread(() -> {
+        servo01.waitTargetPos();
+        targetLatch.countDown();
+    }).start();
+
+    // wait for the move to complete
+    targetLatch.await();
+    log.error("finished at {}", System.currentTimeMillis());
```
AutonomicPerfectionist: Doesn't need to be error

src/test/java/org/myrobotlab/service/ServoTest.java
```diff
@@ -273,10 +276,43 @@ public void moveToBlockingTest() throws Exception {
...
     assertTrue("Servo should be enabled", servo01.isEnabled());
     assertFalse("Servo should not be moving now.", servo01.isMoving());
-    // Now let's wait for the idle disable timer to kick off + 1000ms
-    // TODO: figure out why smaller values like 100ms cause this test to fail.
-    // there seems to be some lag
-    Thread.sleep(servo01.getIdleTimeout() + 1000);
+    
+    CountDownLatch moveLatch = new CountDownLatch(1);
+    CountDownLatch targetLatch = new CountDownLatch(1);
+    CountDownLatch disableLatch = new CountDownLatch(1);
```
AutonomicPerfectionist, line 282: I don't see this being counted down anywhere, is it supposed to be?

src/main/java/org/myrobotlab/service/Servo.java
supertick: yeah .. probably InMoov helper methods .. not ready to get all that refactored .. yet

src/test/java/org/myrobotlab/service/ServoTest.java
supertick: fixed

src/test/java/org/myrobotlab/service/ServoTest.java
supertick:
```java

    new Thread(() -> {
        log.info("starting at {}", System.currentTimeMillis());
        servo01.moveTo(0);
        moveLatch.countDown();
    }).start();

    // wait for the move to start
    moveLatch.await();

    // wait for the move to complete using waitTargetPos
    new Thread(() -> {
        servo01.waitTargetPos();
        targetLatch.countDown();
    }).start();
```

supertick: it is

src/main/java/org/myrobotlab/service/Servo.java
supertick: maybe later :thinking:

PR #1302, opened 2023-06-15
framework fixes

supertick:


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1294,6 +1294,14 @@ final public Object invokeOn(boolean blockLocally, Object obj, String methodName
       }
       retobj = method.invoke(obj, params);
       if (blockLocally) {
+        Outbox outbox = null;
+        if (obj instanceof ServiceInterface) {
+          outbox = ((ServiceInterface)obj).getOutbox();
```
AutonomicPerfectionist, line 1299: What does this do?

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1448,8 +1461,23 @@ public ServiceConfig getFilteredConfig() {
   }
 
   @Override
-  public void setConfig(ServiceConfig config) {
+  public ServiceConfig setConfig(ServiceConfig config) {
```
kwatters: normally setters don't return anything.  why change this signature?

src/main/java/org/myrobotlab/framework/interfaces/ServiceInterface.java
```diff
@@ -138,7 +138,17 @@ public interface ServiceInterface extends ServiceQueue, LoggingSink, NameTypePro
    * 
    * @param config
    */
-  void setConfig(ServiceConfig config);
+  ServiceConfig setConfig(ServiceConfig config);
```
kwatters: why return ServiceConfig here?

src/main/java/org/myrobotlab/framework/Service.java
supertick: I'll change it back, was thinking the echo would be useful ..
like fluent style - https://medium.com/miro-engineering/fluent-setter-breaking-the-convention-33ce3433126e

src/main/java/org/myrobotlab/framework/interfaces/ServiceInterface.java
supertick: https://medium.com/miro-engineering/fluent-setter-breaking-the-convention-33ce3433126e

src/main/java/org/myrobotlab/framework/Service.java
supertick: gets the target si's outbox .. invokeOn isn't necessarily the service that is invoking itself, it may be invoking on another service, so its the outbox of the "invoked" service that is required here ...
most of the time it is the same, but not guaranteed .. its a bug

AutonomicPerfectionist: It doesn't seem to be used anywhere though, the outbox declared in this if statement is not accessible outside of it so it's not shadowing `this.outbox` later

src/main/java/org/myrobotlab/framework/Service.java
supertick: set to void

src/main/java/org/myrobotlab/framework/interfaces/ServiceInterface.java
supertick: set to void

src/main/java/org/myrobotlab/framework/Service.java
supertick: broadcast can invokeOn .. when called it invokeOn inside of itself on a different service .. so when it tries to do this and uses its own outbox, you won't get the correct outbox that needs to be blocking ..

AutonomicPerfectionist: I understand that, but this code doesn't seem to do anything with the foreign outbox, it just gets it and then the if statement ends. Since Java has block scoping, the newly assigned outbox variable is not accessible outside of the if statement, so effectively this conditional doesn't do anything

supertick: image

supertick: 1 - beginning of scope
2 - assignment
3 - usage

supertick: if the target is another local service then it needs that services notficationList to process the other subscriptions

AutonomicPerfectionist: Oh, I misread the code, I thought the outer if statement was closed before the usage. My mistake

PR #1304, opened 2023-06-15
Servo-Mixer updates

supertick:


supertick:
davinci

src/main/java/org/myrobotlab/service/ServoMixer.java
```diff
@@ -9,75 +9,188 @@
...
+      if (runningGesture.parts != null) {
+        invoke("publishGestureStarted", gestureName);
+        for (int i = 0; i < runningGesture.parts.size(); ++i) {
+          if (!running) {
+            break;
+          }
+          GesturePart sp = runningGesture.parts.get(i);
+          invoke("publishPlayingGesturePart", sp);
+          invoke("publishPlayingGesturePartIndex", i);
+          switch(sp.type) {
```
AutonomicPerfectionist, line 67: This section would be a great candidate for switch pattern matching in Java 21. Probably should add a comment about it to make refactoring easier when we eventually make our way there

src/main/java/org/myrobotlab/kinematics/Gesture.java
```diff
@@ -0,0 +1,22 @@
...
+
+/** represent a set of servo positions at a given point in time */
+public class Gesture {
+
+  public final static Logger log = LoggerFactory.getLogger(InMoov2.class);
+
+  /**
+   * sequence of poses and offset times
+   */
+  public List<GesturePart> parts = new ArrayList<>();
```
AutonomicPerfectionist: Should either be final or be private with getters/setters

src/main/java/org/myrobotlab/kinematics/GesturePart.java
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * type determines how to handle the value 
+   * depending on what is desired ...
+   */
+  public String type; // Pose | Text | Delay | Message
+  
+  /**
+   * delay type when type is Delay, String when type is Text
+   */
+  public Object value;
```
AutonomicPerfectionist, line 19: This raw Object field along with an implicitly linked String scares me, there are a lot of ways this could be done that embrace the type system instead of eschewing it.

An example: an interface `GesturePartData<T>` with concrete classes `PoseData`, `TextData`, `DelayData`, `MessageData`. `TextData` implements `GesturePart<String>` and `DelayData` sets `T` to whatever the delay type is. `GesturePartData` would have one method: `T getData()`

Just one example, it might make more sense to refactor all of this to eliminate the dispatch-style codepaths and use more polymorphism-based styles

src/main/java/org/myrobotlab/kinematics/Pose.java
```diff
@@ -1,89 +1,11 @@
...
-    } else if (!positions.equals(other.positions))
-      return false;
-    return true;
-  }
-
-  @Override
-  public String toString() {
-    return "Pose [positions=" + positions + ", speeds=\" + speeds + \", name=" + name + "]";
-  }
+  public Map<String, PoseMove> moves = new TreeMap<>();
```
AutonomicPerfectionist: Should be final or private

src/main/java/org/myrobotlab/kinematics/PoseMove.java
```diff
@@ -0,0 +1,23 @@
...
+  
+  public PoseMove(double position, Double speed) {
+    this.position = position;
+    this.speed = speed;
+  }
+
+  /**
+   * frame position
+   */
+  public Double position;
```
AutonomicPerfectionist, line 24: Why the wrapper class instead of primitive?

src/main/java/org/myrobotlab/kinematics/PoseMove.java
```diff
@@ -0,0 +1,23 @@
...
+  
+  /**
+   * frame speed
+   */
+  public Double speed;
```
AutonomicPerfectionist, line 29: Why wrapper?
Also, these fields should be moved above the constructors

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
```diff
@@ -154,17 +156,15 @@ public class JMonkeyEngine extends Service implements Gateway, ActionListener, S
   long deltaMs;
 
   transient DisplayMode displayMode = null;
-
-  transient FlyByCamera flyCam;
+  
+  public  ChaseCamera chaseCamera;
```
AutonomicPerfectionist: Should it be transient? Private?

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
```diff
@@ -226,11 +221,25 @@ public class JMonkeyEngine extends Service implements Gateway, ActionListener, S
...
 
   int width = 1024;
+  
+  float orbitRadius = 10f;
+  
+  float orbitSpeed = 0.5f;
+  
+  float mouseX = 0f;
+  
+  float mouseY = 0f;
```
AutonomicPerfectionist: All of these fields should have a visibility modifier, package private is weird

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
```diff
@@ -984,6 +995,23 @@ public Node getRootNode() {
   public Spatial getSelected() {
     return selectedForView;
   }
+  
+  public String setSelectedPath(String path) {
```
AutonomicPerfectionist, line 1019: Javadoc please since this does more than just simply setting the field

src/main/java/org/myrobotlab/service/ServoMixer.java
```diff
@@ -9,75 +9,188 @@
...
+ * in a gesture file to play back later.
+ * 
+ * @author GroG, kwatters, others...
+ * 
  */
-public class ServoMixer extends Service {
+public class ServoMixer extends Service implements ServiceLifeCycleListener {
 
-  public final static Logger log = LoggerFactory.getLogger(InMoov2.class);
+  public class Player implements Runnable {
```
AutonomicPerfectionist, line 50: Javadoc please

src/main/java/org/myrobotlab/service/ServoMixer.java
```diff
@@ -9,75 +9,188 @@
...
+    String gestureName = null;
+    int poseIndex = 0;
+    boolean running = false;
+    Gesture runningGesture = null;
+    Thread thread = null;
```
AutonomicPerfectionist: All of these fields should have visibility modifiers

src/main/java/org/myrobotlab/kinematics/Pose.java
```diff
@@ -1,89 +1,11 @@
...
-    if (positions == null) {
-      if (other.positions != null)
-        return false;
-    } else if (!positions.equals(other.positions))
-      return false;
-    return true;
-  }
-
-  @Override
-  public String toString() {
```
AutonomicPerfectionist, line 85: Should probably still override `toString()`, `equals()`, and `hashCode()`

src/main/java/org/myrobotlab/kinematics/PoseMove.java
```diff
@@ -0,0 +1,23 @@
+package org.myrobotlab.kinematics;
+
+public class PoseMove {
```
AutonomicPerfectionist, line 11: Needs javadoc

src/main/java/org/myrobotlab/service/ServoMixer.java
```diff
@@ -9,75 +9,188 @@
...
+        error(e);
+      }
+      running = false;
+    }
+
+    public void start(String name, Gesture seq) {
+      gestureName = name;
+      runningGesture = seq;
+      poseIndex++;
+      thread = new Thread(this, String.format("%s-player-%d", getName(), poseIndex));
```
AutonomicPerfectionist: Should probably use an executor pool of some kind, creating a thread for every gesture is expensive. Also, I take it this is not just a subclass of a thread so that it can run multiple gestures simultaneously? If so, there needs to be thread locals or concurrent maps for `running` and other shared state

src/main/java/org/myrobotlab/service/ServoMixer.java
```diff
@@ -175,305 +256,345 @@ public Pose loadPose(String name) {
...
+    }
+    return files;
+  }
+
+  public String getPosesDirectory() {
+    ServoMixerConfig c = (ServoMixerConfig)config;
+    return c.posesDir;
+  }
+
+  public List<ServoControl> listAllServos() {
```
AutonomicPerfectionist, line 319: Shouldn't be needed, equivalent to `Runtime.getServicesFromInterface(ServoControl.class)`. Speaking of which, that method is terribly unoptimized and could use some cleaning up

src/main/java/org/myrobotlab/service/ServoMixer.java
```diff
@@ -175,305 +256,345 @@ public Pose loadPose(String name) {
...
         return;
       }
-      String filename = new File(servoMixerDirectory).getAbsolutePath() + File.separator + name + ".seq";
-      File del = new File(filename);
-      if (del.exists()) {
-        del.delete();
+      
+      for (String sc : p.moves.keySet()) {
+        PoseMove pm = p.moves.get(sc);
+        ServoControl servo = (ServoControl) Runtime.getService(sc);
```
AutonomicPerfectionist, line 347: Potential ClassCastException here if `sc` was created as a different service. Mainly a reminder for me to add a new overload to getService() that takes the expected service type

src/main/java/org/myrobotlab/service/ServoMixer.java
```diff
@@ -175,305 +256,345 @@ public Pose loadPose(String name) {
...
-      } catch (Exception e) {
-      }
-      running = false;
-    }
+  /**
+   * handles incoming selected events and publishes
+   * search events
+   * @param selected
+   */
+  public void onSelected(String selected) {
```
AutonomicPerfectionist, line 389: Not really sure what this does, something to do with the GUI?

src/main/java/org/myrobotlab/service/ServoMixer.java
```diff
@@ -175,305 +256,345 @@ public Pose loadPose(String name) {
...
+    invoke("getPoseFiles");
+  }
+  
+  /**
+   * publishing a search to allow UI searching based on other
+   * service selections
+   * @param servos
+   * @return
+   */
+  public String search(String servos) {
```
AutonomicPerfectionist, line 604: Not sure why this is needed, it only ever seems to be invoked from `onSelected` so couldn't the GUI just use whatever comes from `publishSelected()`?

AutonomicPerfectionist (changes requested):

src/main/java/org/myrobotlab/service/ServoMixer.java
supertick: not sure I know what feature your talking about .. :thinking:

AutonomicPerfectionist: A simple primer article on pattern matching and switch:

https://www.baeldung.com/java-switch-pattern-matching

Essentially, it would let you get rid of the `sp.type` entirely and switch on the type of the value instead. so:

```java
switch(sp.value) {
    case Integer delay -> sleep(delay);
    case String speech -> invoke("publishText", speech);
    // Etc
}
```

It also enforces exhaustive checking for sealed types (a different feature available in Java 17+ that lets you restrict the subclasses/implementors of an interface or class) so you can guarantee you don't miss a case

src/main/java/org/myrobotlab/kinematics/PoseMove.java
supertick: A PoseMove is a combination of position and speed ...
Potentially, if one is not supplied its not changed.
So position can be moved to using a pre-existing speed
Or speed can be changed without having to update position

src/main/java/org/myrobotlab/kinematics/PoseMove.java
supertick: A PoseMove is a combination of position and speed ...
Potentially, if one is not supplied its not changed.
So position can be moved to using a pre-existing speed
Or speed can be changed without having to update position

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
supertick: heh, "weird" ... I agree ;)

src/main/java/org/myrobotlab/kinematics/Gesture.java
supertick: blech .. one of the things i don't like about java .. 
To me it seems to violate the concept of POJOs to mandate unnecessary encapsulation 
If the encapsulation is necessary to provide "special" access, then its not a POJO
Fine ... making setters and getters and hopefully serialization doesn't explode :roll_eyes:

AutonomicPerfectionist: Thankfully later versions have a dedicated Record type that does pretty much what we want out of POJOs/data classes, including automatic getter generation. They're basically replacing the legacy POJO style so we don't have to worry about all the ceremony

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
supertick: updating all to "protected" ... this is imo should have been the undecorated version .. instead weird package level access is .. :shrug:

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
supertick: updated

src/main/java/org/myrobotlab/service/ServoMixer.java
supertick: updated

src/main/java/org/myrobotlab/service/ServoMixer.java
supertick: updated to protected

src/main/java/org/myrobotlab/kinematics/PoseMove.java
supertick: updated

src/main/java/org/myrobotlab/service/ServoMixer.java
supertick: interesting

src/main/java/org/myrobotlab/kinematics/Gesture.java
supertick: updated

src/main/java/org/myrobotlab/kinematics/Pose.java
supertick: updated

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
supertick: updated to protected

src/main/java/org/myrobotlab/kinematics/Pose.java
supertick: updated

src/main/java/org/myrobotlab/service/ServoMixer.java
supertick: yup .. updated

supertick: no .. not multiple gestures simultaneously .. that's why lots of the old python gestures scripts have finishGesture() calls at the end .... in that they want to lock & unlock a gesture ...
It just needs a single thread ... I find defining it as Runnable allows more flexibility, including moving it to an executor pool .. which i've done (should at least cut down on the startup time...)

src/main/java/org/myrobotlab/service/ServoMixer.java
supertick: it should be an interface ServoControl not service type.. at least that would give someone the flexibility to use DIYServo or potentially something besides a servo which implements the same interface.
For the most access patterns Runtime.getService "should" be return interfaces only

src/main/java/org/myrobotlab/service/ServoMixer.java
supertick: sort of .. it doesn't currently set any state on the service, but it is an important publishing receiver for servo mixer
I created a SelectListener interface to more firmly tie things up ..

supertick: selectlistener

src/main/java/org/myrobotlab/service/ServoMixer.java
supertick: publishSelectedPath comes from JMonkeyEngine
so JMonkeyEngine -> ServoMixer -> ServoMixerGui

src/main/java/org/myrobotlab/service/ServoMixer.java
AutonomicPerfectionist: Sorry, my comment wasn't clear enough. I meant if, for example, the pose moves a servo called "test". But there's a service called "test" that is instead a ProgramAB service for example. Usually shouldn't happen, but if it does we should probably fail gracefully instead of throwing an arcane exception

src/main/java/org/myrobotlab/kinematics/GesturePart.java
supertick: yup .. this makes sense, but I don't want to do it until after its been used in the wild and some feedback is provide regarding all the possible different "things" are needed to put into a gesture ...

src/main/java/org/myrobotlab/service/ServoMixer.java
supertick: Service.interfaceSet  is the way to go !   The webguis use this, but java-land could be using this more too...
We should:
* change it to a TreeSet not Map
* have it done only "once" then services of the same type just need a reference to it

AutonomicPerfectionist (approved):

PR #1313, opened 2023-07-09
ProgramAB config and predicates

supertick:
ProgramAB now publishes switching topics
also has the capability of accepting another services config into predicates, for more direct access from aiml

### DIFF
--- a/src/main/java/org/myrobotlab/programab/Session.java
+++ b/src/main/java/org/myrobotlab/programab/Session.java
@@ -14,6 +14,7 @@ import org.alicebot.ab.Predicates;
 import org.myrobotlab.io.FileIO;
 import org.myrobotlab.logging.LoggerFactory;
 import org.myrobotlab.service.ProgramAB;
+import org.myrobotlab.service.config.ProgramABConfig;
 import org.slf4j.Logger;
 
 /**
@@ -82,6 +83,11 @@ public class Session {
         predicatesFile = userPredicates;
         chat.predicates.getPredicateDefaults(userPredicates.getAbsolutePath());
       }
+      
+      ProgramABConfig config = (ProgramABConfig)programab.getConfig();
+      if (config.startTopic != null){
+        chat.predicates.put("topic", config.startTopic);
+      }
     }
     predicates = chat.predicates;
     return chat;
@@ -93,10 +99,9 @@ public class Session {
     sort.addAll(getChat().predicates.keySet());
     for (String predicate : sort) {
       String value = getChat().predicates.get(predicate);
-      if (predicate.equals("test")) {
-        log.info("here");
+      if (!predicate.startsWith("cfg_")) {
+        sb.append(predicate + ":" + value + "\n");
       }
-      sb.append(predicate + ":" + value + "\n");
     }
     File predicates = new File(FileIO.gluePaths(botInfo.path.getAbsolutePath(), String.format("config/%s.predicates.txt", userName)));
     predicates.getParentFile().mkdirs();
--- a/src/main/java/org/myrobotlab/service/ProgramAB.java
+++ b/src/main/java/org/myrobotlab/service/ProgramAB.java
@@ -46,6 +46,7 @@ import org.myrobotlab.service.interfaces.TextPublisher;
 import org.myrobotlab.service.interfaces.UtteranceListener;
 import org.myrobotlab.service.interfaces.UtterancePublisher;
 import org.slf4j.Logger;
+import org.yaml.snakeyaml.Yaml;
 
 /**
  * Program AB service for MyRobotLab Uses AIML 2.0 to create a ChatBot This is a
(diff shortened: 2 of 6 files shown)


src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1427,8 +1339,8 @@ public void onUtterance(Utterance utterance) throws Exception {
         // TODO: don't talk to bots.. it won't go well..
         // TODO: the discord api can provide use the list of mentioned users.
         // for now.. we'll just see if we see Mr. Turing as a substring.
-        sleep = (sleep || utterance.text.contains("@")) && !utterance.text.contains(botName);
-        if (!sleep) {
+        c.sleep = (c.sleep || utterance.text.contains("@")) && !utterance.text.contains(botName);
```
AutonomicPerfectionist, line 1346: If I understand this correctly, I think it would cause the bot to wake up if its name is mentioned anywhere in the utterance. Could be problematic if people are talking about the bot but not directly talking to it.

Also, just searching for the @ symbol makes it impossible to include it in a message to the bot, so things like email addresses would cause it to go to sleep.

Could there be a `sleepListener` interface instead, that Discord could use to set the sleep state based on its own conditions? I think this would be useful for other services too, like Inmoov so other services can set the sleep state and listen for the sleep state changes

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1148,81 +1079,53 @@ public void saveAimlFile(String botName, String filename, String data) {
...
+      c.botDir = getResourceDir();
     }
 
-    List<File> botsFromScanning = scanForBots(config.botDir);
+    List<File> botsFromScanning = scanForBots(c.botDir);
     for (File file : botsFromScanning) {
       addBotPath(file.getAbsolutePath());
     }
 
-    if (config.currentBotName != null) {
```
AutonomicPerfectionist, line 1198: Why's this removed? Is the bot name deprecated?

supertick: good catch - fixed ... I thought initially it wouldn't make a difference, now that config is normalized, but the set functions broadcast state too .. so yeah, it should do it as it did before

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: Yeah, sounds like a good idea.  "Utterance" vs plain text is supposed to keep some of the necessary meta data in order to do appropriate processing .. vs just a text in a string.  I agree it seems very Discord centric and the loss of talking about email.  This pr is only about normalizing the member variables with config, but a SleepListener sounds like a good idea for a future one :+1:

src/main/java/org/myrobotlab/programab/Session.java
```diff
@@ -82,6 +83,11 @@ private synchronized Chat getChat() {
         predicatesFile = userPredicates;
         chat.predicates.getPredicateDefaults(userPredicates.getAbsolutePath());
       }
+      
+      ProgramABConfig config = (ProgramABConfig)programab.getConfig();
+      if (config.startTopic != null){
```
kwatters, line 88: I don't think this is the right thing to do here.  When a session is first created with a new user..  someone that hasn't been seen before.. it makes sense to set the start topic..  (which should be unknown by default) ...  
Think of the scenario where the robot is talking with multiple people...   you don't want to reset the topic each time the user switches between sessions.

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1462,17 +1378,51 @@ public void onUtterance(Utterance utterance) throws Exception {
       }
     }
   }
+  
+  /**
+   * This receiver can take a config published by another service and sync
+   * predicates from it
+   * @param cfg
+   */
+  public void onConfig(ServiceConfig cfg) {
```
kwatters, line 1387: What's the use case for this?    In general, I like the idea that a service could listen for a config, but I don't understand your specific motivation here for this one...

src/main/java/org/myrobotlab/programab/Session.java
supertick: I'm not sure I agree.  Yes multi session is a complexity to begin addressing in a meaningful way, but right now there is just "unknown" as far as the consistency of starting a topic.  I think there is value in being able to consistently start a new session with a known topic that isn't "unknown" depending the the user's needs

src/main/java/org/myrobotlab/service/ProgramAB.java
supertick: Trying to tame the chaos that was made by gluing Jython with ProgramAB and a bajillion global variables..
at least in this context its a step towards structured data, by providing easier access to config in aiml

PR #1314, opened 2023-07-12
servo improvements

supertick:


src/main/java/org/myrobotlab/service/Servo.java
```diff
@@ -448,4 +352,8 @@ public void onStopped(String name) {
   public void onReleased(String name) {
   }
 
+  public boolean isAttached() {
```
AutonomicPerfectionist: Should be in the AbstractServo superclass

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -267,12 +268,13 @@ public void onStarted(String name) {
...
-    } else if (EncoderControl.class.isAssignableFrom(service.getClass())) {
+      attachServoController(service.getName());
+    } 
+    if (EncoderControl.class.isAssignableFrom(service.getClass())) {
+      
       attach((EncoderControl) service);
-    } else {
-      warn(String.format("%s.attach does not know how to attach to a %s", this.getClass().getSimpleName(), service.getClass().getSimpleName()));
     }
+      warn(String.format("%s.attach does not know how to attach to a %s", this.getClass().getSimpleName(), service.getClass().getSimpleName()));
```
AutonomicPerfectionist, line 277: Neither conditionals return, so I think this will warn for every attach() call

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
```diff
@@ -339,68 +323,60 @@ public AngleData publishJointAngle(AngleData angle) {
...
+
+    if (getPin() == null) {
+      error("cannot attach servo if pin is null");
+      return;
     }
-    // update speed if non-null value supplied
-    if (speed != null) {
-      setSpeed(speed);
+
+    if (isAttached && !service.equals(service)) {
```
AutonomicPerfectionist: What's this supposed to be doing? It seems to be checking the service name against itself

src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
```diff
@@ -401,7 +401,7 @@ public interface ServoControl extends AbsolutePositionControl, EncoderListener,
   void writeMicroseconds(int uS);
 
   // for instance attachment
-  void attachServoController(String sc, Integer pin, Double pos, Double speed);
+  void attachServoController(String sc);
```
AutonomicPerfectionist, line 404: Should probably have a javadoc even if it's pretty self-explanatory

src/main/java/org/myrobotlab/service/Servo.java
supertick: agreed, done

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
supertick: yup, fixed

src/main/java/org/myrobotlab/service/abstracts/AbstractServo.java
supertick: good catch :roll_eyes: 
it should be 
```java
    if (isAttached && !service.equals(c.controller)) {
```
updated

src/main/java/org/myrobotlab/service/interfaces/ServoControl.java
supertick: updated ...

PR #1316, opened 2023-07-17
cron update

supertick:


README.md
```diff
@@ -121,4 +121,21 @@ resources:
...
+id: null
+listeners:
+locale: null
+logLevel: info
+peers: null
+registry:
+- runtime
+- security
+- webgui
+- raspi
```
AutonomicPerfectionist: Probably don't need raspi in this example config

src/main/java/org/myrobotlab/arduino/BoardInfo.java
```diff
@@ -16,17 +16,38 @@ public class BoardInfo implements Serializable {
...
+  /**
+   * Number of microseconds arduino uses to pass through a
+   * control loop in MrlComm - very Arduino/MrlComm specific
+   * FIXME - make generalized BoardType that can report useful information
+   * from any type of board with pins
+   */
+  public Integer microsPerLoop;
+  
+  /**
+   *
```
AutonomicPerfectionist, line 38: Probably should add the javadoc here, and for the other fields without one

src/main/java/org/myrobotlab/service/Cron.java
```diff
@@ -1,44 +1,73 @@
...
 import org.myrobotlab.logging.LoggingFactory;
+import org.myrobotlab.service.config.CronConfig;
 import org.slf4j.Logger;
 
 import it.sauronsoftware.cron4j.Scheduler;
 
 /**
- * Cron - This is a cron based service that can execute a "task" at some point
- * in the future such as "invoke this method on that service"
+ * Cron - This is a cron based service that can execute a "task".
```
AutonomicPerfectionist, line 18: Probably should specify it doesn't require a Unix like operating system or even for the real `cron` program to be installed. Also maybe say what Cron actually is for those unfamiliar with it

src/main/java/org/myrobotlab/service/Cron.java
```diff
@@ -1,44 +1,73 @@
...
+     * data parameters to invoke
+     */
     public Object[] data;
 
-    public Task(Cron myService, String cronPattern, String name, String method) {
-      this(myService, cronPattern, name, method, (Object[]) null);
+    /**
+     * unique hash the scheduler uses (only)
+     */
+    transient public String hash;
```
AutonomicPerfectionist, line 44: Do these fields need to be public, or can they be private/protected?

src/main/java/org/myrobotlab/service/RasPi.java
```diff
@@ -81,48 +72,94 @@ public String toString() {
...
+    bcmToWiring.put("GPIO 22", "GPIO 3");
+    bcmToWiring.put("GPIO 23", "GPIO 4");
+    bcmToWiring.put("GPIO 24", "GPIO 5");
+    bcmToWiring.put("GPIO 25", "GPIO 6");
+    bcmToWiring.put("GPIO 26", "GPIO 24");
+    bcmToWiring.put("GPIO 27", "GPIO 2");
+
+    for (String pin : bcmToWiring.keySet()) {
+      String wiring = bcmToWiring.get(pin);
+      wiringToBcm.put(wiring, pin);
```
AutonomicPerfectionist, line 110: If I understand this correctly, it's just making an inverse of the first map. There's a special data structure just for this purpose called a BiMap, which is a map that you can get the inverse of. I used one for my Android app, it might be worth looking to see any of our existing dependencies have an implementation already

src/main/java/org/myrobotlab/service/RasPi.java
```diff
@@ -376,6 +541,110 @@ public PinData publishPin(PinData pinData) {
...
+  public void reset() {
+    // TODO Auto-generated method stub
+    // reset pins/i2c devices/gpio pins
+  }
+
+  public void scan() {
+    scan(null);
+  }
+
+  public void scan(Integer busNumber) {
```
AutonomicPerfectionist, line 601: Probably needs javadoc. And is wrapper type needed or can primitive be used?

src/main/java/org/myrobotlab/service/RasPi.java
```diff
@@ -376,6 +541,110 @@ public PinData publishPin(PinData pinData) {
...
+
+      I2CBus bus = I2CFactory.getInstance(busNumber);
+
+      if (!validI2CAddresses.containsKey(busNumber)) {
+        validI2CAddresses.put(busNumber, new HashSet<>());
+      }
+
+      Set<String> addresses = validI2CAddresses.get(busNumber);
+
+      for (int i = 1; i < 128; i++) {
```
AutonomicPerfectionist, line 617: Probably should put a comment explaining why i starts at 1

src/main/java/org/myrobotlab/service/RasPi.java
```diff
@@ -376,6 +541,110 @@ public PinData publishPin(PinData pinData) {
...
+        String hex = Integer.toHexString(i);
+        try {
+          I2CDevice device = bus.getDevice(i);
+          device.read();
+          if (!addresses.contains(hex)) {
+            addresses.add(hex);
+            info("found new i2c device %s", hex);
+          }
+        } catch (Exception ignore) {
```
AutonomicPerfectionist, line 626: Can the set of caught exceptions be narrowed a bit? Right now it would end up ignoring critical exceptions like NPEs

README.md
supertick: good catch .. changing it to python

src/main/java/org/myrobotlab/service/Cron.java
supertick: updated

src/main/java/org/myrobotlab/service/Cron.java
supertick: its overly smart .. it should be defined here, it should probably just be part of CronConfig...  "if" that was the case, it would be public.  So next refactor I'll make it part of config directly.

src/main/java/org/myrobotlab/arduino/BoardInfo.java
supertick: I kind of lost interest in updating the javadocs, because an Arduino agnostic  BoardType definition is **really** needed.

src/main/java/org/myrobotlab/service/RasPi.java
supertick: right, I don't want to roll my  own :) ... and apache commons where one version is defined isn't part of RuntimeMeta ... yet

src/main/java/org/myrobotlab/service/RasPi.java
supertick: I debated if the parameter should be taken out completely.  @Cyber-One informed me that raspi only exposes bus 1 for i2c .. so there is no choice

src/main/java/org/myrobotlab/service/RasPi.java
supertick: some info says its reserved

Reserved Addresses:
0x00 (general call address)
0x01 to 0x07 (reserved)
0x78 to 0x7F (10-bit addressing, not usable in 7-bit addressing)

image

some tools scan at 03 ... :shrug: 

it can also depend on if its 7-bit or 10-bit addressing (raspi is 7 I believe) 

I'd document it if I was confident I understood it.

src/main/java/org/myrobotlab/service/RasPi.java
AutonomicPerfectionist: That reminds me, actually, I *think* we do have a soft dependency on Apache Commons lang 3 somewhere. I get a class not found exception when running through my IDE and not installing any ~~exceptions~~ services.

src/main/java/org/myrobotlab/service/RasPi.java
supertick: heh, totally fair 
... but I don't trust the i2c read to only throw an IOException.  If I separate the possibilities, I'd probably want the processing to be the same, and this would mean I have to copy/paste duplicate code.

I guess I'm just dealing with ptsd from seeing 
```java
} catch(X a) {
   ... do something ...
} catch(Y b) {
    ... do exactly the same something
} catch(Z c) {
    ... do exactly the same something
```
then someone has a fix for error handling behavior, but doesn't update all the other places....

AutonomicPerfectionist: Java allows unions in catch() statements for exactly that reason:

```java
} catch (X | Y | Z e) {
   // Do thing for all exceptions in only one place
}
```

PR #1317, opened 2023-07-22
Improved Cron and Cron history

supertick:


src/main/java/org/myrobotlab/service/Hd44780.java
```diff
@@ -198,14 +201,39 @@ public void display(String string, int line) {
         setDdramAddress((byte) 0x14);
         break;
       case 3:
-        setDdramAddress((byte) 0x3C);
+        setDdramAddress((byte) 0x54);
```
AutonomicPerfectionist, line 204: Case 1 still needs updated. Should probably put comments in explaining why the addresses are split and why case 1 and case 3 are in a different range than the other two, so we don't get confused later and undo it all

src/main/java/org/myrobotlab/service/Hd44780.java
```diff
@@ -468,13 +497,14 @@ public void clearDisplay() {
    * @param address
    */
   public void setDdramAddress(int address) {
-    if (address < 80) { // Make sure the address is in a valid range
-      lcdWriteCmd((byte) (address | 0b10000000));
-    } else {
-      error("%d Outside allowed DDRAM Address range 0 - 79", address);
+    if (address < 0 || (address > 40 && address < 63) || address > 108) { 
+      error("%d Outside allowed DDRAM Address range 0 - 108", address);
```
AutonomicPerfectionist, line 501: Message should be updated to say the two allowed ranges, since putting in 41 will log this error even though the message says it should be fine

src/main/java/org/myrobotlab/service/Hd44780.java
```diff
@@ -553,29 +583,37 @@ public boolean getVerifyBusyFlag() {
   }
 
   /**
-   * This method will first make sure the HD44780 is in a known state
-   * and that we are syncrnised to the state.
-   * It does this by setting the module into 8 bit mode
-   * then setting it back to 4 bit mode.
+   * This method will first make sure the HD44780 is in a known state and that
+   * we are syncrnised to the state. It does this by setting the module into 8
```
AutonomicPerfectionist, line 587: Synchronized is misspelled 🙂

src/main/java/org/myrobotlab/service/Cron.java
```diff
@@ -81,29 +85,55 @@ public Task(Cron cron, String id, String cronPattern, String name, String method
...
+  }
 
   public final static Logger log = LoggerFactory.getLogger(Cron.class);
 
   private static final long serialVersionUID = 1L;
 
+  /**
+   * history buffer of tasks that have been executed
+   */
+  final protected List<TaskHistory> history = new ArrayList<>();
```
AutonomicPerfectionist, line 120: Array list is a poor implementation for a FIFO-like queue usage, maybe a linked list would be better. The reason is, removing the head of an array list requires shifting all of the values to the left, an O(n) operation, while removing the head of a linked list involves changing the head pointer and unlinking the head from the second element, an O(1) time operation.

The field type could also be changed to Queue instead of List to better show how it's being used, but we lose precise indexing and can only traverse the collection through iterators or popping the head. Not sure indexing is needed here anyway, but it is a trade-off

src/main/java/org/myrobotlab/service/Cron.java
```diff
@@ -81,29 +85,55 @@ public Task(Cron cron, String id, String cronPattern, String name, String method
...
+
+  /**
+   * max size of history buffer
+   */
+  final int HISTORY_SIZE = 30;
```
AutonomicPerfectionist, line 125: Might be nice to be able to change this on the user side, if I have a lot of tasks or some fire more frequently than others I might miss the ones I'm more interested in

src/main/java/org/myrobotlab/service/Cron.java
```diff
@@ -81,29 +85,55 @@ public Task(Cron cron, String id, String cronPattern, String name, String method
...
+  
   /**
    * the thing that translates all the cron pattern values and implements actual tasks
    */
   transient private Scheduler scheduler = new Scheduler();
 
+  /**
+   * map of tasks organized by id
+   */
+  public Map<String, Task> tasks = new LinkedHashMap<>();
```
AutonomicPerfectionist, line 135: Should probably be final, maybe not public

src/main/java/org/myrobotlab/service/Hd44780.java
supertick: updated

src/main/java/org/myrobotlab/service/Hd44780.java
supertick: updated

src/main/java/org/myrobotlab/service/Hd44780.java
supertick: french ;)

src/main/java/org/myrobotlab/service/Cron.java
supertick: Yeah, can also get into nasty multi thread bugs with arraylist .. but the problem i was trying to solve is normalized the id.  
I made sure the only area where add & remove are in the same function block ... in theory i should synchronize it or better yet use a queue ... anyway future refactoring ...

src/main/java/org/myrobotlab/service/Cron.java
supertick: yeah thought about it ... i'll wait till someone asks ;)

src/main/java/org/myrobotlab/service/Cron.java
supertick: oop :see_no_evil:

PR #1319, opened 2023-07-25
Network update

supertick:


doc/README.md
```diff
@@ -0,0 +1,83 @@
...
+    runtime@p1->>webgui@p2: onDescribe 
+    Note left of webgui@p2: Based on the results of the describe,<br/> more querying and subscriptions can be processed
+
+
+    %% opt Server Add Listener, Describe and Reserve
+    %% end
+
+
+```
+### 1 Connection
```
AutonomicPerfectionist, line 35: Should explain what the query string does, I thought the `id` parameter was important for connection but a quick look through `WebGui.handle()` doesn't show any usage of it

doc/README.md
```diff
@@ -0,0 +1,83 @@
...
+  "status": null,
+  "encoding": "json",
+  "method": "addListener",
+  "data": [
+    "{\"topicMethod\":\"describe\",\"callbackName\":\"runtime@caring-hector\",\"callbackMethod\":\"onDescribe\",\"class\":\"org.myrobotlab.framework.MRLListener\"}"
+  ],
+  "class": "org.myrobotlab.framework.Message"
+}
+```
+### 3 Send Describe
```
AutonomicPerfectionist, line 61: Should explain what the describe message is actually for. Notably, what is "fill-uuid" and why is it important? Should specify the stage at which a Runtime instance is `register()`ed, which I believe is in the receiving end of this message

doc/README.md
```diff
@@ -0,0 +1,83 @@
...
+  ],
+  "class": "org.myrobotlab.framework.Message"
+}
+
+```
+
+### 4 Process onDescribe Response
```
AutonomicPerfectionist, line 82: Should make sure to specify that services in the `DescribeResults` parameter are registered, and a new subscription is made to the remote runtime's `registered()` methd

doc/README.md
supertick: Valid point, initially I thought its how one process could identify itself, but registration does that anyway.  Additionally, I haven't worked out the details between where authentication and authorization should be.  Currently, its part of the Security service, but not really utilized yet.

doc/README.md
supertick: updated, fill-uuid should be refactored out.  There should be more simple set of criteria and responses, starting wtih just a list of local services and their names & "types".. although "Type" information is of limited use.  To be polyglot, it really should be the ability to get itnterface information which parameters in the form of openapi schema

PR #1320, opened 2023-07-25
gateway fix

supertick:
image

### DIFF
--- a/src/main/java/org/myrobotlab/net/WsClient.java
+++ b/src/main/java/org/myrobotlab/net/WsClient.java
@@ -139,6 +139,11 @@ public class WsClient extends WebSocketListener {
       log.debug(String.format("MESSAGE: %s", text));
     }
     if (handler != null) {
+      if ("X".equals(text)) {
+        // ignore Atmosphere does a weird sending of X characters I assume
+        // to make sure the connection is unbroken
+        return;
+      } 
       handler.onRemoteMessage(uuid, text);
     }
   }
--- a/src/main/java/org/myrobotlab/service/interfaces/Gateway.java
+++ b/src/main/java/org/myrobotlab/service/interfaces/Gateway.java
@@ -28,7 +28,6 @@ package org.myrobotlab.service.interfaces;
 import java.util.List;
 import java.util.Map;
 
-import org.myrobotlab.codec.CodecUtils;
 import org.myrobotlab.framework.DescribeQuery;
 import org.myrobotlab.framework.Message;
 import org.myrobotlab.framework.Platform;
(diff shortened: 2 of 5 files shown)


src/main/java/org/myrobotlab/net/WsClient.java
```diff
@@ -139,6 +139,11 @@ public void onMessage(WebSocket webSocket, String text) {
       log.debug(String.format("MESSAGE: %s", text));
     }
     if (handler != null) {
+      if ("X".equals(text)) {
+        // ignore Atmosphere does a weird sending of X characters I assume
```
AutonomicPerfectionist, line 143: It's actually MRL doing that, there's a heartbeat timer somewhere in the WebGui service from what I remember

src/main/java/org/myrobotlab/service/interfaces/Gateway.java
```diff
@@ -82,8 +83,8 @@ default Message getDescribeMsg(String connId) {
             "runtime",
             "describe",
             new Object[] {
-                    FILL_UUID_MAGIC_VAL,
-                    new DescribeQuery(Platform.getLocalInstance().getId(), connId)
+                    CodecUtils.toJson(FILL_UUID_MAGIC_VAL),
+                    CodecUtils.toJson(new DescribeQuery(Platform.getLocalInstance().getId(), connId))
```
AutonomicPerfectionist: This is going to result in triple encoding of the message again, we removed that because it caused some bug awhile ago

src/main/resources/resource/WebGui/app/mrl.js
```diff
@@ -346,7 +346,7 @@ angular.module('mrlapp.mrl', []).provider('mrl', [function() {
      */
     this.describe = function(request) {
         console.log('--> got describe: and set jsRuntimeMethodCallbackMap')
-        hello = request.data[1]
+        hello = JSON.parse(request.data[1])
```
AutonomicPerfectionist: What's this for? The triple encoding?

AutonomicPerfectionist (changes requested):

src/main/java/org/myrobotlab/net/WsClient.java
supertick: Its not MRL its nettosphere/atmosphere

src/main/java/org/myrobotlab/service/interfaces/Gateway.java
supertick: I guess the flow needs to be followed, but the current error is easy to reproduce
Connect a runtime with another runtime + webgui

src/main/resources/resource/WebGui/app/mrl.js
supertick: parse decodes

src/main/java/org/myrobotlab/service/interfaces/Gateway.java
supertick: fixed correctly, thanks !

src/main/resources/resource/WebGui/app/mrl.js
supertick: removed .. thanks !

PR #1325, opened 2023-08-01
clean up tika slf4j logging imports in document pipeline

kwatters:
regenerating the pom and excluding logging dependencies from tika imports.

### DIFF
--- a/src/main/java/org/myrobotlab/service/meta/Py4jMeta.java
+++ b/src/main/java/org/myrobotlab/service/meta/Py4jMeta.java
@@ -22,8 +22,8 @@ public class Py4jMeta extends MetaData {
 
     // Used just as a Python exe redistributable.
     // ABSOLUTELY NO JNI/JNA IS USED
-    addDependency("org.bytedeco", "cpython-platform", "3.10.8-1.5.8");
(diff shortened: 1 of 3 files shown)


src/main/java/org/myrobotlab/service/meta/Py4jMeta.java
```diff
@@ -22,8 +22,8 @@ public Py4jMeta() {
 
     // Used just as a Python exe redistributable.
     // ABSOLUTELY NO JNI/JNA IS USED
-    addDependency("org.bytedeco", "cpython-platform", "3.10.8-1.5.8");
-    addDependency("org.bytedeco", "cpython", "3.10.8-1.5.8");
+    addDependency("org.bytedeco", "cpython-platform", "3.11.3-1.5.9");
```
AutonomicPerfectionist: As far as I am aware, upgrading this will break Py4j on raspberry pis because the arm hf and aarch64 artifacts are not listed as dependencies of cpython-platform for 3.11, we would need to add an explicit dependency on the arm 64 and armhf classifiers of the package, which we don't support in our ivy and maven wrappers

AutonomicPerfectionist (changes requested):

AutonomicPerfectionist (approved):

PR #1328, opened 2023-08-06
Boilerplate config changes 2

supertick:


doc/service-life-cycle.md
```diff
@@ -0,0 +1,52 @@
...
+    [*] --> create
+    create --> load
+
+```
+
+### start(name, type)
+Creates and starts a service with the given name and type
+
+### load
+Starts loading the heirarchy of configuration
```
AutonomicPerfectionist: heirarchy -> hierarchy

src/main/java/org/myrobotlab/document/connector/AbstractConnector.java
```diff
@@ -17,7 +17,7 @@
  * service.
  * 
  */
-public abstract class AbstractConnector extends Service implements DocumentPublisher, DocumentConnector {
+public abstract class AbstractConnector extends Service<ServiceConfig> implements DocumentPublisher, DocumentConnector {
```
AutonomicPerfectionist, line 20: I think all abstract classes should have a type variable so downstream classes can change the config type, even if none do right now. So it should be AbstractConnector<C extends ServiceConfig> extends Service<C>, and all subclasses should provide the config type when extending it

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1445,49 +1478,30 @@ public ServiceConfig getConfig() {
...
           // so doesn't conflict with remote routes
-          Listener newConfigListener = new Listener(
-                  listener.topicMethod,
-                  CodecUtils.removeEnd(
-                          listener.callbackName,
-                          '@' + Platform.getLocalInstance().getId()
-                  ),
-                  listener.callbackMethod
-          );
+          Listener newConfigListener = new Listener(listener.topicMethod, CodecUtils.removeEnd(listener.callbackName, '@' + Platform.getLocalInstance().getId()),
```
AutonomicPerfectionist: Just a note, we should probably move the removeEnd method to StringUtils, being it doesn't have anything to do with codecs

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -148,7 +155,6 @@ public abstract class Service implements Runnable, Serializable, ServiceInterfac
   /**
    * the last config applied to this service
    */
```
AutonomicPerfectionist: Remove the dangling Java doc

doc/service-life-cycle.md
supertick: :+1:

supertick: :+1:

src/main/java/org/myrobotlab/document/connector/AbstractConnector.java
supertick: :+1:

src/main/java/org/myrobotlab/framework/Service.java

src/main/java/org/myrobotlab/framework/Service.java

PR #1331, opened 2023-08-12
Better process stream handling and Outbox filter

supertick:


src/main/java/org/myrobotlab/framework/Outbox.java
```diff
@@ -223,17 +232,32 @@ public void run() {
...
+  
+  public FilterInterface removeFilter(String name, String method) {
+    return filters.remove(String.format("%s.%s", name, method));
+  }
+  
+  public boolean isFiltered(Message msg) {
+    if (filters.size() == 0 || !filters.containsKey(String.format("%s.%s", CodecUtils.getFullName(msg.name), msg.method))) {
+      return false;
+    } else {
+      return filters.get(String.format("%s.%s", msg.name, msg.method)).filter(msg);
```
AutonomicPerfectionist: msg.name needs to be converted to full name like above. Probably be easiest to factor the string format above the conditional so it can be done once and used in both branches

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -176,6 +178,19 @@ public abstract class Service<T extends ServiceConfig> implements Runnable, Seri
...
+   * the queue's max size is currently hardcoded to 10.  I waffled a bit on
+   * it being static or not. ie Should it be a global one managed by runtime,
+   * or individual ones for each service.  Since each status contains info
+   * from which service it came from, having them non-static seems redundant.
+   * From the UI perspective its more simple to manage a single one. So,
+   * for now we'll make it static. I chose the non thread safe LinkedList for
+   * performance, and currently there is no iteration over the queue, so 
+   * concurrent modification exceptions should not occur.
+   */
+  static final protected Queue<Status> statusList = new LinkedList<>();
```
AutonomicPerfectionist: ArrayBlockingQueue might have better performance due to cache coherency, linked lists require an additional layer of indirection and the array queue is optimized such that no moving of elements is required for removing the head. But, it's probably a micro optimization in the grand scheme of things so really doesn't matter that much

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -2162,8 +2178,20 @@ public Status publishError(Status status) {
...
+  /**
+   * All status are published through here error, warn and info.
+   * Usually these are events which are sent back to the user. 
+   * They are now available on th statusList queue with a buffer size of 20
+   */
   @Override
   public Status publishStatus(Status status) {
+    statusList.add(status);
+    int size = statusList.size();
+    if (size > 20) {
```
AutonomicPerfectionist: All of this could be replaced with a call to offer(), and if false just call poll once. Would need a lock around the whole section, can use a read write lock instead of synchronized so that multiple threads can read at the same time so long as none are writing.

Would need to change the status list to a queue that has a fixed capacity to use offer like this

PR #1336, opened 2023-09-03
Initial Draft of InMoov2 and Finite State Machine

supertick:


src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -1543,7 +1525,7 @@ public QueueStats publishQueueStats(QueueStats stats) {
    * @return the service
    */
   @Override
-  public Service publishState() {
+  public Service<?> publishState() {
```
AutonomicPerfectionist: Should be `Service<T>` since this is an instance method so we have access to the generic type

src/main/java/org/myrobotlab/framework/interfaces/JsonInvoker.java
```diff
@@ -0,0 +1,22 @@
+package org.myrobotlab.framework.interfaces;
+
+import org.myrobotlab.framework.Message;
+
+public interface JsonInvoker {
```
AutonomicPerfectionist: Is this for invoking messages encoded as json? Wouldn't that be covered by codec utils?

src/main/java/org/myrobotlab/opencv/OpenCVFilterYolo.java
```diff
@@ -64,10 +65,11 @@ public class OpenCVFilterYolo extends OpenCVFilter implements Runnable {
 
   transient private OpenCVFrameConverter.ToIplImage converterToIpl = new OpenCVFrameConverter.ToIplImage();
 
-  boolean debug = false;
+  protected boolean debug = false;
   transient private Net net;
   ArrayList<String> classNames;
-  public ArrayList<Classification> lastResult = null;
+  // quick fix for exploding serialization of Classification
+  transient public ArrayList<Classification> lastResult = null;
```
AutonomicPerfectionist: The serialization issue appears to occur because of the Classification `getObject()` method, Yolo sets the object to an IplImage but its never read as far as I can tell. Might be wise to convert the object to something serializable before storing it, then transient won't be needed here

src/main/java/org/myrobotlab/service/FiniteStateMachine.java
```diff
@@ -92,6 +108,13 @@ public String getNext(String key) {
 
   public void init() {
     stateMachine.init();
+    State state = stateMachine.getCurrent();
+    if (history.size() > 100) {
```
AutonomicPerfectionist: Is this simulating a circular fifo? If so, its a good candidate to be replaced with an implementation I made in a different branch

src/main/java/org/myrobotlab/service/FiniteStateMachine.java
```diff
@@ -209,21 +233,21 @@ public List<Transition> getTransitions() {
...
+   * Publishes state change (current, last and event) 
    * 
-   * @param state
+   * @param stateChange
    * @return
    */
-  public String publishNewState(String state) {
-    log.error("publishNewState {}", state);
+  public StateChange publishStateChange(StateChange stateChange) {
+    log.error("publishStateChange {}", stateChange);
```
AutonomicPerfectionist: Should be debug

src/main/java/org/myrobotlab/service/FiniteStateMachine.java
```diff
@@ -209,21 +233,21 @@ public List<Transition> getTransitions() {
...
     for (String listener : messageListeners) {
```
AutonomicPerfectionist: I think this can all be replaced with `invoke("publishState" + stateChange.current)` with a preprocess hook to not try to invoke that method, I think that should still call the listeners right?

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -93,20 +106,55 @@ public static boolean loadFile(String file) {
...
+   * (or configured) at this time, so a member variable was created to guarantee
+   * this information is available
    */
   protected boolean configStarted = false;
 
-  String currentConfigurationName = "default";
+  /**
+   * the config that was processed before booting, if there was one.
+   */
+  String bootedConfig = null;
```
AutonomicPerfectionist: Should have visibility modifer

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -140,9 +186,61 @@ public static boolean loadFile(String file) {
...
+    ledDisplayMap.put("error", new LedDisplayData(120, 0, 0, 3, 30, 30));
+    ledDisplayMap.put("info", new LedDisplayData(0, 0, 120, 1, 30, 30));
+    ledDisplayMap.put("success", new LedDisplayData(0, 0, 120, 2, 30, 30));
+    ledDisplayMap.put("warn", new LedDisplayData(100, 100, 0, 3, 30, 30));
+    ledDisplayMap.put("heartbeat", new LedDisplayData(210, 110, 0, 2, 100, 30));
+    ledDisplayMap.put("heartbeat", new LedDisplayData(210, 110, 0, 2, 100, 30));
+    ledDisplayMap.put("pirOn", new LedDisplayData(80, 200, 90, 3, 100, 30));
+    ledDisplayMap.put("onPeakColor", new LedDisplayData(180, 53, 21, 3, 60, 30));
+
+    customSoundMap.put("boot", FileIO.gluePaths(getResourceDir(), "system/sounds/Notifications/confirmation.wav"));
```
AutonomicPerfectionist, line 200: I think the sound mappings can be moved to config, right?

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -140,9 +186,61 @@ public static boolean loadFile(String file) {
...
+    stateDefaults.add("wake");
+    stateDefaults.add("firstInit");
+    stateDefaults.add("idle");
+    stateDefaults.add("random");
+    stateDefaults.add("sleep"); // listens & dreams, no opencv, waits for
+                                // wakeWord, pir active
+    stateDefaults.add("powerDown"); // stops heartbeat, listening ?
+    stateDefaults.add("shutdown");// ends mrl
+
+    ledDisplayMap.put("error", new LedDisplayData(120, 0, 0, 3, 30, 30));
```
AutonomicPerfectionist: These should be able to be moved to config as well right?

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -714,6 +802,7 @@ public boolean accept(File dir, String name) {
         for (File file : files) {
           Python p = (Python) Runtime.start("python", "Python");
```
AutonomicPerfectionist, line 797: This can be moved out of the loop

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -714,6 +802,7 @@ public boolean accept(File dir, String name) {
...
           if (p != null) {
+            // FIXME error("x") when an error occurs
             p.execFile(file.getAbsolutePath());
```
AutonomicPerfectionist, line 800: I think it would be best if we used invoke() here, should still maintain the order of files but it leverages the multithreaded pub/sub system to prevent InMoov's inbox thread from blocking for awhile

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -814,8 +780,8 @@ public void setColor(int red, int green, int blue) {
 
   @Override
   public void writeMatrix() {
-    NeoPixelConfig c = (NeoPixelConfig)config;
-    
+    NeoPixelConfig c = (NeoPixelConfig) config;
```
AutonomicPerfectionist: Shouldn't need this cast anymore

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -858,7 +822,7 @@ public void playIronman() {
   @Override
   public ServiceConfig getConfig() {
 
-    NeoPixelConfig config = (NeoPixelConfig)super.getConfig();
+    NeoPixelConfig config = (NeoPixelConfig) super.getConfig();
```
AutonomicPerfectionist: Shouldn't need cast here either

src/main/java/org/myrobotlab/service/NeoPixel.java
```diff
@@ -993,9 +969,9 @@ public void setPin(String pin) {
       error(e);
     }
   }
-  
+
   public boolean setAutoClear(boolean b) {
-    NeoPixelConfig c = (NeoPixelConfig)config;
+    NeoPixelConfig c = (NeoPixelConfig) config;
```
AutonomicPerfectionist: No cast needed

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3616,13 +3567,14 @@ public static Double getBatteryLevel() {
       } else if (platform.isLinux()) {
         // TODO This is incorrect, will not work when unplugged
         // and acpitool output is different than expected,
-        // at least on Ubuntu 22.04
-        String ret = Runtime.execute("acpitool");
-        int pos0 = ret.indexOf("Charging, ");
+        // at least on Ubuntu 22.04 - consider oshi library
```
AutonomicPerfectionist, line 3580: oshi is already pulled in as a transient dependency of some other service, in a different branch I added it as a Runtime dep and it worked great

src/main/java/org/myrobotlab/framework/Service.java

src/main/java/org/myrobotlab/opencv/OpenCVFilterYolo.java
supertick: Yep, I was hoping @kwatters would look into it, until then I'm keeping it transient, if derived from Document class.  I think Classification should be a more structured Pojo that you can expect to serialize.  Dunno the intent on the Document class

src/main/java/org/myrobotlab/service/FiniteStateMachine.java
supertick: Yes, would welcome the pr

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: it was removed

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: yup

src/main/java/org/myrobotlab/service/Runtime.java
supertick: Ya, makes sense ...  I've been looking at that for a while

src/main/java/org/myrobotlab/service/NeoPixel.java
supertick: removed all of them

src/main/java/org/myrobotlab/service/FiniteStateMachine.java
supertick: changing to info

.gitignore
```diff
@@ -116,4 +116,7 @@ build/
 /lastRestart.py
 /.factorypath
 start.yml
+config
+src/main/resources/resource/InMoov2
```
kwatters: why are these (ProgramAB / InMoov2 resources ) git ignored?

supertick: Because during development, they are the most convenient place to clone them, without going into the complexities of using git submodules

src/main/java/org/myrobotlab/service/data/Position.java
```diff
@@ -0,0 +1,43 @@
+package org.myrobotlab.service.data;
+
+public class Position {
```
kwatters: why , yet another point class?  what about roll,pitch & yaw?

src/main/java/org/myrobotlab/service/data/Pose.java
```diff
@@ -0,0 +1,22 @@
+package org.myrobotlab.service.data;
+
+public class Pose {
```
kwatters: there is already a Pose class in the kinematics package used by IK3d... why are we creating yet another Pose class?

src/main/java/org/myrobotlab/service/data/Orientation.java
```diff
@@ -11,6 +11,7 @@ public class Orientation {
   public Double roll = null;
   public Double pitch = null;
   public Double yaw = null;
+  public String src = null;
```
kwatters: should have getters/setters for src if we want it as a member here.

src/main/java/org/myrobotlab/service/data/LedDisplayData.java
```diff
@@ -1,28 +1,65 @@
...
+    this.red = red;
+    this.green = green;
+    this.blue = blue;
+    this.count = count;
+    this.timeOn = timeOn;
+    this.timeOff = timeOff;
+  }
+
+  
+  public LedDisplayData(String hexColor, int count, int timeOn, int timeOff) {
```
kwatters: add javadoc for the description that hexColor is a string with 0xRRBBGG..

src/main/java/org/myrobotlab/service/data/Position.java
supertick: I could reuse Point3f I suppose

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -2161,9 +2144,19 @@ public Status info(String format, Object... args) {
...
     return status;
   }
+  
+  public Status publishWarn(Status status) {
+    return status;
+  }
 
   @Override
   public Status publishStatus(Status status) {
+    // demux over different channels
```
AutonomicPerfectionist, line 2245: What about info status, if that's a thing?

src/main/java/org/myrobotlab/service/data/Pose.java
supertick: The Pose in IK has a lot of "stuff" not related to a instance of time Pose
Speed as an example
PoseMove ?   This doesn't seem to related to instance positional information

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
```diff
@@ -404,9 +405,9 @@ public void installDependency(String location, ServiceDependency library) {
   }
 
   @Override
-  synchronized public void install(String location, String[] serviceTypes) {
+  synchronized public void install(String location, String[] serviceTypes) throws IOException {
 
-    try {
+    // try {
```
AutonomicPerfectionist: I think catching errors and providing a more helpful error message would be good for a critical component like `install()`. We can add a `System.exit()` in the catch to prevent attempts to continue with bad state.

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
```diff
@@ -490,34 +507,43 @@ synchronized public void install(String location, String[] serviceTypes) {
...
-
-      publishStatus(Status.newInstance(Repo.class.getSimpleName(), StatusLevel.INFO, Repo.INSTALL_FINISHED, String.format("finished install of %s", (Object[]) serviceTypes)));
 
-    } catch (Exception e) {
-      error(e.getMessage());
-      log.error(e.getMessage(), e);
-    }
+        publishStatus(Status.newInstance(Repo.class.getSimpleName(), StatusLevel.INFO, Repo.INSTALL_FINISHED, String.format("finished install of %s", (Object[]) serviceTypes)));
+      }
+//    } catch (Exception e) {
```
AutonomicPerfectionist: Like above, I think catching and providing a helpful error message would be good here, like `delete ivy cache`. Then we can also terminate the program with `System.exit()` or rethrow.

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
```diff
@@ -436,8 +437,24 @@ synchronized public void install(String location, String[] serviceTypes) {
...
+      // templates [originalname](-[classifier])(-[revision]).[ext] parens are
+      // "optional"
+
+      List<String> cmd = new ArrayList<>();
+      cmd.add("-settings");
+      cmd.add(location + "/ivysettings.xml");
+      cmd.add("-ivy");
+      cmd.add(location + "/ivy.xml");
+      cmd.add("-retrieve");
+      cmd.add(location + "/jar" + "/[originalname].[ext]");
```
AutonomicPerfectionist: Should probably be using `fs` for file separator, technically the `/` works on Windows, it's just not advised.

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
```diff
@@ -436,8 +437,24 @@ synchronized public void install(String location, String[] serviceTypes) {
...
+
+      int msgLevel = 1;
```
AutonomicPerfectionist: What does 1 mean here? If it's not meant to mean anything, you can disregard the initialization entirely, like `int msgLevel;`

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
```diff
@@ -453,14 +470,14 @@ synchronized public void install(String location, String[] serviceTypes) {
...
-      ivy.getLoggerEngine().pushLogger(new IvyWrapperLogger(Message.MSG_INFO));
+      ivy.getLoggerEngine().pushLogger(new IvyWrapperLogger(msgLevel));
 
       ResolveReport report = null;
       List<String> err = new ArrayList<>();
       try {
-         report = Main.run(cmd);
-      } catch(Exception e) {
-         err.add(e.toString());
+        report = Main.run(cmd.toArray(new String[cmd.size()]));
```
AutonomicPerfectionist: Don't need `cmd.size()` here, the only reason a String array is needed is to work around generic type erasure. So you can do `new String[0]`, reduces clutter

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
```diff
@@ -490,34 +507,43 @@ synchronized public void install(String location, String[] serviceTypes) {
...
-        ArtifactDownloadReport ar = artifacts[i];
-        Artifact artifact = ar.getArtifact();
-        // String filename = IvyPatternHelper.substitute("[originalname].[ext]",
-        // artifact);
-
-        File file = ar.getLocalFile();
-        String filename = file.getAbsoluteFile().getAbsolutePath();
-        log.info("{}", filename);
+      if (report == null) {
+        String errorDetail = String.format("there was problems resolving dependencies %s", (Object[]) serviceTypes);
```
AutonomicPerfectionist:
```suggestion
        String errorDetail = String.format("There were problems resolving dependencies %s", (Object[]) serviceTypes); 

```

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
```diff
@@ -490,34 +507,43 @@ synchronized public void install(String location, String[] serviceTypes) {
...
+          String filename = file.getAbsoluteFile().getAbsolutePath();
+          log.info("{}", filename);
+
+          if ("zip".equalsIgnoreCase(artifact.getExt())) {
+            info("unzipping %s", filename);
+            try {
+              Zip.unzip(filename, "./");
+              info("unzipped %s", filename);
+            } catch (Exception e) {
+              log.error("unable to unzip file {}", filename, e);
```
AutonomicPerfectionist: I think a failure to unzip should constitute a harder error than a single error line in the log, maybe rethrow so it's caught higher up / stops the JVM?

src/main/java/org/myrobotlab/framework/repo/Repo.java
```diff
@@ -360,19 +360,19 @@ static public void info(String format, Object... args) {
     publishStatus(Status.info(format, args));
   }
 
-  synchronized public void install() {
+  synchronized public void install() throws Exception {
```
AutonomicPerfectionist: Can the exception type be more specific?

src/main/java/org/myrobotlab/kinematics/DHRobotArm.java
```diff
@@ -235,7 +235,10 @@ public boolean moveToGoal(Point goal) {
     int numSteps = 0;
     double iterStep = 0.05;
     // we're in millimeters..
-    double errorThreshold = 2.0;
+    double errorThreshold = 20.0;
+    
+    maxIterations = 1000;
```
AutonomicPerfectionist: Can all of these be given visibility modifiers?

src/main/java/org/myrobotlab/opencv/OpenCVFilterYolo.java
```diff
@@ -356,7 +358,7 @@ public void release() {
     // }
   }
 
-  volatile Object lock = new Object();
+  transient volatile Object lock = new Object();
```
AutonomicPerfectionist: Should probably use an actual `Lock` object instead of synchronizing on a regular object. Synchronization is kind of an older pattern no longer recommended

src/main/java/org/myrobotlab/service/FiniteStateMachine.java
```diff
@@ -58,6 +63,21 @@ public class Tuple {
     public Transition transition;
     public StateTransition stateTransition;
   }
+  
+  public class StateChange {
```
AutonomicPerfectionist: Probably should have javadoc for all of this, and use getters/setters instead of public fields

src/main/java/org/myrobotlab/service/Random.java
```diff
@@ -183,14 +183,17 @@ public void addRandom(long minIntervalMs, long maxIntervalMs, String name, Strin
...
     broadcastState();
   }
 
   public void process(String key) {
-    if (!enabled) {
-      return;
-    }
+    // if (!enabled) {
+    // return;
+    // }
```
AutonomicPerfectionist: Should just remove commented-out code

src/main/java/org/myrobotlab/service/InMoov2Arm.java
```diff
@@ -89,7 +89,7 @@ public static DHRobotArm getDHRobotArm(String name, String side) {
     return arm;
   }
   
-  public void onMoveArm(HashMap<String, Double> map) {
+  public void onMove(Map<String, Double> map) {
```
kwatters: do we / should we be concerned with backwards compatibility here?  perhaps instead just add this new method and have the old method name call it.. mark the old method (onMoveArm as deprecated? )

src/main/java/org/myrobotlab/service/InMoov2Hand.java
```diff
@@ -489,7 +489,7 @@ public LeapData onLeapData(LeapData data) {
     return data;
   }
 
-  public void onMoveHand(HashMap<String, Double> map) {
+  public void onMove(Map<String, Double> map) {
```
kwatters: same comment as above regarding backwards compatibility ..  we should add this as a new method and mark the only one as deprecated.

src/main/java/org/myrobotlab/service/InMoov2Head.java
```diff
@@ -199,7 +218,7 @@ public void lookAt(Double x, Double y, Double z) {
     log.info("object distance is {},rothead servo {},neck servo {} ", distance, rotation, colatitude);
   }
 
-  public void onMoveHead(HashMap<String, Double> map) {
+  public void onMove(Map<String, Double> map) {
```
kwatters: backwards compatability... we should add this new method and mark the old one as deprecated.

src/main/java/org/myrobotlab/service/InMoov2Torso.java
```diff
@@ -93,7 +93,7 @@ public void disable() {
       lowStom.disable();
   }
   
-  public void onMoveTorso(HashMap<String, Double> map) {
+  public void onMove(Map<String, Double> map) {
```
kwatters: backwards compatibility should be maintained here..  add this method.. mark the old one as deprecated.. make the old one call this one instead.

src/main/java/org/myrobotlab/service/InverseKinematics3D.java
```diff
@@ -247,6 +247,12 @@ public void publishTelemetry(String name) {
       log.info("Servo : {}  Angle : {}", jointName, angleMap.get(jointName));
     }
     invoke("publishJointAngles", angleMap);
+    
+    InMoov2 i01 = (InMoov2)Runtime.getService("i01");
```
kwatters: this should definitely not be needed.  inmoov should subscribe to publishJointAngles if it cares...

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
```diff
@@ -2157,7 +2157,7 @@ public void simpleInitApp() {
     new File(getDataDir()).mkdirs();
     new File(getResourceDir()).mkdirs();
 
-    // assetManager.registerLocator("./", FileLocator.class);
+    assetManager.registerLocator("./", FileLocator.class);
```
kwatters: is there a way to do this without a relative path?  i worry 2 things.. 1 the current working directory might change.. also , 2 that this won't work on windows.

src/main/java/org/myrobotlab/service/data/Position.java
AutonomicPerfectionist: There's also `org.myrobotlab.kinematics.Point`, and for roll, pitch, and yaw there's `org.myrobotlab.service.data.Orientation`

Really, all of these should be consolidated into one hierarchy at some point, with a `Point` specifying only coordinates, an `Orientation` specifying only rotational values, and `Position` containing both a Point marking its origin and an Orientation

src/main/java/org/myrobotlab/service/InverseKinematics3D.java
supertick: EEK ! WTF ?!?!?
that was an accident .. thx

src/main/java/org/myrobotlab/service/data/Pose.java
AutonomicPerfectionist: This can be consolidated into a singular type hierarchy along with Point, Orientation, and Position. If a Position has a Point origin and an Orientation, then a Pose can have a Position and a time and name

src/main/java/org/myrobotlab/service/JMonkeyEngine.java
supertick: The assetManager is owned by JMonkey ... I've found it a bit of a pita loading resources.
relative pathing is required to work, that's how default config for jmonkey or jmonkey config in general works
(both relative and absolute)
In the past there was issues because stuff was re-translated into absolute, it should work as its expressed in config

src/test/java/org/myrobotlab/service/ClockTest.java
```diff
@@ -37,7 +37,7 @@ public void testService() throws Exception {
     assertEquals(interval, clock.getInterval());
 
     clock.startClock();
-    Service.sleep(10);
+    Service.sleep(100);
```
kwatters, line 40: why the need to increase this sleep?  it'd be nice to not have any sleep ...

supertick: I agree - but it fairly consistently failed the build... lots of tests probably could use refactoring and better mocking of asynchronous things

src/main/java/org/myrobotlab/service/config/LogConfig.java
```diff
@@ -0,0 +1,12 @@
+package org.myrobotlab.service.config;
+
+public class LogConfig extends ServiceConfig {
+
+  public String currentUserName;
```
kwatters: why does log config care about currentUserName?

supertick: copy paste config from programab .. trying to keep up get services their own real config vs ServiceConfig 
removing...

src/main/java/org/myrobotlab/service/config/NeoPixelConfig.java
```diff
@@ -14,7 +14,7 @@ public class NeoPixelConfig extends ServiceConfig {
   public Integer brightness = 255;
   public boolean fill = false;
   // auto clears flashes
-  public boolean autoClear = true;
+  public boolean autoClear = false;
```
kwatters: why the change in default behavior here?

supertick: as it was originally implemented, it will interfere with animations 
I think the purpose was to clear when a flash was done, but a flash should specify its own clearing or resetting

supertick: also, it wasn't defined if "clear" was restore what was there before, or make blank

src/main/java/org/myrobotlab/service/config/OakDConfig.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.myrobotlab.framework.Plan;
+
+public class OakDConfig extends ServiceConfig {
+
+  
+  /**
+   * install through py4j
+   */
+  public boolean py4jInstall = true;
```
kwatters: why does this need to be part of the config?  setting this to false is just not valid...

src/main/java/org/myrobotlab/service/data/LedDisplayData.java
```diff
@@ -1,28 +1,65 @@
...
+
+    // remove "#" or "0x" prefix if present
+    hexColor = hexColor.replace("#", "").replace("0x", "");
+
+    this.count = count;
+    this.timeOn = timeOn;
+    this.timeOff = timeOff;
+    this.red = Integer.parseInt(hexColor.substring(0, 2), 16);
```
kwatters: what if the color is something like 0xFFFFFE  this is valid hex color but parseInt will fail...

src/test/java/org/myrobotlab/service/ServoMixerTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.junit.Test;
+import org.myrobotlab.kinematics.Gesture;
+import org.myrobotlab.kinematics.GesturePart;
+import org.myrobotlab.test.AbstractTest;
+
+public class ServoMixerTest extends AbstractTest {
+
+  @Test
+  public void testService() throws Exception {
```
kwatters, line 11: there is no assert here... what are we testing?

src/main/java/org/myrobotlab/framework/interfaces/JsonInvoker.java
```diff
@@ -0,0 +1,22 @@
...
+   */
+  public Object invoke(String method);
+
+  /**
+   * Encoded parameters as a JSON String (encoded once!)
+   * @param method
+   * @param encodedParameters
+   * @return
+   */
+  public Object invoke(String method, String encodedParameters);
```
kwatters: seems like if we are going to support json invoking, we should have default implementation in the interface to handle it.

src/test/java/org/myrobotlab/service/ServoMixerTest.java
supertick: ya needs some

src/main/java/org/myrobotlab/framework/interfaces/JsonSender.java
```diff
@@ -0,0 +1,12 @@
+package org.myrobotlab.framework.interfaces;
+
+public interface JsonSender {
+
+  /**
+   * Send interface which takes a json encoded Message.
+   * For schema look at org.myrobotlab.framework.Message
+   * @param jsonEncodedMessage
+   */
+  public void send(String jsonEncodedMessage);
```
kwatters: similar to the JsonInvoker interface, we should have a default implementation here so that specific services don't need to parse this jsonEncodedMessage.

src/main/java/org/myrobotlab/framework/interfaces/JsonInvoker.java
supertick: ya .. probably possible

src/main/java/org/myrobotlab/service/config/OakDConfig.java
supertick: I think @AutonomicPerfectionist added it ...
It is valid in the sense that Py4j service java-land can connect to a non-self installed python instance if the Py4j.py script is run.

src/main/java/org/myrobotlab/kinematics/DHRobotArm.java
kwatters: if you want a different default value for this it should be specified above, not overridden here.

src/main/java/org/myrobotlab/service/config/OakDConfig.java
AutonomicPerfectionist: I did not add that; my method of python services doesn't use py4j nor does it use really any Java-side per-service components, it's all handled in Python land

supertick: oop, thought this was in the Py4j service context ...
yeah, this is starting to think about how Py4j & OakD would work with single click ...
The OakD demo installs were non trivial ... 
Non trivial in that, If I remember they did a lot more than just install a pip module ...
https://docs.luxonis.com/en/latest/pages/tutorials/first_steps/ 
I'm not sure how we want to handle it .. from a consumer's point of view the goal almost always is ... I click worky
Getting it to work that way makes me think docker might be the way to go ... :shrug:

AutonomicPerfectionist: Looking through the code, this looks much more akin to my first attempt at Python services, where each service needed a Java-side proxy. I think focusing efforts on the new system I've designed would be a better approach, as adding a new Python service would not involve duplicating method definitions between Java and Python, and python services are allowed to exist separately from a singular gateway implementation (nothing stops mrlpy from supporting py4j as a gateway, but it also works across websockets and potentially other gateways)

src/main/java/org/myrobotlab/kinematics/DHRobotArm.java
supertick: reverting

src/main/java/org/myrobotlab/service/config/OakDConfig.java
AutonomicPerfectionist: > Non trivial in that, If I remember they did a lot more than just install a pip module ...
https://docs.luxonis.com/en/latest/pages/tutorials/first_steps/

Oh yeah, we definitely want Docker for that

supertick: Yeah .. docker .. but that means making a Docker service in java land if we want to use it there ...
it also means adding infrastructure e.g. Docker registry account etc ... 
As regards to the Java-Land proxy vs Python stand alone service I totally agree and don't want required Java proxies to exists for Python.  I'd prefer it be network only requirements.  
It's trivial to have a python client with mrl .. but there currently is no service framework within Python that can mirror a Java-land instance - until that happens I don't believe a python only service is an option.

AutonomicPerfectionist: My mrlpy project can, and has had the ability for a while. You can write Python services with it right now, and it will autogenerate python-land proxies for Java services. The only missing part is allowing a user to start a Python service from the WebGui, which means that Java-land needs the capacity to start the Python-side runtime. That's what #1341 is doing.

An example of this is here:

https://github.com/AutonomicPerfectionist/mrlpy/blob/develop/examples/llama-with-solr.py

That particular example won't work with the latest MRL because it uses service interfaces that haven't been merged, but simpler examples do work out of the box (or at least did last I checked a few weeks ago)

src/main/java/org/myrobotlab/service/data/Pose.java
kwatters: The original Pose class didn't have the junk that you mention.  Actually, in the commit history, @supertick  you were the one that added that stuff.  The Pose class as originally intended was to have the servo positions, the servo speeds , a name , and some metadata about when it was created.  The main point of the Pose class was so that it could serialize into the pose.json...

src/main/resources/resource/BoofCV/intrinsic.yaml
```diff
@@ -0,0 +1,19 @@
+# Pinhole camera model with radial and tangential distortion
+# (fx,fy) = focal length, (cx,cy) = principle point, (width,height) = image shape
+# radial = radial distortion, (t1,t2) = tangential distortion
+
+pinhole:
```
kwatters, line 5: why is this file checked in?  What uses it?  why these specific calibrations?

src/main/resources/resource/BoofCV/visualdepth.yaml
```diff
@@ -0,0 +1,20 @@
+# RGB Depth Camera Calibration
```
kwatters, line 1: why is this file checked in?  what uses it?  I see no references to it in the code base.

src/main/java/org/myrobotlab/service/data/Pose.java
supertick: alright .. I'll normalize them

kwatters: If this class if fundamentally something different, we should name it something other than Pose because we already have a Pose class..

supertick: its not, it should be reconciled/normalized

src/main/java/org/myrobotlab/service/data/Orientation.java
supertick: Its silly to have both setters and getters and public fields ... I'll remove the setters/getters

src/main/java/org/myrobotlab/service/InMoov2Arm.java
supertick: mkay

src/main/java/org/myrobotlab/service/InMoov2Hand.java
supertick: :+1:

src/main/java/org/myrobotlab/service/InMoov2Head.java
supertick: :+1:

src/main/java/org/myrobotlab/service/InMoov2Torso.java
supertick: :+1:

src/main/java/org/myrobotlab/framework/Service.java
supertick: Not yet .. not sure of how to handle this .. but, I find it frustrating in the webgui that there is errors, from publishError (part of the publishStatus set) disappears on refresh :( 
I think maybe there needs to be a circular buffer of status .. but don't know if there should be a centralized on or per service ... per service would make it cleaner.  Haven't started anything but the comment.

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
supertick: Ya, the nicest thing I can think about is mrl startsup (since it has all required packages to run Runtime, Security & WebGUI - it should never be thwarted by bad dependency install) ... When it starts up is should clearly show what dependencies have failed .. I think I made it show services not installed, but dunno all the details of its current behavior in that regard.

src/main/java/org/myrobotlab/service/Random.java
supertick: :+1:

pom.xml
```diff
@@ -1734,375 +1774,386 @@
...
+</dependency>
+
+
+  	<dependency>
+	    <groupId>org.mockito</groupId>
+	    <artifactId>mockito-core</artifactId>
+	    <version>3.12.4</version>
+	    <scope>test</scope>
+	</dependency>
+  	<dependency>
```
kwatters: the pom for this is missing in the mrl repo so it doesn't cache properly when doing offline builds.

pom.xml
```diff
@@ -1646,6 +1674,18 @@
...
 <!-- WebSocketConnector end -->
 
+  <dependency>
+      <groupId>au.edu.federation.caliko</groupId>
+      <artifactId>caliko</artifactId>
+      <version>1.3.8</version>
+    </dependency> 
+    
+        <dependency>
+      <groupId>au.edu.federation.caliko.visualisation</groupId>
```
kwatters: the pom for this needs to be added to the mrl repo for it to build and cache the jar properly.

pom.xml
kwatters: you can see the issue when running a mvn build from the command line with the following error message.

[WARNING] The POM for au.edu.federation.caliko.visualisation:caliko-visualisation:jar:1.3.8 is missing, no dependency information available
[WARNING] The POM for au.edu.federation.caliko.demo:caliko-demo:jar:1.3.8 is missing, no dependency information available

pom.xml
supertick: hmm .. thought i put it there ..
https://repo.myrobotlab.org/artifactory/myrobotlab/au/edu/federation/caliko/visualisation/caliko-visualisation/1.3.8/

supertick: mebbe i messed up the path

src/main/java/org/myrobotlab/service/config/VertxConfig.java
```diff
@@ -0,0 +1,9 @@
+package org.myrobotlab.service.config;
+
+public class VertxConfig extends ServiceConfig {
```
kwatters: why are there 2 vertxconfigs, one with a capital X and one with a lower case x ?

src/main/java/org/myrobotlab/service/config/VertXConfig.java
```diff
@@ -0,0 +1,17 @@
+package org.myrobotlab.service.config;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class VertXConfig extends ServiceConfig {
```
kwatters: there are two of these classes.. one with a capital X and one with a lower case x... please remove/rename one of them as appropriate.

supertick: deleted the capital X one

src/main/java/org/myrobotlab/opencv/OpenCVFilterYolo.java
kwatters: I'll polish this one up.. I don't see any need for the IplImage in the classification at all  ..  perhaps the easiest thing here is to just not add the original frame.  (if we care about the original frame that generated the classifications (which we might) then we can serialize into a buffered image.  (this is serializeable.)

src/main/java/org/myrobotlab/service/config/VertxConfig.java
supertick: removed the incorrect one

src/main/java/org/myrobotlab/service/InMoov2Arm.java
supertick: updated

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
supertick: heh .... I find the String[0] cluttery as well ;)

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
supertick: updated with meaning

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
supertick: kinda adverse to add more clutter .. it would need to be Service.fs

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
supertick: updated

src/main/java/org/myrobotlab/framework/repo/IvyWrapper.java
supertick: updated

src/main/java/org/myrobotlab/framework/repo/Repo.java
supertick: updated

src/main/java/org/myrobotlab/service/data/Pose.java
supertick: on closer inspection this is fundamentally different ...
The Pose you constructed is a set of positions of servo rotations...
This Pose I'm trying to describe is 3D position and orientation of some point - in the case where its being used its describing the position and orientation of :
* the headset
* the left controller
* the right controller
its more than a 3d point because it has orientation

src/main/java/org/myrobotlab/service/data/Position.java
supertick: added https://github.com/MyRobotLab/myrobotlab/issues/1363

src/main/java/org/myrobotlab/service/data/LedDisplayData.java
supertick: why would it fail its 255, 255, 254

AutonomicPerfectionist: For hex you'd need to pass in the radix, otherwise it will assume base ten and throw an exception on a non numeric character

Edit: nevermind, I see you already did that

PR #1343, opened 2023-09-25
fix speed issue

hairygael:


kwatters (approved):

kwatters:
looks good.  hopefully now setting the head speed will work as expected.

PR #1345, opened 2023-10-10
mouth listen servoMixer.java

hairygael:


supertick (changes requested):
other way around ... the mouth is a listener of servo mixer ;)

hairygael:
Ok thanks,
I am also trying to have the servoMixer text published in the chatBot UI but without the chatBot answering to it.

supertick (approved):

PR #1358, opened 2023-10-21
Peer config utilities

supertick:
set of utilities that will allow parent services to save config of peers

### DIFF
--- a/src/main/java/org/myrobotlab/service/Runtime.java
+++ b/src/main/java/org/myrobotlab/service/Runtime.java
@@ -61,6 +61,7 @@ import org.myrobotlab.framework.ProxyFactory;
 import org.myrobotlab.framework.Registration;
 import org.myrobotlab.framework.Service;
 import org.myrobotlab.framework.ServiceReservation;
+import org.myrobotlab.framework.StaticType;
 import org.myrobotlab.framework.Status;
 import org.myrobotlab.framework.interfaces.ConfigurableService;
 import org.myrobotlab.framework.interfaces.MessageListener;
@@ -1152,6 +1153,14 @@ public class Runtime extends Service<RuntimeConfig> implements MessageListener, 
     return registry;// FIXME should return copy
   }
 
+  public static ServiceInterface getService(String inName) {
+    return getService(inName, new StaticType<>() {});
+  }
+
+  public static <C extends ServiceConfig, S extends ServiceInterface & ConfigurableService<C>> S getConfigurableService(String inName, StaticType<S> serviceType) {
+    return getService(inName, serviceType);
+  }
+
   /**
    * Gets a running service with the specified name. If the name is null or
    * there's no such service with the specified name, returns null instead.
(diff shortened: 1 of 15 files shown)


src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -141,7 +141,7 @@ public class Runtime extends Service<RuntimeConfig> implements MessageListener,
    * all these requests.
    */
   @Deprecated /* use the filesystem only no memory plan */
-  transient final Plan masterPlan = new Plan("runtime");
+  protected final Plan masterPlan = new Plan("runtime");
```
AutonomicPerfectionist, line 145: This probably needs to be transient, right?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -5154,6 +5152,10 @@ private void savePlanInternal(String configName) {
 
     for (String s : masterPlan.keySet()) {
       String filename = CONFIG_ROOT + fs + configName + fs + s + ".yml";
+      File check = new File(filename);
+      if (check.exists()) {
+        continue;
```
AutonomicPerfectionist, line 5175: Doesn't this not update the config if the file already exists? Seems counterintuitive that it would refuse to overwrite it silently

src/main/java/org/myrobotlab/service/config/OpenCVConfig.java
```diff
@@ -15,5 +15,9 @@ public class OpenCVConfig extends ServiceConfig {
   public boolean webViewer = false;
   public boolean capturing = false;
   public Map<String, OpenCVFilter> filters = new LinkedHashMap<>();
+  /**
+   * flip the video vertically
+   */
+  public boolean flip = false;
```
AutonomicPerfectionist, line 21: Probably fine for now, but in the future we probably want to move this to a filter-specific config that can be embedded in the OpenCV config

src/main/java/org/myrobotlab/service/Runtime.java
supertick: Turns out there is a real purpose for this.
Probably, the class and the variable should have a better name, but its a map of references to "all" config .. both running and all peer config regardless if the services are running or not.   It's kinda like a configuration service, but its just a map maintained by the Runtime service.  As far as I know, its the easiest (best?) way to manage a sync'd reference for the UI to use.

src/main/java/org/myrobotlab/service/Runtime.java
supertick: I need to check, but I think its different from a Service save .. or a save all.
This function saves a "plan" .. a plan includes peer configuration, even if the peers are not running (ie defaults).
A save all, saves all services that are running.
If a file exists on the filesystem, it was either created by the default definition, or by the user and neither of these would need to be overwritten.

src/main/java/org/myrobotlab/service/config/OpenCVConfig.java
supertick: Filters can be embedded currently, there is a filters field, the same functionality could be added by adding a Filter definition to OpenCVConfig.filters.
This is a convenience config, because of the very often occurrence or people mounting cameras upside down

AutonomicPerfectionist (approved):

PR #1370, opened 2023-11-29
Proposal to be able to set botname or username in aiml

supertick:


supertick:
found better way to create new user without python in aiml

PR #1388, opened 2024-01-07
Changes allowing vertx to replace webgui

supertick:


src/main/java/org/myrobotlab/service/Vertx.java
```diff
@@ -184,7 +196,34 @@ public boolean isLocal(Message msg) {
...
+   * 
+   * @param port
+   */
+  public void setPort(int port) {
+    config.port = port;
+    boolean wasListening = isListening;
+    if (isListening) {
+      stop();
+    }
+    sleep(2000);
```
kwatters, line 211: do we really need this sleep here?  shouldn't stop block until it's actually stopped?

src/main/java/org/myrobotlab/service/Vertx.java
```diff
@@ -41,6 +43,11 @@ public class Vertx extends Service<VertxConfig> implements Gateway {
 
   private transient io.vertx.core.Vertx vertx = null;
 
+  /**
+   * If listening currently on port
+   */
+  protected boolean isListening = false;
```
kwatters: we should rename this to just "listening"

src/main/java/org/myrobotlab/vertx/ApiHandler.java
```diff
@@ -0,0 +1,205 @@
...
+      // content type always json for now
+      request.response().putHeader("content-type", "application/json");
+
+      // POST vs GET handler ?
+      if ("get".equals(verb)) {
+        handleGet(rc);
+      } else if ("post".equals(verb)) {
+        handlePost(rc);
+      } else {
+        log.info("got a request to handle http verb %s", verb);
```
kwatters: this should probably be a warn level.

src/main/java/org/myrobotlab/service/Vertx.java
supertick: updated

src/main/java/org/myrobotlab/vertx/ApiHandler.java
supertick: updated

PR #1390, opened 2024-01-10
Runtime updates

supertick:


src/main/java/org/myrobotlab/service/config/ServiceConfig.java
```diff
@@ -257,6 +255,10 @@ public static Plan getDefault(Plan plan, String name, String inType) {
         log.info("could not find config class for {}, loading generalized ServiceConfig", inType);
         ServiceConfig sc = new ServiceConfig();
         sc.type = inType;
+        if (name.equals("i01")) {
+          log.info("here");
+        }
+        Collections.sort(sc.listeners, new MrlListenerComparator());
```
kwatters, line 261: why the need for the sort here?  it'd be nice to preserve the original order, wouldn't it?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1399,7 +1402,7 @@ public static final long getTotalMemory() {
    */
   public static final synchronized long getUniqueID() {
     ++uniqueID;
-    return uniqueID;
+    return System.currentTimeMillis();
```
kwatters, line 1405: This seems odd..  current time milliseconds isn't very unique..  why not a uuid like the fixme mentions above?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1552,7 +1555,8 @@ public void run() {
             r.getRepo().install(serviceType);
           }
         } catch (Exception e) {
-          r.error(e);
+          r.error("dependencies failed - install error", e);
+          throw new RuntimeException(String.format("dependencies failed - install error %s", e.getMessage()));
```
kwatters, line 1559: Does this cause the install to exit if it fails?  (I believe that's a good behavior. just curious)

src/main/java/org/myrobotlab/service/config/ServiceConfig.java
supertick: Considering the amount of some service's subscriptions its much easier to examine and maintain if they are sorted.
The way they were entered in has no impact at how subscriptions are processed.
Trying to diff a checked in config vs new unsorted is difficult

src/main/java/org/myrobotlab/service/Runtime.java
supertick: not sure, ivy in general doesn't throw, but the install process does log out errors to a preserved myrobotlab-install.log
its been slightly improved upstream, but still room for clear notification for the user

src/main/java/org/myrobotlab/service/Runtime.java
supertick: heh, sure ... I thought the same thing,
but considering it will be changing the core message from long to string I figured it be worth investigating in a separate pull request ;)

src/main/java/org/myrobotlab/framework/MrlListenerComparator.java
```diff
@@ -0,0 +1,12 @@
+package org.myrobotlab.framework;
+
+import java.util.Comparator;
+
+import org.myrobotlab.service.config.ServiceConfig.Listener;
+
+public class MrlListenerComparator implements Comparator<Listener> {
```
AutonomicPerfectionist, line 13: Just a simple javadoc would be good here, explaining it sorts listeners by the listener method only

src/main/java/org/myrobotlab/framework/StartYml.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Configuration for starting.  If a start.yml is saved in the root mrl will 
+ * use this on startup to determine which configuration set to use.  Also instance id is
+ * maintainable here.
+ * 
+ * @author GroG
+ *
+ */
+public class StartYml {
```
AutonomicPerfectionist, line 11: Can this be in a different package since it's only used as a POJO? `framework` is already a tad cluttered

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -133,7 +136,7 @@ public class Runtime extends Service<RuntimeConfig> implements MessageListener,
    * a registry of all services regardless of which environment they came from -
    * each must have a unique name
    */
-  static private final Map<String, ServiceInterface> registry = new TreeMap<>();
+  static private Map<String, ServiceInterface> registry = new TreeMap<>();
```
AutonomicPerfectionist: Why can't this be final anymore?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -436,9 +434,8 @@ synchronized private static Map<String, ServiceInterface> createServicesFromPlan
         runtime.error("could not get %s from plan", service);
         continue;
       }
-      sc.state = "CREATING";
+      // sc.state = "CREATING";
```
AutonomicPerfectionist: If we don't need this anymore, can we entirely remove this line?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -2006,11 +2006,19 @@ synchronized public static void unregister(String inName) {
 
     // FIXME - release autostarted peers ?
 
-    // last step - remove from registry
-    registry.remove(name);
+    // last step - remove from registry by making new registry
+    // thread safe way
+    Map<String, ServiceInterface> removedService = new TreeMap<>();
```
AutonomicPerfectionist, line 2011: A better way would be to use a `ConcurrentHashMap` and simply call `remove()` on it. If the key was already removed elsewhere, the call simply returns, no error thrown

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -3381,13 +3361,16 @@ static public void disableLogging() {
...
   @Override
   public void releaseService() {
-    super.releaseService();
     if (runtime != null) {
+      runtime.purgeTasks();
+      runtime.stopService();
       runtime.stopInteractiveMode();
-
       runtime.getRepo().removeStatusPublishers();
+      registry = new TreeMap<>();
```
AutonomicPerfectionist, line 3369: This can be done with a `ConcurrentHashMap` and a call to `clear()`

src/main/java/org/myrobotlab/framework/StartYml.java
supertick: Its a global that affects the whole instance... I'd consider it if someone came up with a great package name ...
I know I cant :laughing:

src/main/java/org/myrobotlab/service/Runtime.java
supertick: ConcurrentHashMap isn't sorted, and TreeMap isn't thread-safe, so I made it thread safe, the consequence is registry can't be final.

src/main/java/org/myrobotlab/service/Runtime.java
supertick: ConcurrentHashMap isn't sorted TreeMap is, to make it threadsafe removal actions need to re-new it

src/main/java/org/myrobotlab/service/Runtime.java
supertick: absolutely ... :+1:

src/main/java/org/myrobotlab/service/Runtime.java
supertick: TreeMap is desired, so I made it threadsafe

src/main/java/org/myrobotlab/framework/MrlListenerComparator.java
supertick: I agree :+1:

src/main/java/org/myrobotlab/framework/MrlListenerComparator.java
```diff
@@ -0,0 +1,20 @@
+package org.myrobotlab.framework;
+
+import java.util.Comparator;
+
+import org.myrobotlab.service.config.ServiceConfig.Listener;
+
+/**
+ * 
+ * @author GroG
```
AutonomicPerfectionist, line 11: `@author` should be at the bottom so it renders correctly for all javadoc parsers

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -2006,11 +2005,19 @@ synchronized public static void unregister(String inName) {
...
-    registry.remove(name);
+    // last step - remove from registry by making new registry
+    // thread safe way
+    Map<String, ServiceInterface> removedService = new TreeMap<>();
+    for (String key : registry.keySet()) {
+      if (!name.equals(key)) {
+        removedService.put(key, registry.get(key));
+      }
+    }
+    registry = removedService;
```
AutonomicPerfectionist, line 2017: This doesn't guarantee that all threads will immediately see the updated registry, since the CPU caches will hang on to it. Needs to be marked volatile. Additionally, any code storing the registry will only ever see the old one, never the new since it hangs onto the old reference.

src/main/java/org/myrobotlab/framework/MrlListenerComparator.java
supertick: :+1:

src/main/java/org/myrobotlab/service/Runtime.java
supertick: Its private, so all consuming classes outside of runtime should be using accessors.
The getRegistry really should be returning a copy, which in general should be the case of a getter on a private, otherwise what's the point, unless its immutable?  The getRegistry does have a FIXME but not including it in this pr. I'll mark it as volatile.

PR #1396, opened 2024-01-20
ProgramAB xml models and parser

supertick:


src/main/java/org/myrobotlab/programab/XmlParser.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Thread safe fasterjackson xml parser.
+ * 
+ * @author GroG
+ *
+ */
+public class XmlParser {
+
+  public static Template parseTemplate(String xml) throws JsonMappingException, JsonProcessingException {
+    ThreadLocal<XmlMapper> xmlMapperThreadLocal = ThreadLocal.withInitial(XmlMapper::new);
```
AutonomicPerfectionist: Thread locals should be a static field, creating it every time the method is run doesn't achieve anything (in fact, it's the same as no thread local at all, since the thread local wrapper is garbage collected once the method returns)

src/main/java/org/myrobotlab/programab/XmlParser.java
```diff
@@ -0,0 +1,31 @@
...
+    XmlMapper xmlMapper = xmlMapperThreadLocal.get();
+    Template template = xmlMapper.readValue(xml, Template.class);
+    return template;
+  }
+
+  public static Sraix parseSraix(String xml) throws JsonMappingException, JsonProcessingException {
+    ThreadLocal<XmlMapper> xmlMapperThreadLocal = ThreadLocal.withInitial(XmlMapper::new);
```
AutonomicPerfectionist: Same as above

src/main/java/org/myrobotlab/programab/handlers/oob/OobProcessor.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.List;
+
+import org.myrobotlab.codec.CodecUtils;
+import org.myrobotlab.framework.Message;
+import org.myrobotlab.programab.models.Mrl;
+import org.myrobotlab.programab.models.Oob;
+import org.myrobotlab.service.ProgramAB;
+
+public class OobProcessor {
```
AutonomicPerfectionist, line 11: Java doc would be good

src/main/java/org/myrobotlab/programab/handlers/oob/OobProcessor.java
```diff
@@ -0,0 +1,93 @@
...
+
+  private static OobProcessor instance;
+  private transient ProgramAB programab;
+  protected int maxBlockTime = 2000;
+
+  private OobProcessor() {
+  }
+
+  public static OobProcessor getInstance(ProgramAB programab) {
```
AutonomicPerfectionist: If I understand this correctly, this won't work for multiple program AB instances

src/main/java/org/myrobotlab/programab/models/Mrl.java
```diff
@@ -0,0 +1,14 @@
+package org.myrobotlab.programab.models;
+
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlElementWrapper;
+
+public class Mrl {
```
AutonomicPerfectionist, line 8: This looks like another Message essentially, anyway we can use Message directly?

src/main/java/org/myrobotlab/programab/models/Sraix.java
```diff
@@ -0,0 +1,10 @@
+package org.myrobotlab.programab.models;
+
+// FIXME add attributes and internal tags
+public class Sraix {
```
AutonomicPerfectionist, line 4: Should have a Java doc

src/main/java/org/myrobotlab/programab/models/Template.java
```diff
@@ -0,0 +1,51 @@
+package org.myrobotlab.programab.models;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.dataformat.xml.XmlMapper;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlText;
+
+//@JacksonXmlRootElement(localName = "template")
+//@JsonIgnoreProperties(ignoreUnknown = true)
```
AutonomicPerfectionist, line 9: Can we remove these commented out pieces?

src/main/java/org/myrobotlab/programab/models/Template.java
```diff
@@ -0,0 +1,51 @@
...
+public class Template {
+  // @JacksonXmlElementWrapper(useWrapping = false)
+  
+  @JacksonXmlProperty(localName = "template")
+  
+  @JacksonXmlText
+  public String text;  
+  
+  
+public Oob oob;
```
AutonomicPerfectionist, line 20: Formatting looks weird here, like it's missing an indent

src/main/java/org/myrobotlab/programab/models/Template.java
```diff
@@ -0,0 +1,51 @@
...
+//
+//  public List<Oob> getOob() {
+//    return oob;
+//  }
+//
+//  public void setOob(List<Oob> oob) {
+//    this.oob = oob;
+//  }
+  
+  public static void main(String[] args) {
```
AutonomicPerfectionist, line 33: Can we remove this and put it in a test instead?

src/main/java/org/myrobotlab/programab/models/Oob.java
```diff
@@ -0,0 +1,14 @@
+package org.myrobotlab.programab.models;
+
+import java.util.List;
+
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlElementWrapper;
+
+public class Oob {
```
AutonomicPerfectionist, line 7: Java doc would be nice

src/main/java/org/myrobotlab/programab/models/Event.java
```diff
@@ -0,0 +1,65 @@
+package org.myrobotlab.programab.models;
+
+/**
+ * Pojo for state change of one of ProgramAB's state info
+ * @author GroG
+ *
+ */
+public class Event {
```
AutonomicPerfectionist, line 8: Can all of these fields be made final by chance? Immutability is really nice for concurrency 🙂

src/main/java/org/myrobotlab/programab/handlers/sraix/MrlSraixHandler.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public MrlSraixHandler(ProgramAB programab) {
+    this.programab = programab;
+  }
+
+  @Override
+  public String sraix(Chat chatSession, String input, String defaultResponse, String hint, String host, String botid, String apiKey, String limit, Locale locale) {
+    try {
+      log.debug("MRL Sraix handler! Input {}", input);
+      String xml = String.format("<sraix>%s</sraix>", input);
```
AutonomicPerfectionist, line 29: Not sure this is a good way to handle this, any input containing non escaped XML characters are going to cause it to blow up

src/main/java/org/myrobotlab/programab/handlers/sraix/MrlSraixHandler.java
```diff
@@ -0,0 +1,58 @@
...
+import org.myrobotlab.programab.models.Sraix;
+import org.myrobotlab.service.ProgramAB;
+import org.myrobotlab.service.data.SearchResults;
+import org.myrobotlab.service.interfaces.SearchPublisher;
+import org.slf4j.Logger;
+
+public class MrlSraixHandler implements SraixHandler {
+  transient public final static Logger log = LoggerFactory.getLogger(MrlSraixHandler.class);
+
+  private ProgramAB programab = null;
```
AutonomicPerfectionist: Final maybe?

src/main/java/org/myrobotlab/programab/handlers/oob/OobProcessor.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+          }
+        }
+      } // for each mrl
+    }
+
+    // Process <oob><mrljson>
+    if (oob != null && oob.mrljson != null) {
+
+      Message[] msgs = CodecUtils.fromJson(oob.mrljson, Message[].class);
```
AutonomicPerfectionist, line 64: Doesn't go through the proper codec utils message deserialization, so things like method cache isn't going to work. Side note: we should just move the specialized logic into a custom message deserializer now that we don't need to support GSON, but that's a project for another day

src/main/java/org/myrobotlab/programab/handlers/oob/OobProcessor.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+    return instance;
+  }
+
+  public Message toMsg(Mrl mrl) {
+    Object[] data = null;
+    if (mrl.params != null) {
+      data = new Object[mrl.params.size()];
+      for (int i = 0; i < data.length; ++i) {
+        data[i] = mrl.params.get(i).trim();
```
AutonomicPerfectionist, line 26: Looks like this will end up only working for String parameters. Any way we can add an XML property to the param tag defining the type? Then we can take the string of the param and deserialize to the given type

AutonomicPerfectionist (changes requested):

src/main/java/org/myrobotlab/programab/XmlParser.java
supertick: it was suggested as a way to make the xml parser threadsafe, making it static would defeat that no ?

src/main/java/org/myrobotlab/programab/handlers/oob/OobProcessor.java
supertick: :+1:

src/main/java/org/myrobotlab/programab/XmlParser.java
AutonomicPerfectionist: The thread local needs to be static, then when you call the `get()` method in the code it returns or creates a new one that is only usable for that thread. The way it is now, the thread local wrapper is created every time and then garbage collected at the end of the method, so it can't reuse a previously created parser in the same thread if the method is called multiple times because the wrapper that keeps track of it is gone.

Technically, it's thread safe still, but only because it recreates the parser on every invocation of the method, which you could do without the thread local wrapper

supertick: Thanks ! :+1:

src/main/java/org/myrobotlab/programab/handlers/oob/OobProcessor.java
supertick: Good point .. there might be problems... not sure I'll needs some tests.
The reason its a singleton is because it needed in a session and the sriax handler.
I was trying to limit resources of xml parser creation.
Probably a simple way is to have this part of ProgramAB, since both the sessions and the handler have references to this.

src/main/java/org/myrobotlab/programab/models/Mrl.java
supertick: No this is the aiml definition .. legacy, its slightly different :shrug:

src/main/java/org/myrobotlab/programab/handlers/oob/OobProcessor.java
supertick: removed over optimizing, no longer a singleton

src/main/java/org/myrobotlab/programab/models/Event.java
supertick: good point ..
but going to hold off that for a bit .. the src isn't being set and it might be a requirement coming up

src/main/java/org/myrobotlab/programab/handlers/sraix/MrlSraixHandler.java
supertick: Ah, another good point.  I'm trying to imagine if it needs to be filtered or not.  Useful processing should all be structured, but some other remote handler :shrug:

src/main/java/org/myrobotlab/programab/handlers/sraix/MrlSraixHandler.java
supertick: sure .. I'll do this one

src/main/java/org/myrobotlab/programab/handlers/oob/OobProcessor.java
supertick: This isn't invoking a single message .. this is unloading all the json messages .. so mrljson can have a set of msgs
[ { "name":"runtime", "method":"getUptime" ... },
{ "name":"runtime", "method":"getUptime" ... },
{ "name":"runtime", "method":"getUptime" ... },
... etc ]

src/main/java/org/myrobotlab/programab/handlers/oob/OobProcessor.java
supertick: Heh, its a bigger topic.
You and @kwatters are pro strong type.  I'm not against it, but putting the burden on a network json interface to always supply a type isn't "friendly" imo.  Especially, if the type is the actual concrete name of some language type.

AutonomicPerfectionist: How about an optional type parameter, and if it's not present we assume string? Technically we could use the method cache to attempt auto cast to other primitives, but then we get into the whole "which overload to call" mess, would like to avoid repeating that

src/main/java/org/myrobotlab/programab/handlers/oob/OobProcessor.java
AutonomicPerfectionist: Yeah, but the issue is there's special logic to deserialize a single message, and that isn't used when going through the normal from json method, so none of the messages in the array will be deserialized properly

src/main/java/org/myrobotlab/programab/handlers/sraix/MrlSraixHandler.java
AutonomicPerfectionist: Maybe do a try-catch, if it fails to parse then instead parse to an XML document (don't know what Jackson calls it, other libraries I've used let you deserialize to an internal type safe representation), then there should be a way to wrap it in the proper tags using type safe method calls

src/main/java/org/myrobotlab/programab/handlers/oob/OobProcessor.java
supertick: Yeah, a possibility.  
But these are 2 different topics.  For this pr I'm forcing the notation in the aiml json to be an array of messages, for simplicity and the fact many of the actions desired in behaviors require multiple messages.  As far as the parameters, its a very important topic - but in this small pr the context is not the network api, but the embedded oob tags.  I'm just thinking this is a +1 step in the right direction, because previously it was "only" strings. This implementation should handle all primitives as well.

src/main/java/org/myrobotlab/programab/handlers/sraix/MrlSraixHandler.java
supertick: There's a try/catch around the whole block, the the user should be informed of the problem code.
At this point I don't want to try to recover, informing the user of the issue hopefully will get them to correct it.

src/main/java/org/myrobotlab/programab/models/Sraix.java
supertick: :+1:

src/main/java/org/myrobotlab/programab/models/Template.java
supertick: :+1:

src/main/java/org/myrobotlab/programab/models/Template.java
supertick: fixed

src/main/java/org/myrobotlab/programab/models/Template.java
supertick: very good call

src/main/java/org/myrobotlab/programab/models/Oob.java
supertick: done

PR #1397, opened 2024-01-20
InMoov gets a heart

supertick:


.github/workflows/build.yml
```diff
@@ -13,22 +13,22 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set up JDK 11
+      - name: Set up java
         uses: actions/setup-java@v3
 
         with:
-          java-version: "11"
+          java-version: "18"
```
kwatters: do you need java 18 for this PR ?  let's leave this out for now.

src/main/java/org/myrobotlab/programab/Session.java
```diff
@@ -122,7 +124,12 @@ public void savePredicates() {
    */
   public Map<String, String> getPredicates() {
     TreeMap<String, String> sort = new TreeMap<>();
-    sort.putAll(getChat().predicates);
+    // copy keys, making this sort thread safe
```
kwatters, line 127: if there's a thread safety concern here, it should be synchronized.. Did you see issues with thread safety here?

.github/workflows/build.yml
supertick: that is fine here, its elves testing in git actions, it doesn't generate artifacts
git actions gives us the potential opportunity to test multiple version (simultaneously ... when i get around to it)

src/main/java/org/myrobotlab/programab/Session.java
supertick: its a thread safety issue around maps and iterations of the map, not really about the function, but yeah I'll synchronize it

src/test/java/org/myrobotlab/test/AbstractTest.java
```diff
@@ -167,31 +163,31 @@ static protected void installAll() {
...
     }
     if (threadsRemaining.size() > 0) {
       log.warn("{} straggling threads remain [{}]", threadsRemaining.size(), String.join(",", threadsRemaining));
     }
 
-    // log.warn("end of test - id {} remaining services after release {}",
-    // Platform.getLocalInstance().getId(),
-    // Arrays.toString(Runtime.getServiceNames()));
+    // resets runtime with fresh new instance
+    Runtime.getInstance();
```
kwatters, line 189: I'm surprised at the comment here... getInstance() ,  I would expect a getInstance() call to reset the state of Runtime ...

supertick: Its fresh, because the teardown of an AbtractTest now tears down Runtime as well .. which it previously didn't ...
We left runtime hanging out previously

src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -711,12 +711,25 @@ public Session startSession(String path, String userName, String botName, java.u
...
 
     session = new Session(this, userName, botInfo);
-    sessions.put(getSessionKey(userName, botName), session);
+    String sessionKey = getSessionKey(userName, botName);
+    sessions.put(sessionKey, session);
 
     log.info("Started session for bot botName:{} , userName:{}", botName, userName);
     setCurrentSession(userName, botName);
+    
+    invoke("publishSession", sessionKey);
```
kwatters, line 727: we should probably check this.. and publish not just when a session is started, but when the session changes.

supertick: :+1:

.github/workflows/build.yml
AutonomicPerfectionist: Since we still recommend people use Java 11, I think testing on 11 would be best. Don't want to accidentally introduce behavior that only works on 18. I do agree we should start running tests for newer versions as well, but that can be done in a separate PR

src/main/java/org/myrobotlab/process/Terminal.java
```diff
@@ -0,0 +1,400 @@
...
+  // });
+  // });
+  // }
+
+  public Path getTemplatePath(String templateName) {
+    Path scriptPath = Paths.get(service.getResourceDir() + File.separator + "templates" + File.separator, templateName + (isWindows() ? ".bat" : ".sh"));
+    return scriptPath;
+  }
+
+  public String getVersion() {
```
AutonomicPerfectionist, line 215: What is this for? Don't we already have a way to determine the MRL version?

src/main/java/org/myrobotlab/process/Terminal.java
```diff
@@ -0,0 +1,400 @@
...
+    return "0.0.0";
+  }
+
+  public boolean isWindows() {
```
AutonomicPerfectionist, line 219: Isn't this also done in the Platform class?

src/main/java/org/myrobotlab/process/Terminal.java
```diff
@@ -0,0 +1,400 @@
...
+
+  // FIXME - should be synchronized with
+  public String processBlockingCommand(String input) {
+    synchronized (lock) {
+      blockingOutputQueue.clear();
+      processCommand(input, true);
+      String ret = null;
+      try {
+        while (isRunning && ret == null) {
+          ret = blockingOutputQueue.poll(100, TimeUnit.MILLISECONDS);
```
AutonomicPerfectionist, line 267: Probably want a timeout here to prevent it from stalling forever. Configurable by the caller as an additional parameter

src/main/java/org/myrobotlab/process/Terminal.java
```diff
@@ -0,0 +1,400 @@
...
+          String text = new String(buffer, 0, length);
+          // asynchronous publishing of all stdout
+          service.invoke("publishLog", name, text);
+          service.invoke("publishStdOut", text);
+          output.append(text);
+          dynamicBuffer.append(text);
+          System.out.print(text);
+          if (dynamicBuffer.toString().contains(BOUNDARY_MARKER)) {
+            // Boundary marker found, handle command completion here
+            System.out.println("Command execution completed.");
```
AutonomicPerfectionist, line 346: Probably don't want a raw System.out.println right? Same for above as well?

src/main/java/org/myrobotlab/process/Terminal.java
```diff
@@ -0,0 +1,400 @@
...
+      if (bytes != null) {
+        return new String(bytes);
+      }
+    } catch (IOException e) {
+      service.error(e);
+    }
+    return null;
+  }
+
+  // private void startStreamGobbler(InputStream inputStream, String streamName)
```
AutonomicPerfectionist, line 197: Can this commented out code be removed?

src/main/java/org/myrobotlab/process/Terminal.java
```diff
@@ -0,0 +1,400 @@
...
+    output = new StringBuilder();
+  }
+
+  private String determineShellCommand() {
+    String osName = System.getProperty("os.name").toLowerCase();
+    if (osName.contains("win")) {
+      return "cmd";
+    } else {
+      // return "/bin/sh"; // Works for Unix/Linux/Mac
+      String bashPath = "/bin/bash";
```
AutonomicPerfectionist, line 123: This location works for POSIX-compliant systems, but might be worth looking into the `/etc/passwd` file. Counter-intuitively, that file doesn't contain passwords but instead contains information on the username, groups, and the user's default login shell. Sometimes, especially on locked-down enterprise systems, the system shell is only used for system administration and each user has their own shell with access to their own PATH

src/main/java/org/myrobotlab/process/Terminal.java
```diff
@@ -0,0 +1,400 @@
...
+      if (bashFile.exists()) {
+          return bashPath;
+      } else {
+          // Fallback to sh if Bash is not found (less ideal)
+          return "/bin/sh";
+      }
+    }
+  }
+
+  public boolean doesExecutableExist(String name) {
```
AutonomicPerfectionist, line 134: What's the purpose of this?

src/main/java/org/myrobotlab/process/Terminal.java
```diff
@@ -0,0 +1,400 @@
...
+        StringBuilder dynamicBuffer = new StringBuilder();
+        while ((length = inputStream.read(buffer)) != -1) {
+          String text = new String(buffer, 0, length);
+          // asynchronous publishing of all stdout
+          service.invoke("publishLog", name, text);
+          service.invoke("publishStdOut", text);
+          output.append(text);
+          dynamicBuffer.append(text);
+          System.out.print(text);
+          if (dynamicBuffer.toString().contains(BOUNDARY_MARKER)) {
```
AutonomicPerfectionist, line 344: Might not need a boundary marker depending on how this is used. On Linux at least, an EOF signal is sent through the stdout when a program completes, so should be able to look for that

src/main/java/org/myrobotlab/process/Terminal.java
```diff
@@ -0,0 +1,400 @@
...
+        }
+      } catch (IOException e) {
+        service.error(e);
+      }
+    });
+  }
+
+  private void startUserInputForwarder() {
+    executorService.submit(() -> {
+      try (BufferedReader reader = new BufferedReader(new InputStreamReader(System.in))) {
```
AutonomicPerfectionist, line 362: Can we remove the `System.in` and allow it to use any input stream? That way we could eventually do things like allowing the user input to come from the WebGUI, through a service, or through a file

src/main/java/org/myrobotlab/process/PythonTerminal.java
```diff
@@ -0,0 +1,106 @@
...
+  public void activateVirtualEnv() {
+    if (isWindows()) {
+      processCommand(venvName + "\\Scripts\\activate");
+    } else {
+      // source is "bash"
+      // processCommand("source " + venvName + "/bin/activate");
+      // the posix way
+      processCommand(". " + venvName + "/bin/activate");
+    }
+    Service.sleep(300);
```
AutonomicPerfectionist, line 57: Can we remove this sleep by using a blocking processCommand? Also, generally not a good idea to use `activate` when not directly in the system shell. Better to use the wrapper binaries that are installed in the `bin/` folder, like `bin/python` or `bin/pip`

src/main/java/org/myrobotlab/process/PythonTerminal.java
```diff
@@ -0,0 +1,106 @@
...
+
+  public void installVirtualEnv() {
+    installVirtualEnv(venvName);
+  }
+
+  public void installVirtualEnv(String venvName) {
+    this.venvName = venvName;
+    // processCommand(getScriptCmd("python -m venv " + venvName));
+    processCommand("python -m venv " + venvName);
+    Service.sleep(300);
```
AutonomicPerfectionist, line 68: Sometimes created a venv can take a very long time, on some of my systems with hard drives it can take upwards of 30 seconds. Best to use a blocking processCommand

src/main/java/org/myrobotlab/service/TerminalManager.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  protected Map<String, Terminal> terminals = new ConcurrentSkipListMap<>();
+
+  public TerminalManager(String n, String id) {
+    super(n, id);
+  }
+
+  public void deleteTerminal(String name) {
+    log.info("deleting terminal {}", name);
+    if (terminals.containsKey(name)) {
```
AutonomicPerfectionist, line 47: If the Terminal is a thread, just removing it won't automatically kill it. Best to stop the thread, call a cleanup method, and then remove

AutonomicPerfectionist (changes requested):

src/main/java/org/myrobotlab/process/Terminal.java
supertick: This goes away with the new Electron App.

src/main/java/org/myrobotlab/process/Terminal.java
supertick: This goes away with the new Electron App.

src/main/java/org/myrobotlab/process/Terminal.java
supertick: This goes away with the new Electron App.

src/main/java/org/myrobotlab/process/Terminal.java
supertick: This goes away with the new Electron App.

src/main/java/org/myrobotlab/process/Terminal.java
supertick: This goes away with the new Electron App.

src/main/java/org/myrobotlab/process/Terminal.java
supertick: This goes away with the new Electron App.

src/main/java/org/myrobotlab/process/Terminal.java
supertick: This goes away with the new Electron App.

src/main/java/org/myrobotlab/process/PythonTerminal.java
supertick: This goes away with the new Electron App.

src/main/java/org/myrobotlab/process/PythonTerminal.java
supertick: This goes away with the new Electron App .. but long blocking processes should be considered in the new system, and how to handle them in a templated sort of way ...

src/main/java/org/myrobotlab/service/TerminalManager.java
supertick: This goes away with the new Electron App.

src/main/java/org/myrobotlab/service/Log.java
```diff
@@ -132,6 +138,7 @@ public void addError(String msg) {
 
   @Override
   public void addError(String arg0, Throwable arg1) {
+    System.out.println("addError");
```
kwatters, line 141: Is this a debugging statement or something?

PR #1400, opened 2024-02-01
Working processor action in servomixer

supertick:


src/main/java/org/myrobotlab/kinematics/Action.java
```diff
@@ -57,4 +57,11 @@ public static Action createGestureToAction(String gestureName) {
     action.value = gestureName;
     return action;
   }
+
+  public static Action createProcessingAction(String methodName) {
```
AutonomicPerfectionist, line 69: Java doc would be good

src/main/java/org/myrobotlab/service/Py4j.java
```diff
@@ -235,17 +236,19 @@ public void connectionStopped(Py4JServerConnection gatewayConnection) {
    * 
    * @param code The Python code to execute in the interpreter.
    */
-  public void exec(String code) {
+  public boolean exec(String code) {
```
AutonomicPerfectionist, line 240: Need `@Override` if it's part of an interface now

src/main/java/org/myrobotlab/service/Python.java
```diff
@@ -49,7 +50,7 @@
  * @author GroG
  * 
  */
-public class Python extends Service<PythonConfig> implements ServiceLifeCycleListener, MessageListener {
+public class Python extends Service<PythonConfig> implements ServiceLifeCycleListener, MessageListener, Processor {
```
AutonomicPerfectionist, line 53: The exec method should have Override annotation

src/main/java/org/myrobotlab/service/interfaces/Processor.java
```diff
@@ -0,0 +1,18 @@
+package org.myrobotlab.service.interfaces;
+
+public interface Processor {
```
AutonomicPerfectionist, line 3: Java doc here explaining what this is used for would be good

AutonomicPerfectionist (changes requested):

src/main/java/org/myrobotlab/kinematics/Action.java
supertick: updated

src/main/java/org/myrobotlab/service/Python.java
supertick: updated

src/main/java/org/myrobotlab/service/interfaces/Processor.java
supertick: updated

PR #1407, opened 2024-02-12
Framework updates

supertick:


Jenkinsfile
```diff
@@ -64,18 +64,18 @@ pipeline {
...
+      //          expression { params.verify == 'true' }
+      //    }
+      //    steps {
+      //       script {
+      //             sh '''
+      //                mvn test -Dtest=org.myrobotlab.framework.DependencyTest -q
+      //             '''
+      //       }
+      //    }
+      // } // stage dependencies
```
AutonomicPerfectionist, line 78: How come this is commented out?

src/main/java/org/myrobotlab/config/ConfigUtils.java
```diff
@@ -21,51 +34,102 @@ public class ConfigUtils {
...
+    StartYml startYml = new StartYml();
+    String defaultStartFile = CodecUtils.toYaml(startYml);
+    File checkStartYml = new File("start.yml");
+    if (!checkStartYml.exists()) {
+      // save default start.yml
+      startYml = new StartYml();
+      try {
+        FileIO.toFile("start.yml", defaultStartFile);
+      } catch (IOException e) {
+        log.error("could not save start.yml");
```
AutonomicPerfectionist: Probably log the exception as well, could be as simple as a permissions error but this would hide that with a nondescript error message

src/main/java/org/myrobotlab/framework/Service.java
```diff
@@ -516,7 +510,7 @@ public String getResourcePath(String additionalPath) {
    */
 
   static public String getResourceRoot() {
-    return Runtime.getInstance().getConfig().resource;
+    return ConfigUtils.getResourceRoot();//Runtime.getInstance().getConfig().resource;
```
AutonomicPerfectionist: Can the commented out part be removed?

src/main/java/org/myrobotlab/io/FileIO.java
```diff
@@ -870,33 +868,22 @@ public static void main(String[] args) throws ZipException, IOException {
...
-        isr = new FileInputStream(localFilename);
-      } catch (Exception e) {
-        Logging.logError(e);
-        log.error("File not found. {}", localFilename, e);
-        return null;
-      }
+    String resource = ConfigUtils.getResourceRoot();
+    String localFilename = resource + File.separator + src;
+    try {
+      isr = new FileInputStream(localFilename);
```
AutonomicPerfectionist, line 878: Is this going to work for finding resources within the jar file?

src/main/java/org/myrobotlab/opencv/OpenCVFilter.java
```diff
@@ -192,7 +192,7 @@ static private Mat read(String filename) {
   /**
    * This will enable/disable the filter in the pipeline
    */
-  protected boolean enabled = true;
+  protected volatile boolean enabled = true;
```
AutonomicPerfectionist, line 195: Probably fine for now, but should probably be replaced with an atomic. Volatiles still can have race conditions when setting based on their current value, since it can change between when you read and set

src/main/java/org/myrobotlab/opencv/OpenCVFilterMiniXception.java
```diff
@@ -59,7 +59,7 @@ public OpenCVFilterMiniXception(String name) {
   }
 
   private void loadDL4j() {
-    dl4j = (Deeplearning4j) Runtime.createAndStart("dl4j", "Deeplearning4j");
+    dl4j = (Deeplearning4j) Runtime.start("dl4j", "Deeplearning4j");
```
AutonomicPerfectionist, line 62: Wonder if we could eventually turn filters into services so they can have proper peers instead of things like this...

src/main/java/org/myrobotlab/opencv/OpenCVFilterYolo.java
```diff
@@ -378,12 +378,12 @@ public void enable() {
...
   @Override
   public void disable() {
+    super.disable();
     if (classifier == null) {
       // already disabled
       return;
     }
-    super.disable();
-    int waitTime = 0;
+    int waitTime = 0;
```
AutonomicPerfectionist, line 386: Any way to remove the wait here?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -188,7 +191,13 @@ public class Runtime extends Service<RuntimeConfig> implements MessageListener,
...
    * be a valid file name for the OS. It's defaulted to "default". Changed often
    */
-  protected String configName = "default";
+  protected static String configName = "default";
+
+  /**
+   * The runtime config which Runtime was started with. This is the config which
+   * will be applied to Runtime when its created on startup.
+   */
+  // protected static RuntimeConfig startConfig = null;
```
AutonomicPerfectionist, line 200: How come this is commented out?

src/main/java/org/myrobotlab/service/Runtime.java
```diff
@@ -1539,35 +1542,36 @@ static public void install(String serviceType) {
...
 
-    installerThread = new Thread() {
-      @Override
-      public void run() {
-        try {
-          if (serviceType == null) {
-            r.getRepo().install();
-          } else {
-            r.getRepo().install(serviceType);
+      installerThread = new Thread() {
```
AutonomicPerfectionist, line 1555: Can be done with a Runnable lambda passed to the Thread constructor, but that's just semantics. I personally find it more readable that way but not a big deal.

But, should probably check if the thread is nonnull and if it's already running, just in case

src/main/java/org/myrobotlab/service/config/InMoov2Config.java
```diff
@@ -520,7 +526,11 @@ public Plan getDefault(Plan plan, String name) {
     listeners.add(new Listener("publishPlayAudioFile", getPeerName("audioPlayer")));
     listeners.add(new Listener("publishPlayAnimation", getPeerName("neoPixel")));
     listeners.add(new Listener("publishStopAnimation", getPeerName("neoPixel")));
-    listeners.add(new Listener("publishProcessMessage", getPeerName("py4j"), "onPythonMessage"));
+    // listeners.add(new Listener("publishProcessMessage",
+    // getPeerName("python"), "onPythonMessage"));
+    listeners.add(new Listener("publishProcessMessage", "python", "onPythonMessage"));
+    
+    listeners.add(new Listener("publishPython", "python"));
```
AutonomicPerfectionist: How come the python peer's name is hard coded here?

src/test/java/org/myrobotlab/codec/CodecUtilsTest.java
```diff
@@ -186,25 +186,6 @@ public void testDefaultSerialization() {
...
-  public void testCheckServiceNameEqual() {
-    Platform.getLocalInstance().setId("test-id");
-    assertTrue(CodecUtils.checkServiceNameEquality("runtime", "runtime"));
-    assertTrue(CodecUtils.checkServiceNameEquality("runtime", "runtime@test-id"));
-    assertTrue(CodecUtils.checkServiceNameEquality("runtime@test-id", "runtime"));
-    assertTrue(CodecUtils.checkServiceNameEquality("runtime@test-id", "runtime@test-id"));
-    assertFalse(CodecUtils.checkServiceNameEquality("runtime", "runtime@not-corr-id"));
-    assertFalse(CodecUtils.checkServiceNameEquality("runtime@not-corr-id", "runtime"));
-
-  }
```
AutonomicPerfectionist, line 206: How come these tests were removed?

src/test/java/org/myrobotlab/framework/ServiceTest.java
```diff
@@ -1,46 +0,0 @@
...
-import static org.junit.Assert.assertEquals;
-
-import java.util.List;
-import java.util.Map;
-
-import org.junit.Test;
-import org.myrobotlab.service.config.ServiceConfig;
-import org.myrobotlab.test.AbstractTest;
-
-public class ServiceTest extends AbstractTest {
```
AutonomicPerfectionist, line 12: Do we really want to delete this?

src/test/java/org/myrobotlab/service/OpenCVTest.java
```diff
@@ -222,8 +229,8 @@ public final void testAllFilterTypes() {
 
     for (String fn : OpenCV.POSSIBLE_FILTERS) {
       log.warn("trying filter {}", fn);
-      if (fn.startsWith("DL4J") || fn.startsWith("FaceTraining") || fn.startsWith("Tesseract") || fn.startsWith("SimpleBlobDetector") || fn.startsWith("Solr") || fn.startsWith("Split")) {
-        log.info("skipping {}", fn);
+      if ( fn.startsWith("FaceDetectDNN") || fn.startsWith("FaceRecognizer") || fn.startsWith("DL4J") || fn.startsWith("FaceTraining") || fn.startsWith("Tesseract") || fn.startsWith("SimpleBlobDetector") || fn.startsWith("Solr") || fn.startsWith("Split")) {
```
AutonomicPerfectionist, line 232: This can be done by putting all the names of filters you want to skip in a List, and then only loop over filters that aren't in that list. Can even do things like Java stream's `filter()`

Jenkinsfile
supertick: Going back and forth on how dependencies are tested and tests are maintained.
At one point the Jenkins build process mandated a dependency test, now its not required for Jenkinsfile to be specialized, and AbstractServiceTest will install and resolve dependencies like it does in Runtime.
I think that perhaps this is better, but I haven't completely convinced myself.

src/main/java/org/myrobotlab/config/ConfigUtils.java
supertick: :+1:

src/main/java/org/myrobotlab/framework/Service.java

src/main/java/org/myrobotlab/io/FileIO.java
supertick: yeah, its not really "inside" the jar...
default action for mrl on startup is to extract the /resource from the jar and put it on the file system
This is the way war files worked many moons ago, so getting resources doesn't need to contend with the challenges of using a zip filesystem.   The new ConfigUtils.getResourceRoote() gets the info from RuntimeConfig.resource and nowhere else.  This gives us the flexibility to set that value, but default it will point to the extracted /resource

src/main/java/org/myrobotlab/opencv/OpenCVFilter.java
supertick: yeah, regrettably, it didn't fix the issue I'm still trying to fix with the mysterious Yolo thread ...
Now, I don't think it has anything to do with the member variable, but somehow I think the entire filter is getting replaced with a shadow, and the thread gets stuck on a zombie filter

src/main/java/org/myrobotlab/service/Runtime.java
supertick: :+1:

src/test/java/org/myrobotlab/codec/CodecUtilsTest.java
supertick: Setting id is immutable, and now "kinda" enforced

src/test/java/org/myrobotlab/framework/ServiceTest.java
supertick: yes :)

src/test/java/org/myrobotlab/service/OpenCVTest.java
supertick: True, good point.
But, I think I'll save that for later...

src/main/java/org/myrobotlab/service/config/InMoov2Config.java
supertick: fixed

src/main/java/org/myrobotlab/opencv/OpenCVFilterMiniXception.java
supertick: Yeah, not sure if its worth the resources ... :shrug:

src/main/java/org/myrobotlab/opencv/OpenCVFilterYolo.java
supertick: I'm not sure, I've been chasing down a yolo straggling thread for quite a while, it still has eluded me.

src/main/java/org/myrobotlab/service/Runtime.java
supertick: these comments are added no ?

src/test/java/org/myrobotlab/framework/ConfigTest.java
```diff
@@ -38,33 +32,27 @@
...
   @BeforeClass
   public static void setUpBeforeClass() {
-      System.out.println("Runs before any test method in the class");
-  }
-
-  @AfterClass
-  public static void tearDownAfterClass() {
-      System.out.println("Runs after all test methods in the class");
+    // clean out services - reset
+    Runtime.releaseAll(true, true);
```
kwatters, line 39: in a perfect world, this wouldn't be necessary.  It's be nice at some point if we have some sort of a warning message that runtime had to release something... that means the previous unit test didn't clean up after itself as expected.

supertick: I'm not sure, in a perfect world you should be able to verify that the state is "clean"
afaik, the cleanest way is to remove/release everything ... so, its a step closer to idempotent

supertick: Also, AbstractTest or AbstractServiceTest has reporting on at least straggling threads currently

PR #1411, opened 2024-02-18
Channel only responses

supertick:
depends on #1407


src/main/java/org/myrobotlab/service/ProgramAB.java
```diff
@@ -1361,6 +1349,16 @@ public void onUtterance(Utterance utterance) throws Exception {
         if (!config.sleep) {
           shouldIRespond = true;
         }
+        
+        if (config.channels != null && config.channels.size() > 0) {
+          // assume false
+          shouldIRespond = false;
+          for (String channelName : config.channels) {
```
kwatters, line 1356: where as this is functional, it's faster/more scalable/easier to use a set instead of a list for the channels here to determine if the bot should respond to a particular channel, maybe just add a comment in the code that is a TODO to refactor this logic to avoid the loop here and use a set instead.

src/main/java/org/myrobotlab/service/interfaces/UtteranceListener.java
```diff
@@ -11,5 +11,28 @@ public interface UtteranceListener {
   public String getName();
 
   public void onUtterance(Utterance utterance) throws Exception;
+  
+  
+  default public void attachUtterancePublisher(UtterancePublisher publisher) {
```
kwatters, line 16: Doesn't this break the directionality of the attach?  i thought publishers always attached listeners in the new convention.. so adding this here is removing that clean definition..  (I understand the convenience, but it also makes the directional attach convention. 

when using these methods programmatically, it doesn't matter, however, it will get confusing to folks if we don't enforce that directionality of attach in the configs.

PR #1418, opened 2024-04-23
Inmoov2 heart 2

supertick:
In general this is the beginning of a much more structured approach of processing in python.
No longer fragile with name dependencies.  No global vars at all except runtime.

* Instead of a variety of python started timers, InMoov2 gets a heart to drive state and other functions. Simple runnable Heart, publishes a heartbeat
* No python processing untill after "boot" which is when all configuration has processed
* Scripts executing in apply before InMoov2 has started completely removed
* Boot function, driven through stages - verifies config has finished before executing scripts
* Direct references removed from fsm and chatbot
* getLastActivityTime checks for null
* isPirActive() added
* InMoov2 onAudioStart and onAudioEnd added - for callbacks in Python which turn off STT when audioPlayer is running
* New functions playAudioFile, playMusic, nextPlay, searchPlay because its fun to turn InMoov into a jukebox (and this replaces InMoov2/system/1_AudioFile.py)
* Removal of onStarted subscription management - it shouldn't be there because it makes it inaccessible to the user
* FileIO generalized isExecutableAvailable to see if executable is available
* Runtime checks to see if acpi is availble before checking battery level on Linux
* Added publishHeartbeat and processMessage - for all future python side processing, potentially compabible with Jython and Py4j
* InMoov2 setLocale exception when "brain" selected doesn't override i01.chatbot.yml currentBotName
* Removed startMouth startChatBot which needs to be replaced by the more general startPeer


.github/workflows/build.yml
```diff
@@ -23,12 +23,12 @@ jobs:
           cache: "maven"
 
       - name: Install Missing Dependencies
-        run: sudo apt-get install -y libv4l-0 libopencv-dev python3-opencv
+        run: sudo apt-get install -y libv4l-0 libopencv-dev python3-opencv libgtk2.0-0
```
kwatters, line 26: what needs lib tk ? some native ui stuff?

.github/workflows/build.yml
```diff
@@ -23,12 +23,12 @@ jobs:
...
 
       - name: Dependency Test # installs all dependencies
         run: mvn test -Dtest=org.myrobotlab.framework.DependencyTest -q
       - name: Build with Maven # currently cannot test opencv
-        run: mvn clean verify -q
+        run: mvn clean verify -q -Dtest=!org.myrobotlab.opencv.*
```
kwatters, line 31: This should be temporary, is this done to speed up the tests or are there too many failures currently?

.github/workflows/build.yml
supertick: attempting to solve openvc issues on github actions - probably more of a resource issue vs dependency, but there might be a UI dependency

.github/workflows/build.yml
supertick: opencv testing on github actions currently fails the dep add might have helped, but the resources are still probably too low - an actual release on Jenkins doesn't filter

src/main/java/org/myrobotlab/io/FileIO.java
```diff
@@ -1424,5 +1426,35 @@ public static String normalize(String dirPath) {
...
+        Process process = java.lang.Runtime.getRuntime().exec(command);
+
+        // Check the exit value of the process
+        // If the process has terminated correctly, the command is available
+        if (process.waitFor() == 0) {
+            return true;
+        }
+
+        // Read any errors from the attempted command
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getErrorStream()))) {
```
kwatters: what happens if the stderr stream gets filled up and blocks before the process completes?  Seems like the waitFor above might cause a deadlock here...  (possibly?)

supertick: isExecutableAvailable isn't expected to be a long running exe or block, the OS should return an error, or the exe should quickly return version, help or some other relevant information.  That at least was the intent.

src/main/java/org/myrobotlab/io/FileIO.java
```diff
@@ -1424,5 +1426,35 @@ public static String normalize(String dirPath) {
       return dirPath.replace("\\", "/");
     }
   }
+  
+  public static boolean isExecutableAvailable(String command) {
```
kwatters, line 1442: the method name here is misleading, some javadoc to describe the behavior please.  (potentially rename the method to   "attemptToRunCommand"    .. because it doesn't just check to see if the file is executable..

supertick: :+1:

supertick: updated, with javadoc, timeout, and default

src/main/java/org/myrobotlab/service/config/InMoov2Config.java
```diff
@@ -538,8 +538,7 @@ public Plan getDefault(Plan plan, String name) {
     listeners.add(new Listener("publishStopAnimation", getPeerName("neoPixel")));
     // listeners.add(new Listener("publishProcessMessage",
     // getPeerName("python"), "onPythonMessage"));
-    listeners.add(new Listener("publishProcessMessage", getPeerName("python"), "onPythonMessage"));
-    
+    listeners.add(new Listener("publishProcessMessage", getPeerName("python"), "onPythonMessage"));
```
kwatters, line 541: should the callback method be   onProcessMessage  ?  to follow our normal convention ?  we can refactor this to make it consistent at a later time.

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -1486,11 +1626,84 @@ public String publishFlash(String flashName) {
...
-      invoke("publishFlash", "heartbeat");
+  /**
+   * FIXME - get rid of all functionality here - should all be controlled by
+   * behaviors
+   * 
+   * A heartbeat that continues to check status, and fire events to the FSM.
+   * Checks battery, flashes leds and processes all the configured checks in
+   * onHeartbeat at a regular interval
+   */
+  public Heartbeat publishHeartbeat() {
```
kwatters, line 1637: this does break with convention a bit.. normally we'd have a publishHeartbeat(Heartbeat beat)   ... that simply returns the Heartbeat passed... the code below here, could be handled in the heart thread itself instead..  and just invoking publishHeartbeat with the beat that it created in that cycle...  just curious why the departure from that convention?

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -562,17 +667,32 @@ public void enableRandomHead() {
     }
   }
 
+  public void enable() {
```
kwatters, line 670: could this list be more dynamic, perhaps from the default inmoov plan instead of this enumerated list here?

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -2018,105 +2229,6 @@ public void speakBlocking(String format, Object... args) {
     }
   }
 
-  @Deprecated /* use startPeers */
-  public void startAll() throws Exception {
-    startAll(null, null);
```
kwatters, line 2023: can we leave this in and have it call startPeers ?

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -2018,105 +2229,6 @@ public void speakBlocking(String format, Object... args) {
...
-  }
-
-  @Deprecated /* use startPeers */
-  public void startAll(String leftPort, String rightPort) throws Exception {
```
kwatters, line 2027: similar to above, can we just have this call startPeers?  leave this method in for convenience sake?

src/main/java/org/myrobotlab/service/InMoov2.java
```diff
@@ -2132,48 +2244,7 @@ public void startedGesture(String nameOfGesture) {
   }
 
   public void startHeartbeat() {
-    addTask(1000, "publishHeartbeat");
```
kwatters, line 2135: this time should be configurable or pulled out as a constant for the interval here..

PR #1425, opened 2024-06-03
llm vs gpt3

hairygael:


supertick (commented):
I gave up trying to standardize (capital) (lower) for abbreviations.
The service class name is LLM

hairygael:
Ok, it was a good goal though, because it was helpful to remove doubts...
But I understand this is an abbreviation so it make sense to have it all capital.

supertick (approved):

